Issue #11, opened 2017-01-02
implement .extend on Axis


gdementen:
... it should accept a list of labels or an Axis and return a new Axis object (not modify the Axis in-place)

PR #14, opened 2017-01-03
Enhancement (#7 & #11)

### LINKED ISSUE #7: display array titles in viewer
- [x] add in LArray.__init__
- [x] session method to create codebook
- [ ] display in session viewer

alixdamman:
Fixes #7 and #11

### DIFF
--- a/larray/core.py
+++ b/larray/core.py
@@ -946,10 +946,9 @@ class Axis(object):
         Parameters
         ----------
         labels : int, iterable or Axis
-            * Number of new labels (only if wildcard axis).
-            * New labels to append to the axis.
-            * Axis to append. Must (not) be wildcard if self
-              is (not) wildcard.
+            New labels to append to the axis.
+            Passing directly another Axis is also possible.
+            If the current axis is a wildcard axis, passing a length is enough.
 
         Returns
         -------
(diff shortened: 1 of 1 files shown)


larray/core.py
```diff
@@ -936,6 +936,56 @@ def labels(self, labels):
...
+            other = labels
+        else:
+            other = Axis(self.name, labels)
+        if self.iswildcard != other.iswildcard:
+            raise NotImplementedError("Axis to append must (not) be wildcard if " +
+                                      "self is (not) wildcard")
+        if self.iswildcard:
+            return Axis(self.name, self._length + other._length)
+        else:
+            return Axis(self.name, np.append(self.labels, other.labels))
```
gdementen: I would write this like:
```python
labels = self._length + other._length if self.iswildcard else np.append(self.labels, other.labels)
return Axis(self.name, labels)
```
Unsure it is clearer though ;-)

larray/core.py
```diff
@@ -5996,8 +6047,17 @@ def info(self):
         2 x 2
          nat [2]: 'BE' 'FO'
          sex [2]: 'M' 'F'
+        >>> mat1 = LArray(np.ones((2, 2)), [nat, sex], 'test matrix')
```
gdementen: we should implement title in all array creation functions, so that we could write something like this instead:
```python
mat1 = ones((nat, sex), 'test matrix')
```
This should be in anohter PR though.

larray/core.py
```diff
@@ -936,6 +936,56 @@ def labels(self, labels):
...
+        Note that `extend` does not occur in-place: a new axis
+        object is allocated, filled and returned.
+
+        Parameters
+        ----------
+        labels : int, iterable or Axis
+            * Number of new labels (only if wildcard axis).
+            * New labels to append to the axis.
+            * Axis to append. Must (not) be wildcard if self
+              is (not) wildcard.
```
gdementen: I guess you wanted to respect the order of the types, but I would rather have the second line "new labels to ..." first, because it is the most useful/usual option for users. In that case, to avoid confusion (with the types order) you could possibly not use an enumeration but a standard text. Something like:
```
labels : int, iterable or Axis
    New labels to append to the axis, possibly in the form of another Axis. If the current axis is a wildcard axis, passing a length is enough.

gdementen:
in the future, I would appreciate one PR per issue (if they are not linked, of course)

gdementen:
LGTM, merged! Thanks!

PR #20, opened 2017-01-03
Add title in argument list of all array creation functions

### LINKED ISSUE #16: Add title in all array creation functions.

alixdamman:
Fixes #16

### DIFF
--- a/larray/core.py
+++ b/larray/core.py
@@ -8443,7 +8443,7 @@ def diag(a, k=0, axes=(0, 1), ndim=2, split=True):
                     for name, labels in zip(axes_names, axes_labels)]
         else:
             axes = [axis] + [axis.copy() for _ in range(ndim - 1)]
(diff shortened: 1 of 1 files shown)


larray/core.py
```diff
@@ -8429,7 +8443,7 @@ def diag(a, k=0, axes=(0, 1), ndim=2, split=True):
                     for name, labels in zip(axes_names, axes_labels)]
         else:
             axes = [axis] + [axis.copy() for _ in range(ndim - 1)]
-        res = zeros(axes, dtype=a.dtype)
+        res = zeros(axes, title=a.title, dtype=a.dtype)
```
gdementen: I wonder if the title would remain valid after applying diag(). I would lean towards not copying it, what do you think?

alixdamman: There is another else below in which we return a.points[indexer]. I think the title will remain in this case.

PR #31, opened 2017-01-05
Generate xxx_by(axis) methods

### LINKED ISSUE #15: add XXX_by aggregate methods
eg a.sum_by(x.age)

which should be equivalent to 
```python
a.sum(a.axes - x.age)
```
(which does not work, because aggregate functions do not support an AxisCollection argument)

but this works:
```python
a.sum(*(a.axes - x.age))
```

alixdamman:
Fixes #15

### DIFF
--- a/larray/core.py
+++ b/larray/core.py
@@ -6237,21 +6237,24 @@ class LArray(object):
             return self._aggregate(func, args, kwargs,
                                    keepaxes=keepaxes,
                                    commutative=commutative)
+        def method_by(self, axis, **kwargs):
+            return method(self, *(self.axes - axis), **kwargs)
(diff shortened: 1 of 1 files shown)


larray/core.py
```diff
@@ -6237,21 +6237,24 @@ def method(self, *args, **kwargs):
             return self._aggregate(func, args, kwargs,
                                    keepaxes=keepaxes,
                                    commutative=commutative)
+        def method_by(self, axis, **kwargs):
```
gdementen: wouldn't it be better to accept *args so that you can use several axes in *_by?
eg. sum_by(age, sex)

PR #32, opened 2017-01-09
Viewer : array table split into 3 tables (xlabelsview, ylabelsview, arrayview)

alixdamman:
Note : 
- Resizing of ylabel columns not allowed. 
- Filtering on anonymous axes still need to be fixed but I prefer in other PR.

### DIFF
--- a/larray/viewer.py
+++ b/larray/viewer.py
@@ -361,6 +361,204 @@ def get_idx_rect(index_list):
     return min(rows), max(rows), min(cols), max(cols)
 
 
+class AbstractTableModel(QAbstractTableModel):
+    """
+    Abstract class implementing common methods
+    for labels and array models
+    """
+    def __init__(self, data=None, readonly=False, font=None, parent=None):
+        QAbstractTableModel.__init__(self)
+
+        self.dialog = parent
+        self.readonly = readonly
+
+        if font is None:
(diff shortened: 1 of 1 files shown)


gdementen:
I finally tried your PR... almost 7 months late is better than never, right ? ;-)

I made a few changes, but since I don't know how to push to your branch, I created a new one on upstream:

https://github.com/liam2/larray/commit/a0d33269f3b96a92f8e9c55ca084254b134e22f7

A few general comments: 
* unsure this is the right approach as it makes the code quite verbose, and I don't know how it would interact with the viewer refactor you did.
* if this approach is continued, it should be taken one step further: split the area in 4 pieces instead of 3, so that the top-left corner never moves. Currently scrolling to the right makes for a very confusing behavior.
* wouldn't the approach recommended by Qt for freezing columns (ie overlay two Tableview) instead of having them next to each other be simpler? See https://gist.github.com/gdementen/21a78ac56258c07dbc1072b806a5097a for a (primitive) example using qtpy.

alixdamman:
This is a very old branch, never updated. I will take a look at this next week.
I started this first by following the freezing columns approach but you told me that it could slow down the viewer, so I changed my approach.

alixdamman:
closed via https://github.com/larray-project/larray-editor/pull/34

PR #46, opened 2017-01-18
Fix #13 :  Replace index_col default value to None instead of [] 

### LINKED ISSUE #13: read_excel(nb_index=...) is broken
should also change other methods index_col default value to None instead of [], but the code needs to be changed too

alixdamman:
and update associated functions implementation.

### DIFF
--- a/larray/core.py
+++ b/larray/core.py
@@ -7729,7 +7729,7 @@ def read_csv(filepath, nb_index=0, index_col=None, sep=',', headersep=None,
         if dialect == 'liam2':
             nb_index -= 1
 
-    if isinstance(index_col, list) and len(index_col) > 0:
+    if isinstance(index_col, list):
         nb_index = len(index_col)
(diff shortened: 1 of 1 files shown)


larray/core.py
```diff
@@ -7724,12 +7724,12 @@ def read_csv(filepath, nb_index=0, index_col=[], sep=',', headersep=None,
...
         axes_names = header[:pos_last + 1]
 
-    if len(index_col) == 0 and nb_index == 0:
+    if index_col is None and nb_index == 0:
         nb_index = len(axes_names)
         if dialect == 'liam2':
             nb_index -= 1
 
-    if len(index_col) > 0:
+    if isinstance(index_col, list) and len(index_col) > 0:
```
gdementen: len(index_col) > 0 is not longer necessary I think (it was to differentiate from the default value)

Issue #41, opened 2017-01-18
Change signature of functions sum, mean, ...


alixdamman:
Replace *args and **kwargs by equivalent arguments of Numpy functions.

alixdamman:
Doc should be enough (see #43).

gdementen:
The point was partially to make PyCharm's Ctrl-P work, and PR #43 currently does not solve this.

PR #57, opened 2017-01-24
Doctstring agg methods (fix #43)

### LINKED ISSUE #43: add docstrings (& examples) for aggregate methods
We might want to use a template for part of it.

alixdamman:
Fix #43

Also split api.rst in several sections.

Doctests of xxx_by methods need to be extended to groups in other PR. 
xxx_by do not support groups currently.


larray/core.py
```diff
@@ -7043,40 +7043,40 @@ def percentile_by(self, q, *args, **kwargs):
...
-          | 2.5 | 0.5 | 1.0 | 1.0
+          | 2.5 | 1.5 | 0.0 | 1.5
         >>> # split axis 'a' in two parts
-        >>> round(arr.std('a0,a1;a2,a3'), 2)
+        >>> arr.std('a0,a1;a2,a3')
           a\\b |  b0 |  b1 |  b2 |  b3
-        a0,a1 | 2.5 | 0.5 | 1.0 | 1.0
-        a2,a3 | 1.5 | 4.0 | 2.5 | 1.5
+        a0,a1 | 2.5 | 1.5 | 0.0 | 1.5
+        a2,a3 | 2.5 | 1.5 | 0.0 | 1.5
```
gdementen: It is a bit annoying that the two groups have the same result. I am unsure it's worth changing though...

doc/source/api.rst
```diff
@@ -4,11 +4,137 @@ API Reference
...
+.. autoclass:: larray.excel.Workbook
+:members:
+
+.. autoclass:: larray.excel.Sheet
+:members:
+
+.. autoclass:: larray.excel.Range
+:members:
+
+Array Generator Functions
```
alixdamman: Agree. Sorry.

larray/core.py
```diff
@@ -3287,48 +3223,18 @@ def any(values, axis=None):
...
     """
-    Sum of array elements over a given axis.
+    Sum of array elements.
+
+    Calls :func:`~larray.LArray.sum` method of LArray class if `array`
+    is an LArray object, uses builtin sum function otherwise.
 
     Parameters
     ----------
     array : LArray or iterable
```
gdementen: We either need to document all possible arguments (after having changed the signature of those functions -- see my comment at #41), or not document them at all (which is fine by me). Having only array is a bit weird.

alixdamman: I prefer to not document them. I guess there are not gonna be used so much.

larray/core.py
```diff
@@ -4078,25 +3679,146 @@ def aslarray(a):
...
+
+            * ('a1, a3, a5') : labels separated by commas in a string
+            * (a='a1, a3, a5') : in case of possible ambiguity, i.e. if labels
+              can belong to more than one axis, you must precise the axis.
+            * ('a1:a5:2') : select labels using a slice (syntax is 'start:end:step')
+            * ('a1:a3; a10:a12') : create several groups with semicolons.
+              Names are simply given by the concatenation of labels
+              (here: 'a1,a2,a3' and 'a10,a11,a12')
+            * ('a1:a3 >> a1_3; a10:a12 >> a12_12') : operator ' >> '
+              allows to rename groups.
```
gdementen: I think this section about groups would be clearer for the user if it was explained in steps:
* coma list or slice
* ambiguous labels need their axis to be specified using kwarg OR 'axis_name[]' OR x.axis_name[]
* several groups are possible
* any group can be renamed

The main problem with this explanation is that it is easy to miss the fact you can do all the above without using the string syntax, which I prefer users to avoid (in their models at least).

alixdamman:
Question: is it possible to mix x.axis_name[] and >> ?

alixdamman:
Do we agree on examples on all method?

gdementen:
LGTM (looks good to me)

gdementen:
Thanks!

PR #58, opened 2017-01-24
fix51

gdementen:


larray/core.py
```diff
@@ -342,12 +346,16 @@ def _range_str_to_range(s):
...
+    >>> list(_range_str_to_range('-1..2'))
+    [-1, 0, 1, 2]
     >>> _range_str_to_range('a..c')
     ['a', 'b', 'c']
     >>> list(_range_str_to_range('2..6 step 2'))
     [2, 4, 6]
+
+    any special character except . and spaces should work
+    >>> _range_str_to_range('a|+*@-b .. a|+*@-d')
+    ['a|+*@-b', 'a|+*@-c', 'a|+*@-d']
```
alixdamman: What is the purpose of that ?

larray/core.py
```diff
@@ -3676,7 +3688,7 @@ def percentile(array, *args, **kwargs):
      a0 |  0 |  1 |  2
      a1 |  3 |  4 |  5
     >>> percentile(arr, 25)
-    array(1.25)
+    1.25
```
alixdamman: here percentile returns a ndarray with one element. 
I guess it's due to how _aggregate is implemented. 
I can replace by print(percentile(arr, 25)) to get a scalar.

gdementen: It shouldn't after this PR.

larray/core.py
gdementen: so that you can write: "a+b .. a+d" or anything like that. The doctest example is obviously extreme, but the point was to test several of them in one go.

larray/core.py
```diff
@@ -3699,11 +3711,9 @@ def ptp(array, *args, **kwargs):
...
         Input data.
     axis : None or int or str or Axis, optional
-        Axis along which to find the peaks.
-        The default (`axis` = `None`) is to find the peaks
-        over all axes of the input array (as if the array is flatten).
-        `axis` may be negative, in which case it counts from the last
-        to the first axis.
+        Axis along which to find the peaks. The default (`axis` = `None`) is to find the peaks over all axes of the
+        input array (as if the array was flattened).
+        `axis` may be negative, in which case it counts from the last to the first axis.
```
alixdamman: docstring of aggregate function are gonna be generated using a special function and moved in LArray class in a next PR. But OK for now.

alixdamman:
Nothing to add. Ok for merging.

PR #61, opened 2017-01-24
multi_groups

gdementen:
This makes the following code work (we should make a unit test out of this), does not break any test, but I fear it breaks an existing model (because of the code I had to change), so we need to check that before merging.

```pycon
>>> from larray import *
>>> arr2 = ndrange('a=a0,a1,a2;b=a0,a1,a2')
>>> groups = x.a['a0,a1'] >> 'a', x.a['a2'] >> 'b'
>>> arr2.sum(groups, x.b[groups])
a\\b |  a | b
  a |  8 | 7
  b | 13 | 8
>>> groups = 'a0,a1 >> a;a2 >> b'
>>> arr2.sum(x.a[groups], x.b[groups])
a\\b |  a | b
  a |  8 | 7
  b | 13 | 8
```

### DIFF
--- a/doc/source/contribute.rst
+++ b/doc/source/contribute.rst
@@ -6,13 +6,10 @@ Getting the code (for the first time)
 
 - install a Git client
 
-  On Windows, TortoiseGit provides a nice graphical wrapper. You need to
-  install both the console client from http://msysgit.github.io/ and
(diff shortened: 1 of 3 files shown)


alixdamman:
Thanks for the multi updates

PR #62, opened 2017-01-24
Add tests for to_excel and read_excel (closes #12 and #13)

### LINKED ISSUE #12: add unit tests for Excel I/O
This is a very important part and it is not tested at all currently, and I manage to break it every other release.

alixdamman:


larray/tests/test_la.py
```diff
@@ -2906,6 +2932,19 @@ def test_to_csv(self):
...
+        self.assertEqual(la.ndim, 5)
+        self.assertEqual(la.shape, (2, 5, 2, 2, 3))
+        self.assertEqual(la.axes.names, ['arr', 'age', 'sex', 'nat', 'time'])
+        assert_array_equal(la[x.arr[1], 0, 'F', x.nat[1], :],
+                           [3722, 3395, 3347])
+
+        la.to_excel('out.xlsx', '5d')
+        out = read_excel('out.xlsx', '5d').i[:2]
+        result = la.i[:2]
+        assert_array_equal(out, result)
```
gdementen: Since we read the full array anyway, why not compare the full array?

larray/tests/test_la.py
```diff
@@ -2855,6 +2860,59 @@ def test_readcsv(self):
         assert_array_equal(la[x.arr[1], 0, 'F', x.nat[1], :],
                            [3722, 3395, 3347])
 
+    def test_read_excel_xlwings(self):
+        if xw is not None:
```
gdementen: We should rather use something like:
```python
    @unittest.skipIf(xw is None, "xlwings is not available")
    def test_read_excel_xlwings(self):
        pass
```
See https://docs.python.org/2/library/unittest.html#skipping-tests-and-expected-failures

larray/tests/test_la.py
```diff
@@ -2855,6 +2860,59 @@ def test_readcsv(self):
...
+        self.assertEqual(la.axes.names, ['time'])
+        assert_array_equal(la, [3722, 3395, 3347])
+
+        la = read_excel(abspath('test.xlsx'), '2d', nb_index=1, engine='xlrd')
+        self.assertEqual(la.ndim, 2)
+        self.assertEqual(la.shape, (5, 3))
+        self.assertEqual(la.axes.names, ['age', 'time'])
+        assert_array_equal(la[0, :], [3722, 3395, 3347])
+
+        la = read_excel(abspath('test.xlsx'), '3d', index_col=[0, 1], engine=None)
```
gdementen: this will test the xlwings code path, not pandas

larray/tests/test_la.py
```diff
@@ -2906,6 +2964,18 @@ def test_to_csv(self):
         with open('test_out1d.csv') as f:
             self.assertEqual(f.readlines(), result)
 
+    def test_to_excel(self):
+        la = read_excel(abspath('test.xlsx'), '5d', nb_index=4, engine=None)
```
gdementen: why the explicit engine=None?

larray/tests/test_la.py
```diff
@@ -2855,6 +2860,59 @@ def test_readcsv(self):
...
+
+            la = read_excel(abspath('test.xlsx'), '5d')
+            self.assertEqual(la.ndim, 5)
+            self.assertEqual(la.shape, (2, 5, 2, 2, 3))
+            self.assertEqual(la.axes.names, ['arr', 'age', 'sex', 'nat', 'time'])
+            assert_array_equal(la[x.arr[1], 0, 'F', x.nat[1], :],
+                               [3722, 3395, 3347])
+
+    def test_read_excel_pandas(self):
+        la = read_excel(abspath('test.xlsx'), '1d', index_col=[0], engine='xlrd')
```
gdementen: it's strange that it works... there should not be any index in this case?!

alixdamman: If we don't use index_col=[0] (i.e. index_col=None), we get
{0} | time |   2007 |   2010 |   2013
    |  nan | 3722.0 | 3395.0 | 3347.0

larray/tests/test_la.py
alixdamman: From read_excel function, I understand that xlwings is used if engine='xlwings'.

larray/tests/test_la.py
alixdamman: Because read_excel from pandas accept either xlrd or None.

larray/tests/test_la.py
gdementen: oups, sorry. This is not the same as in to_excel. We should probably have the same behavior regarding engine in both. Unsure which, but we need to be consistent.

larray/core.py
```diff
@@ -7808,20 +7808,50 @@ def read_hdf(filepath, key, na=np.nan, sort_rows=False, sort_columns=False,
...
     """
     Reads excel file from sheet name and returns an LArray with the contents
-        nb_index: number of leading index columns (e.g. 4)
-    or
-        index_col: list of columns for the index (e.g. [0, 1, 3])
+
+    Parameters
+    ----------
+    filepath : str
+        Path where the Excel file has to be written.
```
gdementen: sorry I did not see it earlier but... written should be read.

gdementen:
LGTM, Thanks!

PR #71, opened 2017-01-26
Fix #66

### LINKED ISSUE #66: implement axes (nb_index) autodetection for read_excel using pandas/xlrd backend

alixdamman:
Maybe new added tests are not enough strong and do not cover all possibilities.


alixdamman:
OK. Problems with Python 2.7 (again). I'll check on that.

gdementen:
Problem is at
https://github.com/liam2/larray/blob/master/larray/core.py#L852
it should be unicode instead of str. See:
https://github.com/liam2/larray/blob/master/larray/utils.py#L22

gdementen:
Thanks!

PR #83, opened 2017-02-01
update import in viewer (matplotlib backend)

alixdamman:
Trying to import FigureCanvas from matplotlib.backends.backend_qt5agg is not enough. 
The import can goes well even if Qt5 is not installed and then FigureCanvas(figure) will crash. 

Another way would be to test the environment variable 'QT_API' set by qtpy when used. 
os.environ['QT_API'] returns pyqt5/pyqt/pyside if PyQt5/PyQt4/Pyside is used. 
The problem is with Pyside since we don't know the Qt version. 

A final option would be to force the Qt version (to 5).

### DIFF
--- a/larray/viewer.py
+++ b/larray/viewer.py
@@ -90,16 +90,18 @@ from qtpy.QtGui import (QColor, QDoubleValidator, QIntValidator, QKeySequence,
 from qtpy.QtCore import (Qt, QModelIndex, QAbstractTableModel, QPoint, QItemSelection,
                          QItemSelectionModel, QItemSelectionRange, QVariant, Slot)
 
+from qtpy import QT_VERSION
+
 import numpy as np
 
 try:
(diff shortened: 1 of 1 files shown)


gdementen:
Nice catch. I did test my code using Qt4 installed, so I wonder why it worked :) The current PR is a no-go though (breaks on Travis). As for the best way to fix this, I let you judge (I have no idea). I would rather keep qt4 compatibility for now IF that is doable in a cross-platform way and it is not too hard. If that is too hard, yes we could scrap the support for Qt4 (but that would make it yet harder to merge back our changes in spyderlib).

As a side note, doesn't qtpy "export" that information within python? It seems strange to go via an environment variable to communicate between two Python modules.

alixdamman:
This is actually done in PR #86 (use labels for x axis when plotting in viewer).

PR #84, opened 2017-02-02
fix #33 : Filtering on anonymous axes in viewer

### LINKED ISSUE #33: Filtering on anonymous axes in viewer does not work

alixdamman:
replaced axis.name by axis itself in cur_filter dictionary. Did you use axis.name for a specific reason?

### DIFF
--- a/larray/viewer.py
+++ b/larray/viewer.py
@@ -1503,15 +1503,13 @@ class ArrayEditorWidget(QWidget):
         cur_filter = self.current_filter
         # if index == 0:
         if not indices or len(indices) == len(axis.labels):
-            # FIXME: anonymous...
-            if axis.name in cur_filter:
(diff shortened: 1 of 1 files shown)


gdementen:
There could have been many reasons, I don't remember which is correct:

* anonymous axes did not exist (were not supported) when I created the viewer.
* Axes were not hashable when I created the viewer
* hashing an Axis was potentially too expensive for large axes (ie hash on the labels)
* I wanted to implement filters on raw np.ndarray later and thus wanted to limit the internal API to something simple.

Now the two last last are the only ones which are still relevant. We might change the hashing of axes in the future, so it might be a good idea to use the axis id (accessible through axes.ids and NOT via axis.id) instead.

gdementen:
Thanks!

PR #85, opened 2017-02-02
Fix #80 : implement a.axis_name

### LINKED ISSUE #80: implement a.axis_name
as a shortcut for a.axes.axis_name

The current `__getattr__` behavior is to get at the axes labels. ie arr.a0 is currently equivalent to arr['a0'].
This is luckily undocumented (and unused AFAIK), so we can change it easily.

The most important problem with the current situation is that integer axes do not work:
arr.10 is a syntax error. so having yet another different way to get a subset of an array is a bit weird.

For axes though, we usually do not have this problem: most axes names are string (or inexistent). As usual, the implementation is easy, but deciding what's best is not.

alixdamman:


gdementen:
In that case, I would be nice to implement `LArray.__dir__` too (see `AxisCollection.__dir__` for an example). Besides, the tutorial/doc should be updated in a few places. eg. pop.axes['age'] should be pop.age instead (it should have been pop.axes.age in the first place but that does not matter now ;-))

gdementen:
thanks!

PR #86, opened 2017-02-03
Fix #74 : use labels for x axis when plotting in viewer

### LINKED ISSUE #74: Add labels to x axis when plotting in viewer

alixdamman:
Notes : 
1) I always assume that columns (i.e. last axis) represent the x axis of the graph. ylabels may represent several axes and I'm not sure it is a good idea to make plot where xticks represent combinations of labels of different axes. 
2) Each row is drawn as a new line.
3) Is there a specific reason why Pandas is not imported in the viewer? 
    The plot method of LArray class uses the one from Pandas. 
    Shouldn't we do the same in the viewer to be consistent?

### DIFF
--- a/larray/viewer.py
+++ b/larray/viewer.py
@@ -1056,8 +1056,9 @@ class ArrayView(QTableView):
 
     def _selection_data(self, headers=True, none_selects_all=True):
         """
-        Returns a Numpy ndarray containing selected data if headers=True.
-        Returns an iterator over labels and data if headers=False.
+        Returns an iterator over selected labels and data
+        if headers=True and a Numpy ndarray containing only
+        the data otherwise.
 
         Parameters
         ----------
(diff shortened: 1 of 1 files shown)


gdementen:
Regarding Pandas. Yes, that is an option. We would have to actually investigate what value it brings to the table though. The original reason was the usual one, to avoid adding a dependency that spyder wouldn't like (to eventually merge back with their code). But I fear we are going further and further away from that. :(

larray/viewer.py
```diff
@@ -1604,6 +1648,103 @@ def map_filtered_to_global(self, k):
...
+    """
+    assert isinstance(data, np.ndarray)
+
+    if data.ndim == 0:
+        data.shape = (1, 1)
+        xlabels = [["{0}"], [0]]
+        ylabels = [[]]
+    elif data.ndim == 1:
+        data = data.reshape(1, data.shape[0])
+        xlabels = [["{0}"], list(range(data.size))]
```
gdementen: isn't ndim==1 just a special case of ndim >=1 regarding xlabels & ylabels (ie can we reuse the same code?)

larray/viewer.py
```diff
@@ -1604,6 +1648,103 @@ def map_filtered_to_global(self, k):
...
+    Parameters
+    ----------
+    data : numpy.ndarray
+        Input array.
+
+    Returns
+    -------
+    data : 2D array
+        Content of input array is returned as 2D array.
+    xlabels : 1D array
```
alixdamman: copy/paste docstring from larray_to_array_and_labels(data) below. I'll change that.

larray/viewer.py
```diff
@@ -1604,6 +1648,103 @@ def map_filtered_to_global(self, k):
...
+        xlabels = [["{0}"], list(range(data.size))]
+        ylabels = [[]]
+    elif data.ndim > 1:
+        xlabels = [["{{{}}}".format(i) for i in range(data.ndim)],
+                   range(data.shape[-1])]
+        coldims = 1
+        prod = Product([range(data.shape[i]) for i in range(data.ndim - 1)])
+        ylabels = [_LazyNone(len(prod) + coldims)] + [
+            _LazyDimLabels(prod, i) for i in range(data.ndim - 1)]
+    else:
```
alixdamman: Empty ndarray

gdementen:
nice job!

PR #96, opened 2017-02-08
fromstring

gdementen:


alixdamman:
Have you tested what happened if one row contains more columns than expected ?
A user could write ,, somewhere by mistake.

gdementen:
No, I did not test that, but I think it's fine to fail in that case as long as the error message is clear, which is probably not the case.

gdementen:
rebased on current master

gdementen:
could be squashed though

alixdamman:
Thanks

gdementen:
hmmm, I just realize this is severly broken... it obviously (when looking at the code) creates a string array not an integer array like one would expect!

PR #103, opened 2017-02-10
(issues 78 & 79) Add metadata to larray objects (preliminary review)

alixdamman:
Partial PR. Do not merge yet. 

Implementation dilemma:
--------------------------
First option (not the one chosen in this PR): 
We implement metadata as attributes in Variable class from Xarray 
( https://github.com/pydata/xarray/blob/master/xarray/core/variable.py ). 
User is forced to use 'attrs' to play with 'netCDF-like' attributes: 
>>> var.attrs['description'] = 'short description'

Second option (the one chosen in this PR):
We use an OrderedDict object to store our metadata (as in Xarray) but 
we modify/override __getattr__,  __setattr__, __delattr__ and __dir__ from LArray. 
From the new unit test 'test_metadata', you'll see that we can do stuff like:
>>> array.creation_date = datetime(2017, 2, 10)
>>> del array.creation_date
instead of:
>>> array.meta['creation_date'] = datetime(2017, 2, 10)
>>> del array.meta['creation_date']

I would like to have your opinion/review first before to proceed and 
start to update I/O functions.


alixdamman:
I wonder if we shouldn't force user to use only str, datetime, int, float or complex as value for metadata?

gdementen:
One more thing... If we choose option 2 or 4, there is also the question of what to do about .title and .description. We could either 
1) scrap them (make them "normal" user-defined attributes), 
2) keep them as "blessed" attributes (that always exist) but have users set/access them via normal .attrs.title
3) keep them as "blessed" attributes and provide shortcut properties .title and .description to set/access them like now.
I don't really know which would be best. Do you have a strong opinion about this?

larray/core.py
```diff
@@ -3763,19 +3778,24 @@ class LArray(object):
...
 
     Attributes
     ----------
     data : NumPy ndarray
         Data.
     axes : AxisCollection
         Axes.
-    title : str
-        Title.
+    metadata : OrdereDict(str, object)
```
alixdamman: Alright. After all, if some users want to keep insertion order, they just need to upgrade to Python 3.6: https://mail.python.org/pipermail/python-dev/2016-September/146327.html
In that case, we can use a simple dictionary to store metadata.

larray/core.py
```diff
@@ -3915,22 +3952,69 @@ def with_axes(self, axes):
...
+            object.__delattr__(self, item)
 
     def __dir__(self):
         names = set(axis.name for axis in self.axes if axis.name is not None)
-        return list(set(dir(self.__class__)) | names)
+        metadata = set(self._metadata.keys())
+        return list(set(dir(self.__class__)) | names | metadata)
+
+    @property
+    def meta(self):
```
alixdamman: I'm OK with 'attrs' instead of 'meta'

larray/core.py
```diff
@@ -3915,22 +3952,69 @@ def with_axes(self, axes):
...
+        """Print all metadata associated with the array.
+
+        Returns
+        -------
+        str:
+            List of metadata associated with the array.
+        """
+        return '\n'.join(["{}: {}".format(k, v) for k, v in self._metadata.items()])
```
alixdamman: Seems legit

larray/core.py
```diff
@@ -3915,22 +3952,69 @@ def with_axes(self, axes):
...
+
+    # XXX : we could simply return self._metadata in property meta
+    # and then make array.meta['key'] = value but I want to check that
+    # key is not 'data' or 'axes'
```
alixdamman: If we go for option 2 or 4, we will no longer need to make such checks

alixdamman:
I would go for option 4: 
>>> arr.attrs.creation_date = datetime(2017, 2, 10)
>>> arr.attrs.creation_date
datetime.datetime(2017, 2, 10, 0, 0)
>>> arr.attrs.creation_date = datetime(2017, 2, 16)
>>> del arr.attrs.creation_date
which is nicer to read than option 2.

Concerning .title and .description, I changed my mind and prefer to let the users decide what attributes they want to set.

alixdamman:
I do not forget to handle:
I missed that in the arr.axes.name -> arr.name change, but since it is not longer the case for __getattr__, it should not be the case for __setattr__ either.

I will do it later.

larray/core.py
```diff
@@ -3978,7 +3992,7 @@ def replace_axes(self, axes_to_replace=None, new_axis=None, **kwargs):
             items += kwargs.items()
             for old, new in items:
                 axes = axes.replace(old, new)
-        return LArray(self.data, axes, title=self.title)
+        return LArray(self.data, axes)
```
gdementen: shouldn't we copy attrs in this case?

larray/core.py
```diff
@@ -3999,7 +4013,26 @@ def __setattr__(self, key, value):
 
     def __dir__(self):
         names = set(axis.name for axis in self.axes if axis.name is not None)
-        return list(set(dir(self.__class__)) | names)
+        attrs = set(self._attrs.keys())
```
gdementen: attrs should not appear in `LArray.__dir__` (since they are not available as arr.xyz), but in `AttrDict.__dir__`

larray/core.py
```diff
@@ -3770,28 +3781,33 @@ class LArray(object):
...
     ----------
     data : NumPy ndarray
         Data.
     axes : AxisCollection
         Axes.
-    title : str
-        Title.
+    \**kwargs :
+        Attributes (title, description, creation date, ...)
+        associated with the array
```
gdementen: Unsure this is a good idea because we have the potential for a conflict between a user-defined argument and a future argument we add. And if the user makes a typo for an existing argument it will just work (and create a user-defined attribute instead). e.g. LArray(data, axis=Axis('sex', 'H,F'))

larray/core.py
```diff
@@ -3999,7 +4013,26 @@ def __setattr__(self, key, value):
...
+        return list(set(dir(self.__class__)) | names | attrs)
+
+    @property
+    def attrs(self):
+        """Returns attributes of the array.
+
+        Returns
+        -------
+        dict:
+            Attributes of the the array.
```
gdementen: the the

larray/utils.py
```diff
@@ -433,3 +433,22 @@ def find_closing_chr(s, start=0):
                     return pos
     raise ValueError("malformed expression: reached end of string without "
                      "finding the expected '{}'".format(match[needle]))
+
+
+# adapted from https://github.com/fabric/fabric/blob/1.13.1/fabric/utils.py
```
gdementen: I am unsure the source citation is warranted for such a simple code...

larray/utils.py
```diff
@@ -433,3 +433,22 @@ def find_closing_chr(s, start=0):
...
+        try:
+            return self[key]
+        except KeyError:
+            raise AttributeError(key)
+
+    def __setattr__(self, key, value):
+        self[key] = value
+
+    def __delattr__(self, key):
+        del self[key]
```
gdementen: It would be nice if it had a nicer repr than the default dict.

larray/core.py
alixdamman: I used kwargs to not break the title='...' argument. If we don't use kwargs for attributes, then we force the user to construct a dictionary.

larray/core.py
alixdamman: We should define a (simple) rule for users to know when attributes are copied and when they are not.

larray/core.py
alixdamman: Indeed.

larray/utils.py
alixdamman: Agree

alixdamman:
First attempt to update to/read_csv so as to handle attributes. If you other better ideas ...

alixdamman:
Concerning what to do with array's attributes in read/write functions: 
Do we save attributes and array in the same CSV/Excel sheet ? Or do we save them in separate CSV files/Excel sheets?

larray/core.py
```diff
@@ -3873,10 +3894,20 @@ def __init__(self,
                 raise ValueError("length of axes %s does not match "
                                  "data shape %s" % (axes.shape, data.shape))
 
+        _attrs = _AttributeDict()
+        title = kwargs.pop('title', '')
```
alixdamman: I'm still not sure what to do. For me, we should either consider all kwargs as attributes or force users to pass a dictionary/use a.attrs.new_key = new_value.

larray/core.py
```diff
@@ -5875,11 +5967,11 @@ def posargsort(self, axis=None, kind='quicksort'):
     def copy(self):
         """Returns a copy of the array.
         """
-        return LArray(self.data.copy(), axes=self.axes[:], title=self.title)
+        return LArray(self.data.copy(), self.axes[:], attributes=self.attrs)
```
alixdamman: Yes, we should. I simply forgot to copy attrs.

larray/core.py
```diff
@@ -8330,11 +8451,14 @@ def to_hdf(self, filepath, key, *args, **kwargs):
...
         Examples
         --------
-        >>> from .tests.test_la import abspath
         >>> a = ndrange('nat=BE,FO;sex=M,F')
-        >>> a.to_hdf(abspath('test.h5'), 'a')
+        >>> # save array in HDF format
+        >>> a.to_hdf('test.h5', 'a') # doctest: +SKIP
         """
-        self.to_frame().to_hdf(filepath, key, *args, **kwargs)
+        store = pd.HDFStore(filepath)
```
alixdamman: Seems that Travis is unhappy for other reasons.

larray/core.py
```diff
@@ -3784,6 +3789,143 @@ def _doc_agg_method(desc, by=False, action="perform",
...
+                value = int(value)
+            elif cell.ctype == 3:
+                value = datetime.datetime(*xldate_as_tuple(value, 0))
+            elif cell.ctype == 4:
+                value = bool(value)
+            return value
+
+        attributes = _AttributeDict()
+        rowx = 0
+        if sheet.cell(rowx=rowx, colx=0).value == 'Attributes':
```
gdementen: this is "dangerous". If, for any reason, you have an Attributes cell without a 'Data' cell, you could get an infinite loop (or a random Exception when trying to access an invalid cell). I usually try to avoid "while True:" unless the "breaking condition" is bullet-proof.

gdementen: the same comment applies above for the xlwings code

alixdamman:
Travis only crashes for Python <= 3.4 now.

gdementen:
You should probably hold on working on this PR, as we will probably need to change the format. At least we need to talk about this with our colleagues.

alixdamman:
I'm not sure .attrs is the best choice. What about .meta? 
xarray uses .attrs but it's because the 'attribute' word has a meaning in the context of netCDF.

gdementen:
As a user, I **personally** do not care either way (both are fine to me). I would think that `attrs` is less "technical" and hence better for our users than `meta`, but we would have to do a little poll about this to know...

alixdamman:
@gdementen Could you review the first commit?

larray/core/metadata.py
```diff
@@ -0,0 +1,124 @@
...
+        if len(self):
+            hdfstore.get_storer(key).attrs.metadata = self
+
+    @classmethod
+    def from_hdf(cls, hdfstore, key):
+        if 'metadata' in hdfstore.get_storer(key).attrs:
+            return hdfstore.get_storer(key).attrs.metadata
+
+
+if PY2:
```
gdementen: having different functionality (and docstring) on Python2 bothers me. Why don't you use a instance of ordered dict as an attribute on Python 2? Something like:

```python
class Metadata(object):
    def __init__(self, *args, **kwargs):
        object.__setattr__(self, '__odict', OrderedDict(*args, **kwargs))

    def __getattr__(self, key):
        od = object.__getattribute__(self, '__odict')
        if hasattr(od, key):
            return getattr(od, key)
        else:
            try:
                return od[key]
            except KeyError:
                raise AttributeError(key)

    def __setattr__(self, key, value):
        od = object.__getattribute__(self, '__odict')
        od[key] = value

    def __delattr__(self, key):
        od = object.__getattribute__(self, '__odict')
        del od[key]

    def method_factory(name):
        fullname = '__%s__' % name
        odict_method = getattr(OrderedDict, fullname)
        def method(self, *args, **kwargs):
            od = object.__getattribute__(self, '__odict')
            return odict_method(od, *args, **kwargs)
        return method

    __getitem__ = method_factory('getitem')
    __setitem__ = method_factory('setitem')
    __delitem__ = method_factory('delitem')
    __contains__ = method_factory('contains')

    __dir__ = method_factory('dir')

    __eq__ = method_factory('eq')
    __ne__ = method_factory('ne')
    __ge__ = method_factory('ge')
    __gt__ = method_factory('gt')
    __le__ = method_factory('le')
    __lt__ = method_factory('lt')

    __iter__ = method_factory('iter')
    __len__ = method_factory('len')

    __reduce__ = method_factory('reduce')
    __reduce_ex__ = method_factory('reduce_ex')

    __reversed__ = method_factory('reversed')

    __sizeof__ = method_factory('sizeof')

    def __repr__(self):
        return '\n'.join(['{}: {}'.format(k, v) for k, v in self.items()])
```
TODO: 
* the binary op methods do not work (because it translates `Metadata().__eq__(Metadata())` to `Metadata().__odict.__eq__(Metadata())` and OrderedDict does not understand Metadata objects (this is simple to fix though).
* `__dir__` should be overridden to return keys too (on Python3 too)
* sizeof, reduce and reduce_ex are probably broken. This is not very important though and should not be too hard to fix.

larray/core/metadata.py
```diff
@@ -0,0 +1,124 @@
...
+    >>> import os.path
+    >>> from datetime import datetime
+    >>> from larray.core.array import ndtest
+    >>> tmpdir = getfixture('tmpdir')
+    >>> attrs = Metadata([('string', 'a string'), ('number', 5), ('float', 5.6), ('time', datetime(2017, 2, 10))])
+
+    Save/Load using HDF format
+
+    >>> from larray.util.misc import LHDFStore
+    >>> fname = os.path.join(tmpdir.strpath, 'metadata.h5')
```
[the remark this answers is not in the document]
alixdamman: unfortunately, you cannot import `getfixture` from pytest

larray/core/array.py
```diff
@@ -735,6 +741,27 @@ def __init__(self, data, axes=None, title=''):
         self.data = data
         self.axes = axes
         self.title = title
+        if attrs is None:
+            attrs = Metadata()
+        self.attrs = attrs
+
+    @ property
```
gdementen: why the space?

larray/core/array.py
```diff
@@ -735,6 +741,27 @@ def __init__(self, data, axes=None, title=''):
...
+        """Returns metadata of the array.
+
+        Returns
+        -------
+        OrderedDict:
+            Metadata of the array.
+        """
+        return self._attrs
+
+    @ attrs.setter
```
gdementen: idem

larray/tests/test_array.py
```diff
@@ -131,6 +155,19 @@ def test_getattr(self):
         with self.assertRaises(AttributeError):
             self.larray.geom
 
+    @pytest.mark.skipif(PY2, reason="Python 2.7")
```
gdementen: The reason is not clear enough. I don't understand why it needs to be skipped on python 2.7... Ah, maybe it's because of the broken `__eq__`. If this is the reason, wouldn't it be better to fix `__eq__` than skip the test?

larray/core/metadata.py
```diff
@@ -67,23 +35,78 @@ class Metadata(BaseMetadata):
...
+                od = object.__getattribute__(self, '__odict')
+                return odict_method(od, *args, **kwargs)
+            return method
+
+        __getitem__ = method_factory('getitem')
+        __setitem__ = method_factory('setitem')
+        __delitem__ = method_factory('delitem')
+        __contains__ = method_factory('contains')
+
+        __eq__ = method_factory('eq')
```
gdementen: AFAICT, eq (and other binary ops in fact) still needs to be fixed

larray/core/metadata.py
```diff
@@ -8,40 +8,8 @@
...
-    number: 5
-    float: 5.6
-    time: 2017-02-10 00:00:00
-    """
-    def __init__(self, *args, **kwargs):
-        super(BaseMetadata, self).__init__(*args, **kwargs)
-
-    def __repr__(self):
-        return '\n'.join(['{}: {}'.format(k, v) for k, v in self.items()])
+class BaseMetadata(object):
```
gdementen: Wouldn't it be cleaner to define an AttributeDict class (differently on PY2 and PY3) then have a single Metadata class (for both PY2 and PY3) inheriting from it?

larray/core/metadata.py
```diff
@@ -67,23 +35,78 @@ class Metadata(BaseMetadata):
...
+
+        __iter__ = method_factory('iter')
+        __len__ = method_factory('len')
+
+        __reduce__ = method_factory('reduce')
+        __reduce_ex__ = method_factory('reduce_ex')
+
+        __reversed__ = method_factory('reversed')
+
+        __sizeof__ = method_factory('sizeof')
```
gdementen: Did you test sizeof, reduce and reduce_ex? They are probably broken (sizeof probably underevaluates the size and reduce might not work). If they don't work completely, it's better to remove them.

larray/core/metadata.py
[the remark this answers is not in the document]
gdementen: doh! But AFAICT, you could use the join method to avoid importing os.path:
```python
with LHDFStore(tmpdir.join('metadata.h5').strpath) as store:`
   ...
```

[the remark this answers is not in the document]
alixdamman: OK. But keep in mind that I never plan to add the Metadata class to the API. The examples in metadata.py are only for developers.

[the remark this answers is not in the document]
gdementen: Make senses... But one should never say never 😄.

larray/tests/test_array.py
alixdamman: The reason is that all of our current users have Python 3 on their computers and Python 2 will be abandon soon. Implementing an AttributeDict class in Python 3 is easy while the implementation for Python 2 is ugly. My first plan was to not implement attributes for Python 2 or using a simple OrderedDict.

larray/core/metadata.py
alixdamman: IO functions will be the same for PY2 and PY3

gdementen: yes, and?

larray/tests/test_array.py
gdementen: We will drop Python2 as soon as possible. Until then, it would make support (of LIAM2) harder if we have different functionality on Python2.

larray/core/metadata.py
alixdamman: we will have duplicated code

gdementen: no, since the io code would only be in the Metadata class.

larray/core/metadata.py
```diff
@@ -0,0 +1,143 @@
...
+
+from collections import OrderedDict
+from larray.util.misc import PY2
+
+
+__all__ = ['Metadata']
+
+
+if PY2:
+    class BaseMetadata(object):
```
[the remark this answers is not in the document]
alixdamman: Well, [AttrDict](https://github.com/bcj/AttrDict) already exists. OK for AttributeDict.

larray/tests/test_array.py
```diff
@@ -61,6 +72,56 @@ def test_slice_strings(self):
...
+    with open(fname, 'rb') as f:
+        attrs2 = Metadata(OrderedDict(pickle.load(f)))
+    assert attrs2 == attrs
+
+
+def test_metadata_hdf(attrs, tmpdir):
+    key = 'meta'
+    fname = os.path.join(tmpdir.strpath, 'test_metadata.hdf')
+    with LHDFStore(fname) as store:
+        ndtest(3).to_hdf(store, key)
```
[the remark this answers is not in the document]
alixdamman: Because `Metadata.to_hdf` uses the standard attribute [`attrs`](https://www.pytables.org/usersguide/tutorials.html#setting-and-getting-user-attributes) associated to each node object. 
So, I first need to store an arbitrary object and then call 
```python 
  return hdfstore.get_storer(key).attrs.metadata
``` as in `Metadata.to_hdf`.

PR #104, opened 2017-02-10
Add install instructions in doc (issue #101)

alixdamman:
Instructions for installation directly written in README.rst and then included in install.rst

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -54,8 +54,8 @@ Required Dependencies
 ---------------------
 
 - Python 2.7, 3.4, 3.5, or 3.6
-- `numpy <http://www.numpy.org/>`__ (1.10 or later)
-- `pandas <http://pandas.pydata.org/>`__ (0.15.0 or later)
+- `numpy <http://www.numpy.org/>`__ (1.10.0 or later)
+- `pandas <http://pandas.pydata.org/>`__ (0.13.1 or later)
 
 
 Optional Dependencies
@@ -112,7 +112,4 @@ If larray has been installed through larrayenv,
 you simply must do ::
 
     conda update larrayenv
-
-# This will update larray and all of its dependencies
-# at the same time.
 
--- a/README.rst
+++ b/README.rst
@@ -1,7 +1,120 @@
-larray
+LArray
 ======
 
 larray provides a Labelled Array class
 
 .. image:: https://travis-ci.org/liam2/larray.svg?branch=master
     :target: https://travis-ci.org/liam2/larray
+
+
+.. start-install
+
+Installation
+============
+
+Pre-built binaries
+------------------
+
+The easiest route to installing larray is through
+`Conda <http://conda.pydata.org/miniconda.html>`_.
+For all platforms installing larray can be done with::
+
+    conda install -c gdementen larray
+
+This will install a lightweight version of larray
+depending on Nympy and Pandas libraries only.
+This version does not allow you to use the included
+graphical user interface, to make plots or to use
+special I/O functions for easy dump/load from Excel
+or HDF files. Optional dependencies are described
+bellow.
+
+Installing larray with all optional dependencies
+can be done with ::
+
+    conda install -c gdementen larrayenv
+
+
+Building from source
+--------------------
+
+The latest release of LArray is available from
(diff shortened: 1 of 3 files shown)


gdementen:
Looks nice. A few comments though:

* I would mention "conda config --add channels gdementen" (in addition to supplying the "-c" argument everytime).
* did you test building form source using "python setup.py install"? I haven't for ages, so I don't know if it still works.
* I always thought build_ext was for C/cython/binary extensions, which LArray does not contain so far. Does it really add any value for a python-only module?
* the line wrapping seems odd.

README.rst
```diff
@@ -1,7 +1,120 @@
...
+
+For non-standard locations, additional build lib & include paths
+can be provided as per-usual at build_ext phase::
+
+    python setup.py build_ext -I/path/to/include -L/path/to/lib
+    python setup.py install
+
+
+Update
+------
```
gdementen: shouldn't this be on the same level than Install (and thus after dependencies)?

README.rst
```diff
@@ -1,7 +1,120 @@
...
+------------------
+
+The easiest route to installing larray is through
+`Conda <http://conda.pydata.org/miniconda.html>`_.
+For all platforms installing larray can be done with::
+
+    conda install -c gdementen larray
+
+This will install a lightweight version of larray
+depending on Nympy and Pandas libraries only.
```
gdementen: Numpy

README.rst
```diff
@@ -1,7 +1,120 @@
...
+This version does not allow you to use the included
+graphical user interface, to make plots or to use
+special I/O functions for easy dump/load from Excel
+or HDF files. Optional dependencies are described
+bellow.
```
gdementen: below

README.rst
gdementen: depending only on...

README.rst
```diff
@@ -1,7 +1,120 @@
...
+
+The easiest route to installing larray is through
+`Conda <http://conda.pydata.org/miniconda.html>`_.
+For all platforms installing larray can be done with::
+
+    conda install -c gdementen larray
+
+This will install a lightweight version of larray
+depending on Nympy and Pandas libraries only.
+This version does not allow you to use the included
```
gdementen: Additional libraries are required to...

README.rst
```diff
@@ -1,7 +1,120 @@
...
+graphical user interface, to make plots or to use
```
gdementen: graphical user interface, make plots or use

README.rst
```diff
@@ -1,7 +1,120 @@
...
+can be provided as per-usual at build_ext phase::
+
+    python setup.py build_ext -I/path/to/include -L/path/to/lib
+    python setup.py install
+
+
+Update
+------
+
+If larray has been installed through larray package,
```
gdementen: through/via conda might be clearer than via larray package.

README.rst
```diff
@@ -1,7 +1,120 @@
...
+
+Be careful if you have installed optional dependencies.
+In that case, you may have to update some of them.
+
+If larray has been installed through larrayenv,
+you simply must do ::
+
+    conda update larrayenv
+
+This will update larray and all of its dependencies
```
gdementen: it will not update the dependencies at this point (it was the idea, but it does not seem to work yet). I guess we will have to fix each dependency version in the metapackage for this to actually be the case.

README.rst
```diff
@@ -1,7 +1,120 @@
...
+at the same time.
+
+
+Required Dependencies
+---------------------
+
+- Python 2.7, 3.4, 3.5, or 3.6
+- `numpy <http://www.numpy.org/>`__ (1.10 or later)
+- `pandas <http://pandas.pydata.org/>`__ (0.15.0 or later)
```
gdementen: where does the 0.15 come from? I don't really have a precise idea, which version we require. If you tested that, we need to update setup.py and meta.yml to be consistent.

alixdamman:
Yes, "python setup.py install" works as long as you have a C++ AND a fortran compiler (to install Numpy).

gdementen:
LGTM

PR #111, opened 2017-02-13
fix #67 : add replace_axes method to LArray class.

### LINKED ISSUE #67: implement replace_axis
to complement with_axes, we need a way to replace only one axis (or a few axes). The goal is to have something nicer than:
```python
a.with_axes(a.axes.replace(x.products, industries))
```
e.g:
```python
a.replace_axis(x.products, industries)
# or
a.with_axes(products=industries)
```

alixdamman:
with signature similar to rename method (to stay consistent).

### DIFF
--- a/larray/core.py
+++ b/larray/core.py
@@ -9847,7 +9847,7 @@ def ndtest(shape, start=0, label_start=0, title='', dtype=int):
                     for length in a.shape]
     new_axes = [Axis(name, [name + str(i) for i in label_range])
                 for name, label_range in zip(axes_names, label_ranges)]
-    return a.with_axes(new_axes)
+    return LArray(a.data, new_axes)
 
 
 def kth_diag_indices(shape, k):
--- a/larray/core.py
+++ b/larray/core.py
@@ -3884,9 +3884,76 @@ class LArray(object):
         #  can do a[a.nonzero()]
         return self.data.nonzero()
 
+    def replace_axes(self, axes_to_replace=None, new_axes=None, **kwargs):
+        """
+        Returns an array with one or several axes replaced.
+
+        Parameters
+        ----------
+        axes_to_replace : axis ref or dict {axis ref: axis} or
+                  list of tuple (axis ref, axis)
+            Axes to replace. If a single axis reference is given,
+            the `new_axes` argument must be used.
+        new_axes : Axis
+            New axis if `axes_to_replace`
+            contains a single axis reference
+        **kwargs : Axis
+            New axis for each axis to replace given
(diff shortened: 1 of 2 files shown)


alixdamman:
OK, the case where we pass a list wasn't tested in docttest of rename. 
The error here is with Python 2.7 : 
'list' object has no attribute 'copy'
I need to change implementation of rename and replace_axes

gdementen:
I **wonder** (not sure at all) whether we should add the case where a single list of Axis is passed to replace_axes/rename for the case where we want to replace all axes, in which case, we could deprecate with_axes. The thing is that when having replaces_axes, it **seems** like with_axes is one too many method. What do you think?

alixdamman:
I would go for setting with_axes as depecrated.

gdementen:
We still need a credible upgrade path. Previously with_axes was mostly used to align axes with those of another array: `arr1.with_axes(arr2.axes)`. The new syntax seems to be annoying when you want to replaces all axes, or do you see something better than:
```python
arr1.replace_axes(zip(arr1.axes, arr2.axes))
```
Hence I wanted to allow axes_to_replace to be a list of Axis in addition to a list of tuples. In that case the list must be exactly the same length as self.axes. As I said I am unsure it is a good idea, but deprecating with_axes without providing a nice syntax for that use case seems like a bad move.

Maybe allowing new_axes to be a list (by the name of the argument one might think it is allowed anyway), and using: 
```python
arr1.replace_axes(arr1.axes, arr2.axes)
```
would be good enough (honestly, I do not really like it)?

alixdamman:
I chose to have a function signature as close as possible to one of rename method but you get a point here. 
I agree to allow axes_to_replace to be a list of Axis and to check its length in that case.

larray/core.py
```diff
@@ -3884,22 +3884,34 @@ def nonzero(self):
...
-        (int, str or  Axis), optional
-            New axes.
+        axes_to_replace : axis ref or dict {axis ref: axis} or
+                  list of tuple (axis ref, axis) or list of Axis
+            Axes to replace. If a single axis reference is given,
+            the `new_axes` argument must be used. If a list of
+            Axis is given, all axes will be replaced by the
+            new ones. In that case, the number of new axes must
+            match the number of the old ones.
+        new_axes : Axis
```
gdementen: we should either rename this to new_axis, or allow a list of Axis

alixdamman: 'new_axis' it will be.

gdementen:
LGTM, merging...

PR #118, opened 2017-02-13
fix #106 : add regex option in split_axis method

### LINKED ISSUE #106: Allow to use regex in split_axis

alixdamman:


larray/core.py
```diff
@@ -3179,6 +3179,10 @@ def split_axis(self, axis, sep='_', names=None):
         names : list of names, optional
             names of resulting axes. Defaults to split the combined axis name
             using the given delimiter string.
+        regex : str, optional
+            use regex instead of simple delimiter to split the axis.
```
gdementen: the axis labels?

larray/core.py
```diff
@@ -8781,8 +8798,13 @@ def split_axis(self, axis, sep='_', names=None):
         a\\b | b0 | b1 | b2
          a0 |  0 |  1 |  2
          a1 |  3 |  4 |  5
+        >>> regex = '([a-z0-9]+).([a-z0-9]+)'
```
gdementen: I would have hoped for a more interesting example (which cannot work using the usual sep argument). And if using "combined" as starting point, I would use "(\w+)_(\w+)" as the regex (and avoid defining it in a separate variable).
An example I would find more interesting is the same labels than combined but without the delimiter:
a0b0, a0b1 etc (assuming < 9 labels in each dim ;-)).
```python
new = combined.set_labels(x.a_b, [l.replace('_', '') for k in combined.a_b.labels])
```
or (maybe clearer) set the labels explicitly:
```python
new = combined.set_labels(x.a_b, ['a0b0', 'a0b1', 'a0b2', 'a1b0', 'a1b1', 'a1b2'])
```
PS: we should allow this syntax in set_labels (to be consistent with array creation funcs):
```python
new = combined.set_labels(x.a_b, 'a0b0,a0b1,a0b2,a1b0,a1b1,a1b2')
```
I'll open another issue for this.

gdementen: In fact, a single string in set_labels already works (see #121).

larray/core.py
```diff
@@ -8761,6 +8774,10 @@ def split_axis(self, axis, sep='_', names=None):
         names : list of names, optional
             names of resulting axes. Defaults to split the combined axis name
             using the given delimiter string.
+        regex : str, optional
+            use regex instead of simple delimiter to split the axis.
+            Names of new axes must be provided.
```
gdementen: We could possibly lift the names must be provided restriction, as there could be case were the normal sep is valid for the axis name but not for the axis labels (such as in the example).

larray/core.py
alixdamman: Indeed

larray/core.py
alixdamman: Regex for labels and delimiter for names if `names` is string or axis name if `names` is None.

gdementen:
LGTM, merging...

PR #124, opened 2017-02-15
fix #122 : update __matmul__ implementation (matrices with dim != 2)

### LINKED ISSUE #122: implement arr2d @ arr1d
currently the workaround is to:
```python
(arr2d @ arr1d.expand(Axis('bidulle', 1)))[x.bidulle[0]]
```

alixdamman:


larray/core.py
```diff
@@ -7667,26 +7667,67 @@ def opmethod(self, other):
...
+          to its dimensions. After matrix multiplication the prepended 1 is removed.
+        * If the second argument is 1-D, it is promoted to a matrix by appending a 1
+          to its dimensions. After matrix multiplication the appended 1 is removed.
+
+        For ND-matrices with N > 2:
+        a @ b[i,j,k,m] = sum(a[i,j,:] * b[k,:,m])
+
+        Examples
+        --------
+        >>> arr1d = ndtest((3))
```
gdementen: ndtest(3) would be nicer

larray/core.py
```diff
@@ -7667,26 +7667,67 @@ def opmethod(self, other):
     __ror__ = _binop('ror')
 
     def __matmul__(self, other):
+        """
+        What doc of Numpy matmul says:
```
gdementen: This sentence looks odd, though I have trouble to find a better phrasing. "Quoting the corresponding documentation in Numpy"?

alixdamman:
OK. @ only works fro python >= 3.5 and I need to update test_matmul.

larray/core.py
```diff
@@ -7667,26 +7667,67 @@ def opmethod(self, other):
...
+            other = LArray(other)
+        other_axes = other.axes.copy()
+        # Numpy can handle matrices with any dimensions:
+        # a @ b[i,j,k,m] = sum(a[i,j,:] * b[k,:,m])
+        res_data = self.data.__matmul__(other.data)
+        res_axes = self.axes[:-2] + other_axes[:-2]
+        if self.ndim > 1:
+            res_axes += self.axes[-2]
+        if other.ndim > 1:
+            res_axes += other_axes[-1]
```
gdementen: the way to compute res_axes seems wrong... or, when I think of it... right but incoherent with res_data.
`+ other_axes[:-2]` will introduce more axes if they are different from `self.axes[:-2]`, which is great but that is not the behaviour you will get for res_data.

larray/core.py
```diff
@@ -7667,26 +7667,67 @@ def opmethod(self, other):
...
+        >>> arr1d @ arr3d
+        c\\e | e0 | e1 | e2
+         c0 | 15 | 18 | 21
+         c1 | 42 | 45 | 48
+         c2 | 69 | 72 | 75
+        >>> arr3d @ arr1d
+        c\\d | d0 | d1 | d2
+         c0 |  5 | 14 | 23
+         c1 | 32 | 41 | 50
+         c2 | 59 | 68 | 77
```
gdementen: adding an example of 2d @ 3d or 3d @ 3d could be interesting I think.

larray/core.py
gdementen: to be clear, that means you have to expand & potentially transpose both arrays manually before doing the matmul. Depending on whether numpy supports broadcasting size 1 axes for the "stack" axes (ie not the 2 last), you would do that using something like:
```python
# if numpy does support...
combined_axes = self.axes[:-2] + other.axes[:-2]
self_exp = self.broadcast_with(combined_axes).transpose(combined_axes)
other_exp = other.broadcast_with(combined_axes).transpose(combined_axes)
```
```python
# or... if numpy does not support...
combined_axes = self.axes[:-2] + other.axes[:-2]
self_exp = self.expand(combined_axes).transpose(combined_axes)
other_exp = other.expand(combined_axes).transpose(combined_axes)
```
Now, it might be tricky to get right when there is an ndim == 1 on either side.
FWIW, to fix the test case (anonymous axes) on the 2d @ 2d case, you could write stuff like:
```python
        if self.ndim > 1 and other.ndim > 1:
            res_axes += [self.axes[-2], other_axes[-1]]
```
or you could fix AxisCollection += for anonymous axes but that part of the code is a real snake nest (it is tricky to get right in all situations).

larray/core.py
```diff
@@ -7667,26 +7667,67 @@ def opmethod(self, other):
...
-        # TODO: implement for matrices of any dimensions
-        if self.ndim != 2 or other.ndim != 2:
-            raise NotImplementedError("matrix multiplication not "
-                                      "implemented for ndim != 2 (%d @ %d)"
-                                      % (self.ndim, other.ndim))
-        # 4,2 @ 2,3 = 4,3
-        res_axes = [get_axis(self, 0), get_axis(other, 1)]
-        # TODO: implement AxisCollection.__init__(copy_duplicates=True)
-        if res_axes[1] is res_axes[0]:
-            res_axes[1] = res_axes[1].copy()
```
gdementen: those two lines were there for a reason: if you have twice the same Axis object in the result, it breaks. The most common case where this can happen is arr @ arr.T (for which there should have been a test, but well...)

alixdamman: I already made some changes yesterday that I didn't push yet. Is there anyway to allow to run doctest for some python versions only ?

larray/core.py
alixdamman: Well, my first intention was not to write a real documentation but just to add some indications to remember how __matmul__ from Numpy works. I'll move that part inside the function.

larray/core.py
gdementen: Not that I know of. The unit test could (should? :)) be transformed to use skipIf but for the doctest, I guess you have do the check in the test itself. Kinda ugly but...

larray/core.py
alixdamman: This is not the only problem. Anonymous axes are annoying... (as in unit tests).

larray/core.py
```diff
@@ -7668,65 +7668,93 @@ def opmethod(self, other):
...
+
+        Notes
+        -----
+        Only available with Python >= 3.5
 
         Examples
         --------
-        >>> arr1d = ndtest((3))
-        >>> arr1d
+        >>> if sys.version_info >= (3,5):
```
gdementen: Given the buildbot failing, I wonder if it wouldn't be better to transform those tests to unit tests and either use 
`# doctest: +SKIP` here or even kill these doctests entirely....

larray/core.py
```diff
@@ -7724,6 +7716,7 @@ def __matmul__(self, other):
         c2 |  d0 | 1203 | 1260 | 1317
         c2 |  d1 | 1392 | 1458 | 1524
         c2 |  d2 | 1581 | 1656 | 1731
+        Hello
```
gdementen: ? :)

gdementen:
LGTM

PR #127, opened 2017-02-16
improve matmul tests and use unittest.skipIf

gdementen:
improve matmul tests and use unittest.skipIf


alixdamman:
OK to merge.

gdementen:
changes already contained in #129

PR #135, opened 2017-02-24
fix #131 : convert data to given dtype or float in from_string function.

### LINKED ISSUE #131: Fix from_string + add unit test

alixdamman:


gdementen:
I would rather have the types autodetected, like in read_csv. Using the StringIO hack I told you about earlier is probably the easiest way around this.

gdementen:
I wonder if we couldn't support the doctests simpler/faster by using `skipinitialspace=True` argument to pd.read_csv, instead of stripping each cell and rebuild a string out of that. Granted, the trailing whitespace wouldn't be removed, but that's not as important, right?

larray/core.py
```diff
@@ -9172,9 +9212,25 @@ def from_string(s, nb_index=None, index_col=None, sep=','):
...
+        index_col = [index_col]
+
+    if nb_index is None and index_col is None:
+        try:
+            header = s.strip().splitlines()[0]
+            pos_last = next(i for i, v in enumerate(header.split(sep)) if
+                            isinstance(v, basestring) and '\\' in v)
+        except StopIteration:
+            pos_last = 0
+        index_col = [pos_last]
```
gdementen: this does not seem correct if ndim > 2. I think it should be index_col = list(range(pos_last))
OR you could move the check for nb_index None & index_col None before the other checks and specify nb_index = pos_last. Should also add a test for 1D array because I think this is broken too.

larray/core.py
```diff
@@ -9172,9 +9212,25 @@ def from_string(s, nb_index=None, index_col=None, sep=','):
...
+
+    df = pd.read_csv(StringIO(s), index_col=index_col, sep=sep, skipinitialspace=True, **kwargs)
+    return df_aslarray(df)
```
gdementen: In fact, I think the whole function could be written as:

    return read_csv(StringIO(s), nb_index=nb_index, index_col=index_col, sep=sep, skipinitialspace=True, **kwargs)

larray/core.py
```diff
@@ -114,6 +114,12 @@
                           ReprString, duplicates, array_lookup2, strip_rows,
                           skip_comment_cells, find_closing_chr, PY3)
 
+if PY3:
+    from io import StringIO
+else:
+    from StringIO import StringIO
```
gdementen: wouldn't it be better to move this to utils?

larray/core.py
```diff
@@ -9139,14 +9145,48 @@ def from_string(s, nb_index=None, index_col=None, sep=','):
...
+            Function to use for converting a sequence of string columns to an array of
+            datetime instances. The default uses ``dateutil.parser.parser`` to do the
+            conversion. Pandas will try to call date_parser in three different ways,
+            advancing to the next if an exception occurs: 1) Pass one or more arrays
+            (as defined by parse_dates) as arguments; 2) concatenate (row-wise) the
+            string values from the columns defined by parse_dates into a single array
+            and pass that; and 3) call date_parser once for each row using one or more
+            strings (corresponding to the columns defined by parse_dates) as arguments.
+
+        See other arguments of pandas read_csv method.
```
gdementen: generally, I don't like this way of documenting kwargs. I would rather have them either as explicit kwargs in the function signature OR not document them and say that all extra kwargs are forwarded to pandas read_csv and invite people to look at that.

larray/core.py
```diff
@@ -9162,26 +9161,42 @@ def from_string(s, nb_index=None, index_col=None, sep=',', dtype=float):
...
     nat\sex | M | F
          BE | 0 | 1
          FO | 2 | 3
     """
-    data = [[cell.strip() for cell in line.split(sep)]
-            for line in s.strip().splitlines()]
-    res = from_lists(data, nb_index=nb_index, index_col=index_col)
-    return res.astype(dtype)
+
+    if nb_index is None and index_col is None:
```
gdementen: why do we need this at all? The autodetection in read_csv fails?

PR #138, opened 2017-03-06
fix #136 : plot method in viewer displays the legend with correct labels

### LINKED ISSUE #136: viewer plot legend is wrong when selecting arbitrary rows
The N first labels are used. If the rows selected are not at the start of the array, that is inconsistent.

alixdamman:
+ let Matplotlib to choose where to put legend.


gdementen:
Out of curiosity, why did you change your mind about the legend position?

alixdamman:
Because, most of the time, we only see half of the legend on screen. I prefer to find a better solution later.

gdementen:
LGTM

PR #144, opened 2017-03-10
fix xticks in plot viewer when zooming (issue #143)

alixdamman:


gdementen:
this only needs a changelog entry (once #145 is merged).

PR #147, opened 2017-03-13
Rename replace_axes as set_axes + update doc of rename method

### LINKED ISSUE #121: document that a single string is allowed in set_labels
allow stuff like:

a.set_labels(x.b, 'B0..B2')
a.set_labels(x.b, 'v1,v2,v3')

In fact, it works already, but this is not documented, so we should update the docstring.

### LINKED ISSUE #133: rename replace_axes to set_axes and update set_labels API?
I wonder if we shouldn't make set_labels, replace_axes and rename as close to each other as possible.
In this regard, I think renaming replace_axes to set_axes would be a good idea (bonus points for being shorter without being less clear IMO).

alixdamman:
fix #121 : Update docstring of set_labels.

fix #133 : Rename replace_axes as set_axes.

Add inplace flag to set_axes and rename methods

### DIFF
--- a/larray/core.py
+++ b/larray/core.py
@@ -3905,7 +3905,7 @@ class LArray(object):
                 raise ValueError("length of axes %s does not match "
                                  "data shape %s" % (axes.shape, data.shape))
 
-        # Because __getattr__ and __setattr__ have been rewritten
+        # Because __getattr__ and __setattr__ have been overridden
         object.__setattr__(self, 'data', data)
         object.__setattr__(self, 'axes', axes)
         object.__setattr__(self, 'title', title)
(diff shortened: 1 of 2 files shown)


larray/core.py
```diff
@@ -4319,6 +4328,10 @@ def rename(self, renames=None, to=None, **kwargs):
         nat2\\sex2 | M | F
                BE | 0 | 1
                FO | 2 | 3
+        >>> arr.rename({'nat': 'nat2', 'sex': 'sex2'})
```
gdementen: duh? this is the same as the previous example?

larray/core.py
```diff
@@ -3905,7 +3905,7 @@ def __init__(self,
                 raise ValueError("length of axes %s does not match "
                                  "data shape %s" % (axes.shape, data.shape))
 
-        # Because __getattr__ and __setattr__ have been rewritten
+        # Because __getattr__ and __setattr__ have been overridden
```
gdementen: even though it is indeed the case, `__setattr__` should actually not be overridden (ever since we changed `__getattr__` to return axes). It might be better to fix that in a separate PR though, because most (all?) `object.__setattr__` could/should be replaced.

larray/core.py
```diff
@@ -3976,28 +3973,37 @@ def replace_axes(self, axes_to_replace=None, new_axis=None, **kwargs):
          a1 |  3 |  4 |  5
         >>> row = Axis('row', ['r0', 'r1'])
         >>> column = Axis('column', ['c0', 'c1', 'c2'])
-        >>> arr.replace_axes(x.a, row)
+
+        Replace one axis (expect 2 arguments)
```
gdementen: the "expect 2 arguments" part is a bit strange. who should expect? the function or the user? => replace by somethng like:
Replace one axis (the second argument is required when using this syntax)

larray/core.py
alixdamman: It was to test the inplace flag. I failded to properly clean the examples.

gdementen: :)

gdementen:
LGTM

Issue #148, opened 2017-03-14
deprecate Axis.group


gdementen:
I think it is not worth having two ways to create groups. It used to be necessary because there was no way to easily set a name to a group after the fact, but now that we have >>, I think we should only support creating groups via axis[]

PR #150, opened 2017-03-14
Set Axis.group as deprecated (issue #148)

alixdamman:
+ update unit tests.

### DIFF
--- a/larray/core.py
+++ b/larray/core.py
@@ -1101,7 +1101,12 @@ class Axis(object):
         return LGroup(key, name, self)
 
     def group(self,  *args, **kwargs):
-        raise NotImplementedError("Axis.group is deprecated. Use the syntax \"age[10:19] >> 'teenagers'\" instead.")
+        group_name = kwargs.pop('name', None)
+        key = args[0] if len(args) == 1 else args
+        syntax = '{}[{}]'.format(self.name if self.name else 'axis', key)
+        if group_name is not None:
+            syntax += ' >> {}'.format(repr(group_name))
+        raise NotImplementedError('Axis.group is deprecated. Use {} instead.'.format(syntax))
 
     def all(self, name=None):
         """
@@ -1116,7 +1121,10 @@ class Axis(object):
         --------
         Axis.group
         """
-        return self._group(slice(None), name=name if name is not None else "all")
+        axis_name = self.name if self.name else 'axis'
+        group_name = name if name else 'all'
+        raise NotImplementedError('Axis.all is deprecated. '
+                                  'Use {}[:] >> {} instead.'.format(axis_name, repr(group_name)))
 
     def subaxis(self, key, name=None):
         """
--- a/larray/core.py
+++ b/larray/core.py
@@ -1046,8 +1046,7 @@ class Axis(object):
     def iswildcard(self):
         return self._iswildcard
 
-    # XXX: not sure I should offer an *args version
-    def group(self, *args, **kwargs):
+    def _group(self, *args, **kwargs):
         """
         Returns a group (list or unique element) of label(s) usable in .sum or .filter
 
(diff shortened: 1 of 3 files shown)


larray/core.py
```diff
@@ -1101,6 +1100,9 @@ def group(self, *args, **kwargs):
             return key.__class__(key.key, name, self)
         return LGroup(key, name, self)
 
+    def group(self,  *args, **kwargs):
+        raise NotImplementedError("Axis.group is deprecated. Use the syntax \"age[10:19] >> 'teenagers'\" instead.")
```
gdementen: it would be nicer for the user to use the actual key/axis he used in the error message instead of age[10:19]

larray/core.py
```diff
@@ -1101,6 +1100,9 @@ def group(self, *args, **kwargs):
...
+
     def all(self, name=None):
```
gdementen: we might want to deprecate this one too?
```python
>>> arr[:] >> 'Belgium'
# is almost as short as:
>>> arr.all('Belgium')

# and

>>> arr[:]
# is shorter than
>>> arr.all()

# The only significant difference is if we want exactly the name 'all'

>>> arr[:] >> 'all'
# is longer than
>>> arr.all()

# but I have not seen it much used in the wild, so my guess is that it is not worth it have this extra function.
```

larray/core.py
```diff
@@ -1335,7 +1337,7 @@ def __getitem__(self, key):
         """
         key is a label-based key (slice and fancy indexing are supported)
         """
-        return self.group(key)
+        return self._group(key)
```
gdementen: even if we do not deprecate all(), I would rather move the functionality here in getitem (and call getitem from _group) than the opposite. It's a good practice to make the deprecated function call the non-deprecated one.

larray/tests/test_la.py
```diff
@@ -145,16 +145,16 @@ def test_equals(self):
     def test_group(self):
```
gdementen: this test is now at least partially redundant with the test_getitem test below, both should be merged (or this one should disappear if it does not bring anything different).

larray/core.py
```diff
@@ -1079,29 +1066,22 @@ def _group(self, *args, **kwargs):
...
-                res = [k.with_axis(self) for k in key]
-                res = tuple(res) if isinstance(key, tuple) else res
-                return res
-
-        if isinstance(key, Group):
-            name = name if name is not None else key.name
-            return key.__class__(key.key, name, self)
-        return LGroup(key, name, self)
+        if isinstance(key, Group) and name is not None:
+            key.name = name
```
gdementen: we should almost never modify function arguments inplace

gdementen: and given the code below, I think it is useless anyway.

larray/core.py
```diff
@@ -1079,29 +1066,22 @@ def _group(self, *args, **kwargs):
...
+        if name is not None:
+            return self[key] >> name
+        else:
+            return self[key]
```
gdementen: I would probably use the ternary operator here: return ... if ... else ...

larray/tests/test_la.py
```diff
@@ -195,16 +195,6 @@ def test_match(self):
...
 
-    def test_getitem(self):
-        age = Axis('age', '0..10')
-        group = age[':3']
-        self.assertEqual(group.key, slice(None, 3, None))
-        self.assertTrue(group.axis.equals(age))
-
-        group = age[:]
-        self.assertEqual(group.key, slice(None))
-        self.assertIs(group.axis, age)
```
gdementen: this test is slightly different from the one in the "new" test_getitem above, so should be moved there instead of being removed (ie. an explicit test for non-string slices) and/or a test for a slice(None) is nice to have.

larray/tests/test_la.py
```diff
@@ -233,17 +210,17 @@ def test_positional(self):
 
     def test_all(self):
         age = Axis('age', '..115')
-        group = age.all()
+        group = age[:] >> 'all'
         self.assertEqual(group.key, slice(None))
```
gdementen: move this test to test_getitem, it no longer justifies a separate method

alixdamman:
Update

PR #163, opened 2017-03-17
Update documentation of read_csv/tsc/eurostat (Issue 162)

alixdamman:


larray/core.py
```diff
@@ -9265,8 +9265,8 @@ def read_csv(filepath, nb_index=None, index_col=None, sep=',', headersep=None, n
 
     Parameters
     ----------
-    filepath : str
-        Path where the csv file has to be read.
+    filepath : str or any object with a read() method
+        Path where the csv file has to be read or a file handle or a StringIO object.
```
gdementen: unsure we should explicitly mention StringIO, because that is just one kind of file handle, right?

larray/core.py
```diff
@@ -9265,8 +9265,8 @@ def read_csv(filepath, nb_index=None, index_col=None, sep=',', headersep=None, n
 
     Parameters
     ----------
-    filepath : str
-        Path where the csv file has to be read.
+    filepath : str or any object with a read() method
```
gdementen: "any file-like object" instead of "any object with a read() method"???

larray/core.py
```diff
@@ -9358,44 +9358,44 @@ def read_csv(filepath, nb_index=None, index_col=None, sep=',', headersep=None, n
...
 
-def read_hdf(filepath, key, na=np.nan, sort_rows=False, sort_columns=False, **kwargs):
+def read_hdf(path_or_buf, key, na=np.nan, sort_rows=False, sort_columns=False, **kwargs):
     """Reads an array named key from a HDF5 file in filepath (path+name)
 
     Parameters
     ----------
-    filepath : str
-        Filepath and name where the HDF5 file is stored.
+    path_or_buf : str or buffer
```
gdementen: the argument name should be coherent. Either filepath_or_buffer everywhere or path_or_buf everywhere (or path_or_buffer). I do not really care either way.

larray/core.py
alixdamman: Taken from doc of pandas.read_csv but I agree with you. I'll change that.

alixdamman:
update

larray/core.py
gdementen: but you did not?

alixdamman: Now it should be OK.

PR #168, opened 2017-03-21
add readthedocs.yml and environment.yml files

alixdamman:
Inspired by the Github repository of Xarray when I tried to understand why the build of Larray doc crashed on Read the Docs. 

The error message was
sphinx.errors.ExtensionError: Could not import extension numpydoc (exception: No module named 'numpydoc')

I don't want to add numpydoc as a requirement in setup.py or to write a requirements.txt file. 
I find using readthedocs.yml cleaner:
http://docs.readthedocs.io/en/latest/yaml-config.html


alixdamman:
Made a test based on my branch 'add_readthedocs_yaml_file'. Result is here:
http://larray-test.readthedocs.io/en/add_reathedocs_yaml_file/intro.html

alixdamman:
Review requested. 
Need help to organize the subsections of section LArray in api.rst

gdementen:
This is GREAT! Nice work. There are a lot of stuff I would like to improve, but we need to give something to our users today. So I would do only a few minor edits, send it to our users but tell them to refrain from printing since they will get a new version by the end of next week.

gdementen:
as for the edits:
* add your name to the authors (I thought I had, but it seems like it's not the case).
* kill the introduction (or complete it, but I guess killing it is easier for now) because it is a bit ridiculous in its current state.
* move update section within install chapter. A chapter for that is a bit overkill.
* add doc for XXX_by (it does not seem to be included?) and fix the links to them (probably happens automatically if the doc is present).
* add the XXX_by methods to the "aggregation functions" section

alixdamman:
Thanks for review. I'm working on it this afternoon.

gdementen:
as for the ordering of the LArray subsections, I would put the array creation functions before or after the LArray methods, not in the middle of them. Except than that, it seems fine to me in the short term.

gdementen:
... and put the Input/Output section before the Sessions section

gdementen:
... and also put the Misc section before the Sessions section

gdementen:
and if you can, have links to each subsection of LArray in either the start of the LArray section and/or directly in the TOC at the start of the doc (preferably at both places)

alixdamman:
Names of the authors are given at the bottom of the page.

alixdamman:
I will merge this PR today. Do you have any other comments ?

gdementen:
concerning authors. I meant on the front page of the PDF version

gdementen:
a last comment (but not that important), the changelog order should be reversed (most recent version first), like is the custom.

alixdamman:
To be continued...

Issue #191, opened 2017-04-03
eradicate VG


gdementen:
There are still a few occurrence of VG appearing in the code, comments and in the tests. This was an acronym for ValueGroup, which was the first name for LGroup. At one point, those were called LabelGroup, which we might check too, but this is not as confusing.

FWIW, we cannot simply do a global replace to LGroup because that would pick up avg.

PR #197, opened 2017-04-03
fix #50 : PGroup --> IGroup

### LINKED ISSUE #50: position or index
We should pick one of the two terms and stick with it. Currently we use both (.i and .ipoints but PGroup and posarg*). We should either have:
.p[], .ppoints[], PGroup and posarg*
or
.i[], .ipoints[], IGroup and iarg* (or indarg*)

alixdamman:


larray/core.py
```diff
@@ -5713,7 +5712,7 @@ def _group_aggregate(self, op, items, keepaxes=False, out=None, **kwargs):
 
     def _prepare_aggregate(self, op, args, kwargs=None, commutative=False,
                            stack_depth=1):
-        """converts args to keys & VG and kwargs to VG"""
+        """converts args to keys & LGoup and kwargs to LGoup"""
```
gdementen: LGroup, not LGoup ;-) (this is not the only occurrence of this)

gdementen:
Oops, put my comment on the issue, not on the PR, sorry.

    I personally like this but I would have asked a few users what they think before implementing this, 
    because I wonder if positions is not more telling to a random user than indices??

larray/core.py
alixdamman: Oops, sorry. Changed.

gdementen:
In either case, I do not know if users used the name in their code or not. I think not, but we cannot be sure, so we might want to add a dummy PGroup object which spits an error.

larray/core.py
```diff
@@ -2243,6 +2243,23 @@ def opmethod(self, other):
     __sub__ = difference
 
 
+class PGroup(Group):
+    """Replaced by IGroup.
+    """
+
+    def translate(self, bound=None, stop=False):
```
gdementen: How about something like this instead?

```python
class PGroup(Group):
    """Replaced by IGroup.
    """
    def __init__(self, *args, **kwargs):
        raise NotImplementedError("PGroup has been replaced by IGroup")
```

alixdamman:
Do I need to add "fix #191 : replaced all occurences of `VG` by `LGroup`" to changelog?

gdementen:
nope, the changelog is for users.

alixdamman:
OK to merge?

gdementen:
Here is a summary concerning the "sorting" functions.

Previous:
  * sort_axis, sort_values
  * argmax, argmin, argsort
  * posargmax, posargmin, posargsort
Problems: "arg" isn't very intuitive (this has a historic reason, see below (*)), sort_values vs argsort (should be called argsort_values?)   

Current state of this PR:
  * sort_axis, sort_values
  * argmax, argmin, indirectsort
  * idxargmax, idxargmin, idxindirectsort
Potential problems: sort_values vs indirectsort (same as above), argmax vs indirectsort. 

Proposition 1:
  * sort_axis, sort
  * indirectmax, indirectmin, indirectsort
  * idxindirectmax, idxindirectmin, idxindirectsort
Potential problems: sort is a bit vague, idxindirect* are a bit long names and I am unsure indirect* is much more telling for our users than arg*. It is for me but...

Proposition 2:
  * sort_axis, sort
  * labelmax, labelmin, labelsort
  * idxmax, idxmin, idxsort

Potential problem: the problem with this proposition is the possible confusion between sort_axis and labelsort.

(*) for reference, argsort is mostly a numpy-only term. It was used to be consistent with argmin/argmax which are relatively standard. The arg comes from the fact that you can view an array as a function:
x -> f(x), where x is the position in the array and f(x) is the value. In that case, argmax, means the argument (ie indice) which maximize the function.

@alixdamman, @gbryon, @jehanneman, @jmpplan: any thoughts, on the above propositions or alternate propositions?

gdementen:
@alixdamman, @gbryon, @jehanneman, @jmpplan: any thoughts on https://github.com/liam2/larray/pull/197#issuecomment-294801199 ?

jehanneman:
An alternative for proposition 2 for the "sorting" functions:
LArray.sort_axis
LArray.sort_values
LArray.sort_labels
LArray.sort_index

LArray.max_labels
LArray.max_index
...
Its a bit longer but it has the advantage that it's consistent...

gdementen:
Thanks for your input.

changing idx to index, adding underscores and keeping the _values is fine by me (I thought you were the one who wanted shortest names possible ;-)), but I fear sort_labels and sort_index are even more confusing than label_sort or index_sort precisely because of the consistency with sort_axis and sort_values while they inherently do something different. Remember they return the labels (and respectively the indices) that would sort the array. The labels themselves are not sorted. To sort the labels (of the axes), you have to use sort_axis.

jehanneman:
Oow, ..., I confess I didn't know what sort_axis and the other functions really did, ..., my knowledge of LArray is still too poor. In that case keep the naming of proposition 2 I don't mind a bit longer names when it makes the usage of methodes or function more self explanatory. Only fools don't change their minds....

alixdamman:
To summary:

- sort_axis, sort_values                         --> sort_axis, sort
- argmax, argmin, argsort                    --> label_max, label_min, label_sort
- posargmax, posargmin, posargsort   --> index_max, index_min, index_sort

gdementen:
What about:

    sort_axis, sort_values --> sort_axis, sort
    argmax, argmin, argsort --> labelofmax, labelofmin, labelofsorted
    posargmax, posargmin, posargsort --> indexofmax, indexofmin, indexofsorted

alixdamman:
updated

alixdamman:
Needs changelog

larray/core/array.py
```diff
@@ -3104,10 +3114,15 @@ def argsort(self, axis=None, kind='quicksort'):
...
         axis = self.axes[axis]
-        pos = self.posargsort(axis, kind=kind)
+        pos = self.idxindirectsort(axis, kind=kind)
         return LArray(axis.labels[pos.data], pos.axes)
 
-    def posargsort(self, axis=None, kind='quicksort'):
+    def posargsort(self, axis=None):
+        """Deprecated
+        """
+        warnings.warn("posargsort has been replaced by idxargsort", FutureWarning, stacklevel=2)
```
gdementen: it needs to call the old method :)

larray/core/array.py
```diff
@@ -3028,6 +3033,11 @@ def argmax(self, axis=None):
             return tuple(axis.labels[i] for i, axis in zip(indices, self.axes))
 
     def posargmax(self, axis=None):
+        """Deprecated
+        """
+        warnings.warn("posargmax has been replaced by idxargmax", FutureWarning, stacklevel=2)
```
gdementen: it needs to call the old method :)

larray/core/array.py
```diff
@@ -3066,7 +3076,7 @@ def posargmax(self, axis=None):
         else:
             return np.unravel_index(self.data.argmax(), self.shape)
 
-    def argsort(self, axis=None, kind='quicksort'):
+    def indirectsort(self, axis=None, kind='quicksort'):
```
gdementen: we also need a dummy method with the old name

larray/core/array.py
gdementen: the new name is (according to the current state of this PR -- but as my comments show I think we can do better) idxindirectsort

larray/core/array.py
```diff
@@ -3134,15 +3149,15 @@ def posargsort(self, axis=None, kind='quicksort'):
...
+        >>> arr.idxindirectsort(X.nat)
         nat\\sex  M  F
               0  2  1
               1  0  2
               2  1  0
         """
         if axis is None:
             if self.ndim > 1:
-                raise ValueError("array has ndim > 1 and no axis specified for posargsort")
+                raise ValueError("array has ndim > 1 and no axis specified for indargsort")
```
gdementen: bad name of method

larray/core/group.py
```diff
@@ -1450,8 +1450,8 @@ def opmethod(self, other):
     __sub__ = difference
 
 
-class PGroup(Group):
-    """Positional Group.
+class IGroup(Group):
```
gdementen: we probably need a dummy object here too

alixdamman:
Let's choose:

    sort_axis, sort_values --> sort_axis, sort
    argmax, argmin, argsort --> labelofmax, labelofmin, labelsofsorted
    posargmax, posargmin, posargsort --> indexofmax, indexofmin, indexofsorted

gdementen:
wait a second, we need to discuss this.

alixdamman:
updated

doc/source/changes/version_0_26.rst.inc
```diff
@@ -10,6 +10,11 @@
 * renamed `na` argument of `read_csv`, `read_excel`, `read_hdf` and `read_sas` functions to `fill_value` to avoid
   confusion as to what the argument does and to be consistent with `reindex` and `align` (closes :issue:`394`).
 
+* renamed several methods (closes :issue:`50`):
+
+    - `sort_values` by `sort`
```
gdementen: s/by/to/

larray/core/group.py
```diff
@@ -1499,5 +1500,10 @@ def eval(self):
             raise ValueError("Cannot evaluate a positional group without axis")
 
     def __hash__(self):
-        return hash(('PGroup', _to_tick(self.key)))
+        return hash(('IGroup', _to_tick(self.key)))
 
+
+class PGroup(IGroup):
+    def __init__(self, key, name=None, axis=None):
+        warnings.warn("PGroup class has been renamed as IGroup", FutureWarning, stacklevel=2)
```
gdementen: s/as/to/

alixdamman:
I thought that `sort_values` --> `sort` was a user demand. I understand it is not. I prefer to keep `sort_values`.

gdementen:
No, it was something I came up with in the different propositions to be consistent with other sort methods: 
* sort vs indirectsort
* sort vs labelsort vs idxsort
but given that we have settled for labelsofsorted and indexofsorted, the consistency argument feels less obvious (but I can understand that to be fully consistent we should name our methods:
labelsofsorted_values and indexofsorted_values or something like that, which is not really an option IMO), so renaming to just sort() is maybe a tiny bit better...

As a side note, I just noticed another small inconsistency: labelsofsorted (plural) vs indexofsorted (singular). We should have either:
* labelsofsorted and indicesofsorted 
* labelofsorted and indexofsorted
but not mix both. I think the first option is better/clearer as to what the methods do.

alixdamman:
With `sort` the fact that the values are sorted is implicit and I don't like implicit stuff. With `sort_value` you know what is sorted.

alixdamman:
I also would choose `labelofsorted` and `indexofsorted` (singular) because as label/indexmin/max, it returns a LArray.

gdementen:
Ok, let's keep sort_values then.

Here is a little summary for posterity. We had a choice between:
* more explicit (and thus less source of confusion) for the "direct" sort but a tad less consistent (and thus more source of confusion) for the indirect sort : sort_values and labelsofsorted
* less explicit but a bit more consistent: sort and labelsofsorted

FWIW, pandas went from sort -> sort_values, so this is probably a hint that the explicit version is better than just "sort".

gdementen:
> I also would choose labelofsorted and indexofsorted (singular) because as label/indexmin/max, it returns a LArray.

I disagree. For the dimension you are working on (sorting/aggregating), the result is very different: min/max returns a single value (ie kills the dimension) while sort returns multiple values. This is especially visible for 1D arrays, but the principle holds for arrays with more dimensions.

alixdamman:
Explained like this, I agree.

alixdamman:
Done. I hope I didn't mess with all these new method names.

larray/core/array.py
```diff
@@ -1641,23 +1641,22 @@ def sort_values(self, key, reverse=False):
             raise NotImplementedError("sort_values key must have one dimension less than array.ndim")
         assert subset.ndim == 1
         axis = subset.axes[0]
-        posargsort = subset.posargsort()
+        indexofsorted = subset.indicesofsorted()
```
gdementen: s/indexof/indicesof/

doc/source/changes/version_0_26.rst.inc
```diff
@@ -10,6 +10,11 @@
 * renamed `na` argument of `read_csv`, `read_excel`, `read_hdf` and `read_sas` functions to `fill_value` to avoid
   confusion as to what the argument does and to be consistent with `reindex` and `align` (closes :issue:`394`).
 
+* renamed several methods (closes :issue:`50`):
```
gdementen: I think we should explain the motivation for any syntax change. eg. 
* to improve consistency with other methods by always using "index" instead of using a mix of "position" (P) and "index" (i).
* to make the meaning of the "arg" methods (hopefully) more intuitive.

You also need to mention PGroup -> IGroup

larray/core/group.py
```diff
@@ -1499,5 +1500,10 @@ def eval(self):
             raise ValueError("Cannot evaluate a positional group without axis")
 
     def __hash__(self):
-        return hash(('PGroup', _to_tick(self.key)))
+        return hash(('IGroup', _to_tick(self.key)))
 
+
+class PGroup(IGroup):
```
gdementen: wouldn't using renamed_to work?

doc/environment.yml
```diff
@@ -10,5 +10,6 @@ dependencies:
   - pytables
   - sphinx
   - numpydoc
+  - libiconv
```
gdementen: remove this commit?

alixdamman:
rebased and updated (used `renamed_to` for `PGroup` as suggested).

PR #204, opened 2017-04-04
Documentation improvement

alixdamman:
Still need to find a solution for larrayenv (conda-forge)


larray/example.py
```diff
@@ -0,0 +1,43 @@
+import os
+import larray as la
+
+EXAMPLE_FILES_DIR = os.path.dirname(__file__) + '/tests/data/'
+
+def load_example_arrays():
```
gdementen: We should have an argument to this function because we will probably include several example datasets in the future. The goal was to have something akin to http://www.stata.com/help.cgi?sysuse. The question is whether pop+hh+qx is a single "Session" dataset, or if those are three individual LArray dataset which happen to be somewhat linked. I would lean towards the former, but then we need a name for the global thing :)

gdementen: also, I never liked that hh array. We cannot combine it with the two others in anyway and it is not super interesting by itself, so I do not think it justifies its place in the tutorial.

doc/source/changes/version_0_22.rst.inc
```diff
@@ -90,4 +90,4 @@ Fixes
 
 * fixed group aggregates on integer arrays for median, percentile, var and std (closes :issue:`193`).
 
-* fixed group sum over boolean arrays (closes :issue:`194`).
+* fixed set_labels when inplace=True
```
gdementen: bad merge?

larray/__init__.py
```diff
@@ -6,6 +6,8 @@
 from larray.excel import open_excel
 from larray.ipfp import ipfp
 
+from larray.example import load_example_data
```
gdementen: this has nothing to do with the set_labels fix

larray/example.py
```diff
@@ -0,0 +1,54 @@
+import os
```
gdementen: this has nothing to do with the set_labels fix

larray/example.py
```diff
@@ -0,0 +1,54 @@
...
+        example_data = 'demography'
+    if not isinstance(example_data, str):
+        raise TypeError("Expected string for argument example_data")
+    if example_data not in AVAILABLE_EXAMPLE_DATA.keys():
+        raise ValueError("example_data must be chosen "
+                         "from list {}".format(list(AVAILABLE_EXAMPLE_DATA.keys())))
+
+    res = la.Session(AVAILABLE_EXAMPLE_DATA[example_data])
+
+    return res
```
gdementen: inline res?

larray/example.py
```diff
@@ -0,0 +1,54 @@
+import os
+import larray as la
+
+EXAMPLE_FILES_DIR = os.path.dirname(__file__) + '/tests/data/'
+AVAILABLE_EXAMPLE_DATA = {
+    'demography' : EXAMPLE_FILES_DIR + 'data.h5'
+}
+
+def load_example_data(example_data=None):
```
gdementen: I am unsure having a default value is a good idea? Wouldn't "name" be a better argument name?

larray/example.py
```diff
@@ -0,0 +1,54 @@
...
+    """Load arrays used in the tutorial so that all examples in it can be reproduced.
+
+    Parameters
+    ----------
+    example_data : str
+        Example data to load. Available example datasets are:
+
+        - demography
```
gdementen: Wouldn't it be better to populate this list automatically, so that we do not have to keep it in sync?

larray/example.py
```diff
@@ -0,0 +1,54 @@
...
+     sex [2]: 'M' 'F'
+     nat [2]: 'BE' 'FO'
+    """
+    if example_data is None:
+        example_data = 'demography'
+    if not isinstance(example_data, str):
+        raise TypeError("Expected string for argument example_data")
+    if example_data not in AVAILABLE_EXAMPLE_DATA.keys():
+        raise ValueError("example_data must be chosen "
+                         "from list {}".format(list(AVAILABLE_EXAMPLE_DATA.keys())))
```
gdementen: That's a good idea to give the list of possible values!

doc/source/changes/version_0_22.rst.inc
alixdamman: yes

alixdamman:
I cleaned my history and removed the commit where I replace gdmenten by conda-forge in README.rst from this PR. I don't know how much time it will take to push larrayenv on conda-forge.

gdementen:
LGTM, thanks!

PR #205, opened 2017-04-05
fix #18 : implement reindex_axis method

### LINKED ISSUE #18: implement .reindex_axis
```python
>>> arr = ndtest((2, 3))
>>> arr.reindex_axis(x.a, ['a1', 'a0', 'a1', 'a3'])
a\\b |  b0 |  b1 |  b2
  a1 |   3 |   4 |   5
  a0 |   0 |   1 |   2
  a1 |   3 |   4 |   5
  a3 | nan | nan | nan
```
It might be easy to implement using something vaguely looking like:
```python
>>> new_axis = Axis(old_axis.name, new_labels)
>>> missing_value = missing[dtype]
>>> old_indices = old_axis.translate(new_labels, missing=-1)
>>> result = self.i[old_indices]
>>> # fix up those which were missing
>>> result[old_indices == -1] = missing_value
```

alixdamman:


larray/core.py
```diff
@@ -4685,6 +4685,78 @@ def rename(self, renames=None, to=None, inplace=False, **kwargs):
...
+        a0 |  b0 | 0.0 | 1.0
+        a1 |  b1 | 6.0 | 7.0
+        a1 |  b2 | nan | nan
+        a1 |  b0 | 4.0 | 5.0
+        """
+        axis = self.axes[axis]
+        if not np.isscalar(fill_value):
+            raise TypeError("fill_value must be a scalar")
+        if np.isnan(fill_value):
+            object.__setattr__(self, 'data', self.data.astype(float))
```
gdementen: if inplace=False, self is modified, which it shouldn't

larray/core.py
```diff
@@ -4685,6 +4685,78 @@ def rename(self, renames=None, to=None, inplace=False, **kwargs):
...
+        new_labels = list(set(labels) - set(axis.labels))
```
gdementen: should use OrderedSet so that the order of new labels is respected

larray/core.py
gdementen: in any case, I *think* it shouldn't be necessary to promote explicitly because .extend should handle mixed dtype fine (and promote to the "larger" type)

larray/core.py
```diff
@@ -4685,6 +4685,78 @@ def rename(self, renames=None, to=None, inplace=False, **kwargs):
...
+        new_items = LArray(np.full((len(new_labels)), fill_value, dtype=self.dtype))
```
gdementen: You could use: `new_items = full(len(new_labels), fill_value)`

larray/core.py
```diff
@@ -4685,6 +4685,78 @@ def rename(self, renames=None, to=None, inplace=False, **kwargs):
...
+        target_axes = self.axes.replace(axis, Axis(axis.name, new_labels))
+        new_items = new_items.broadcast_with(target_axes)
```
gdementen: these two lines shoulnd't be necessary (.extend *should* handle this by default).

larray/core.py
gdementen: I see now that it's not strictly required because the final order does not depend on new_labels but this might somewhat improve performance since the memory access pattern would be more sequential.

larray/core.py
```diff
@@ -4685,6 +4685,78 @@ def rename(self, renames=None, to=None, inplace=False, **kwargs):
...
+            Array with reindexed axis.
+
+        Notes
+        -----
+        When introducing NAs into an array containing integers via reindex_axis,
+        all data will be promoted to float in order to store the NAs.
+
+        Examples
+        --------
+        >>> arr = ndtest((2, 2, 2))
```
gdementen: a 2D example (or even 1D?) would be just as expressive and shorter

larray/core.py
```diff
@@ -4685,6 +4685,78 @@ def rename(self, renames=None, to=None, inplace=False, **kwargs):
...
+    def reindex_axis(self, axis, labels, fill_value=np.nan, inplace=False):
+        """Change order of labels of an axis and/or add new ones.
+
+        Place NaN or given `fill_value` in locations having no value previously.
+
+        Parameters
+        ----------
+        axis : int, str or Axis
+            Axis along with to reindex.
+        labels : list of str
```
gdementen: the function should also accept an Axis argument, so that one can write: `arr.reindex_axis(x.a, other.a)`, which will probably be one of the main use case for this method.

gdementen:
As a general comment, given this method, users will want an method which reindex all axes at once.
`arr.reindex(target_axes)` and potentially `arr.reindex_like(target_array)` (which would reindex all common axes) and expand to add extra axes. Sure we could sequentially use .reindex_axis on each axis in turn, but it should be possible to be more efficient than that (ie reindex all axes at once).

gdementen:
One more implementation comment. I think that using something along the lines of:
```python
res_axes = self.axes.replace(axis, Axis(axis.name, labels))
res = full(res_axes, common_type(self.dtype, fill_value))
res[axis.labels] = self
```
might be more efficient (would have to test that though) because only a single copy is done.

alixdamman:
Need help.

I updated AxisCollection.replace so that we can do:
res_axes = self.axes.replace(axes_to_reindex, new_axis, **kwargs)
but I am stuck with how to implement reindex with multiple axes. 

For now, I have something like:
def reindex(self, axes_to_reindex=None, new_axis=None, fill_value=np.nan, inplace=False, **kwargs):
        if not np.isscalar(fill_value):
            raise TypeError("fill_value must be a scalar")
        res_axes = self.axes.replace(axes_to_reindex, new_axis, **kwargs)
        res = full(res_axes, common_type([self.data, [fill_value]]))
        res[???] = self
        return res

gdementen:
something along the lines of:
```python
res[tuple(axis[:] for axis in self.axes)] = self
```
should work (and if it does not, we should probably make it work). Or, if this does not work, we could be more explicit and do (this should be equivalent):
```python
res[tuple(res_axes[axis][axis.labels] for axis in self.axes)] = self
```
We also might want to make this work directly (but that should be left to another PR):
```python
res[tuple(self.axes)] = self
```
that adding support for passing Axis objects directly in `__getitem__`.

An the final step would be to support this too:
```python
res[self.axes] = self
```
that is, add support for an AxisCollection directly.

alixdamman:
I tried 
res[[res_axes[axis][axis.labels] for axis in self.axes]] = self
which didn't work but 
res[tuple(res_axes[axis][axis.labels] for axis in self.axes)] = self
works.

alixdamman:
New version. Do not merge, I need to cleanup my history first.

doc/source/changes/version_0_22.rst.inc
```diff
@@ -76,6 +76,19 @@
 
   even if this could suffer, to a lesser extent, the same problem as above when stacking many arrays.
 
+* updated AxisCollection.replace so as to replace one, several or all axes.
+
+* implemented reindex_axis allowing to change order of labels of an axis and/or add new ones (closes :issue:`18`).
```
gdementen: we need a mention of .reindex and in fact, I am not sure having .reindex_axis is worth it. I would rather add the "provide labels instead of an Axis" part of reindex_axis to .reindex, instead of having an extra method. This would have the added benefit that you could write stuff like:
```python
ndtest((2, 2)).reindex(a=['a1', 'a2', 'a0'], b=['b1', 'b0'], fill_value=-1)
```

larray/core.py
```diff
@@ -4180,9 +4239,9 @@ def set_axes(self, axes_to_replace=None, new_axis=None, inplace=False, **kwargs)
             Axes to replace. If a single axis reference is given, the `new_axis` argument must be provided.
             If a list of Axis or an AxisCollection is given, all axes will be replaced by the new ones.
             In that case, the number of new axes must match the number of the old ones.
-        new_axis : Axis
+        new_axis : , optional
```
gdementen: what is the type of the argument?

larray/core.py
```diff
@@ -4685,6 +4726,144 @@ def rename(self, renames=None, to=None, inplace=False, **kwargs):
...
+         a | b\\c |  c0 |  c1 |  c2
+        a0 |  b0 | 0.0 | 1.0 | nan
+        a0 |  b1 | 2.0 | 3.0 | nan
+        a0 |  b2 | nan | nan | nan
+        a1 |  b0 | 4.0 | 5.0 | nan
+        a1 |  b1 | 6.0 | 7.0 | nan
+        a1 |  b2 | nan | nan | nan
+        >>> b = Axis('b', ['b1', 'b2', 'b0'])
+        >>> c = Axis('c', ['c2', 'c1', 'c0'])
+        >>> arr.reindex({'b' : b, 'c' : c}, fill_value=-1)
```
gdementen: no spaces before :

larray/core.py
```diff
@@ -4685,6 +4726,144 @@ def rename(self, renames=None, to=None, inplace=False, **kwargs):
...
+         a | b\\c | c2 | c1 | c0
+        a0 |  b1 | -1 |  3 |  2
+        a0 |  b2 | -1 | -1 | -1
+        a0 |  b0 | -1 |  1 |  0
+        a1 |  b1 | -1 |  7 |  6
+        a1 |  b2 | -1 | -1 | -1
+        a1 |  b0 | -1 |  5 |  4
+        >>> arr.reindex({x.b : b, x.c : c})
```
gdementen: no spaces before :

larray/core.py
```diff
@@ -4685,6 +4726,144 @@ def rename(self, renames=None, to=None, inplace=False, **kwargs):
...
+        a0 |  b2 | nan | nan | nan
+        a0 |  b0 | nan | 1.0 | 0.0
+        a1 |  b1 | nan | 7.0 | 6.0
+        a1 |  b2 | nan | nan | nan
+        a1 |  b0 | nan | 5.0 | 4.0
+        """
+        if not np.isscalar(fill_value):
+            raise TypeError("fill_value must be a scalar")
+        res_axes = self.axes.replace(axes_to_reindex, new_axis, **kwargs)
+        res = full(res_axes, fill_value, dtype=common_type([self.data, [fill_value]]))
```
gdementen: you can use common_type((self.data, fill_value)) instead (ie no [] necessary around fill_value)

larray/core.py
```diff
@@ -4685,6 +4726,144 @@ def rename(self, renames=None, to=None, inplace=False, **kwargs):
...
+            New axis for each axis to reindex given as a keyword argument.
+
+        Returns
+        -------
+        LArray
+            Array with reindexed axes.
+
+        Notes
+        -----
+        When introducing NAs into an array containing integers via reindex_axis,
```
gdementen: reindex?

larray/core.py
```diff
@@ -4685,6 +4726,144 @@ def rename(self, renames=None, to=None, inplace=False, **kwargs):
...
+        Place NaN or given `fill_value` in locations having no value previously.
+
+        Parameters
+        ----------
+        axes_to_reindex : axis ref or dict {axis ref: axis} or list of tuple (axis ref, axis)
+                          or list of Axis or AxisCollection
+            Axes to reindex. If a single axis reference is given, the `new_axis` argument must be provided.
+            If a list of Axis or an AxisCollection is given, all axes will be reindexed by the new ones.
+            In that case, the number of new axes must match the number of the old ones.
+        new_axis : , optional
```
gdementen: arg type

larray/core.py
```diff
@@ -4685,6 +4726,144 @@ def rename(self, renames=None, to=None, inplace=False, **kwargs):
...
+            New axis if `axes_to_replace` contains a single axis reference.
+        fill_value : scalar, optional
+            Value set to data corresponding to added labels.
+            By defaults NaN.
```
gdementen: Defaults to NaN. <-- this is my preferred.
or
NaN by default.

larray/core.py
```diff
@@ -4685,6 +4726,144 @@ def rename(self, renames=None, to=None, inplace=False, **kwargs):
...
+        inplace : bool, optional
+            Whether or not to modify the original object or return a new array and leave the original intact.
+            By defaults False.
```
gdementen: same as above

larray/core.py
gdementen: this old comment is still valid. Why not use a 2D example?

larray/core.py
```diff
@@ -3062,37 +3062,96 @@ def copy(self):
...
-        old : Axis or list of Axis
-            Axis or axes to be replaced.
-        new : Axis or list of Axis
-            Axis or axes to be put in place of the `old` axis.
+        axes_to_replace : axis ref or dict {axis ref: axis} or list of tuple (axis ref, axis)
+                          or list of Axis or AxisCollection
+            Axes to replace. If a single axis reference is given, the `new_axis` argument must be provided.
+            If a list of Axis or an AxisCollection is given, all axes will be replaced by the new ones.
+            In that case, the number of new axes must match the number of the old ones.
+        new_axis : Axis
```
gdementen: needs optional mention

larray/core.py
```diff
@@ -3062,37 +3062,96 @@ def copy(self):
...
+            New axis if `axes_to_replace` contains a single axis reference.
+        inplace : bool
```
gdementen: needs "optional" and "defaults to" mentions,

larray/core.py
gdementen: still needs something better for the argument type

larray/core.py
```diff
@@ -4180,9 +4239,9 @@ def set_axes(self, axes_to_replace=None, new_axis=None, inplace=False, **kwargs)
             Axes to replace. If a single axis reference is given, the `new_axis` argument must be provided.
             If a list of Axis or an AxisCollection is given, all axes will be replaced by the new ones.
             In that case, the number of new axes must match the number of the old ones.
-        new_axis : Axis
+        new_axis : , optional
             New axis if `axes_to_replace` contains a single axis reference.
-        inplace : bool
+        inplace : bool, optional
             Whether or not to modify the original object or return a new array and leave the original intact.
```
gdementen: needs "defaults to" mention

larray/core.py
```diff
@@ -4236,37 +4295,19 @@ def set_axes(self, axes_to_replace=None, new_axis=None, inplace=False, **kwargs)
...
-                items = [(axes_to_replace, new_axis)]
-            else:
-                items = []
-            items += kwargs.items()
-            for old, new in items:
-                axes = axes.replace(old, new)
+        new_axes = self.axes.replace(axes_to_replace, new_axis, **kwargs)
         if inplace:
-            object.__setattr__(self, 'axes', axes)
+            if new_axes.ndim != self.ndim:
```
gdementen: I am unsure... but maybe add a mention of this in the changelog?

doc/source/changes/version_0_22.rst.inc
```diff
@@ -76,6 +76,29 @@
...
+    a\\b | b1 | b2 | b0
+     a0 |  1 | -1 |  0
+     a1 |  3 | -1 |  2
+    >>> a = Axis('a', ['a1', 'a2', 'a0'])
+    >>> b = Axis('b', ['b2', 'b1', 'b0'])
+    >>> arr.reindex({'a': a, 'b': b}, fill_value=-1)
+    a\\b | b2 | b1 | b0
+     a1 | -1 |  3 |  2
+     a2 | -1 | -1 | -1
+     a0 | -1 |  1 |  0
```
gdementen: Unsure but we might want to add an example of reindexing using another array axes (possibly even showing you can do a binary operation with them afterwards?)

larray/core.py
```diff
@@ -4236,37 +4295,19 @@ def set_axes(self, axes_to_replace=None, new_axis=None, inplace=False, **kwargs)
...
+                raise ValueError("number of axes (%d) does not match "
+                                 "number of dimensions of data (%d)"
+                                 % (new_axes.ndim, self.ndim))
+            if new_axes.shape != self.data.shape:
+                raise ValueError("length of axes %s does not match "
+                                 "data shape %s" % (new_axes.shape, self.data.shape))
+            object.__setattr__(self, 'axes', new_axes)
```
gdementen: if PR#213 is merged before this one, this should be changed

larray/core.py
```diff
@@ -4685,6 +4726,93 @@ def rename(self, renames=None, to=None, inplace=False, **kwargs):
...
+            raise TypeError("fill_value must be a scalar")
+        if isinstance(new_axis, (int, basestring, list, tuple)):
+            new_axis = Axis(self.axes[axes_to_reindex].name, new_axis)
+        res_axes = self.axes.replace(axes_to_reindex, new_axis, **kwargs)
+        res = full(res_axes, fill_value, dtype=common_type((self.data, fill_value)))
+        labels = tuple(axis[axis.labels] for axis in self.axes)
+        res[labels] = self
+        if inplace:
+            object.__setattr__(self, 'axes', res.axes)
+            object.__setattr__(self, 'data', res.data)
```
gdementen: if PR#213 is merged before this one, this should be changed

gdementen:
LGTM

alixdamman:
Thanks fro review

PR #208, opened 2017-04-06
fix #88 : (viewer) added AbstractEditor and a menu bar

### LINKED ISSUE #88: Add Menu File with Open, Save, Save As and Exit

alixdamman:
Partial implementation --> new, open, save, ... only operational for MappingEditor.  
I prefer to wait for a first review before to proceed. 

Not sure what to do with New? Currently, it is equivalent to a reset.
Not sure how to handle Open Recent? See comment in code.

Maybe, I need to find a way to warn the user that some arrays has been modified without being saved before to leave the viewer?

MappingEditor and ArrayEditor inherit from QMainWindow new instead of QDialog.
I need to find a solution for the implementation of accept() and reject() methods. QMainWindow does not have accept() and reject() methods. Currently Close button is broken (and other stuff).


gdementen:
Partial implementation --> new, open, save, ... only operational for MappingEditor.

> Not sure what to do with New? Currently, it is equivalent to a reset.

That's fine with me. We should avoid having that error box pop up in that case (or in every case?)

> Not sure how to handle Open Recent? See comment in code.

No idea, sorry. I guess you will need to experiment.

> Maybe, I need to find a way to warn the user that some arrays has been modified without being saved before to leave the viewer?

We need this indeed. This is very important for the standalone viewer. But this should be an option, because I do not think we want this when used from within a script (called via view()).

> MappingEditor and ArrayEditor inherit from QMainWindow new instead of QDialog.
I need to find a solution for the implementation of accept() and reject() methods. QMainWindow does not have accept() and reject() methods. Currently Close button is broken (and other stuff).

I do not think ArrayEditor should have this, only the MappingEditor. Implementing it for both would make things more complicated, for only a small benefit. Also for the accept and reject methods, you could try having an AbstractMappingEditor class with two children:
```python
# this would contain accept and reject stuff
class MappingEditorDialog(QDialog, AbstractMappingEditor):
    pass

# this would contain menu stuff etc
class MappingEditorApp(QMainWindow, AbstractMappingEditor):
    pass
```
I am unsure pyqt handle multiple-inheritance in this way though, but it's worth a shot.

larray/viewer.py
```diff
@@ -1851,8 +1859,150 @@ def __init__(self, parent=None):
...
+        file_menu.addAction(create_action(self, _('Open'), shortcut=QKeySequence.Open,
+                                          icon=None, triggered=self.open,
+                                          statustip=_('Load session from file')))
+        file_menu.addAction(create_action(self, _('Save'), shortcut=QKeySequence.Save,
+                                          icon=None, triggered=self.save,
+                                          statustip=_('Save all arrays and executed commands in new directory')))
+        file_menu.addAction(create_action(self, _('Save Data'), shortcut=QKeySequence("Shift+S"),
+                                          icon=None, triggered=self.saveData,
+                                          statustip=_('Save all current arrays in a Session file')))
+        file_menu.addAction(create_action(self, _('Save As'), shortcut=QKeySequence.SaveAs,
```
gdementen: None of these keyboard shortcuts seem to work currently. Unsure how to fix that nicely, but hardcoding the shortcuts as QShortcut like I did in `ArrayView.__init__` seems to work. There ought to be a better way.

larray/viewer.py
```diff
@@ -1851,8 +1859,150 @@ def __init__(self, parent=None):
...
+        if filepath:
+            self._openFile(filepath)
+        else:
+            messageBox = QMessageBox()
+            messageBox.warning(self, "Warning", "No file selected")
+
+    @Slot()
+    def openRecentFile(self):
+        action = self.sender()
+        if (action):
```
gdementen: extra parentheses

larray/viewer.py
```diff
@@ -1949,44 +2088,24 @@ def get_value(self):
...
-        if not title:
-            title = _("Session viewer") if readonly else _("Session editor")
-        if readonly:
-            title += ' (' + _('read only') + ')'
-        self.setWindowTitle(title)
-
-        layout = QVBoxLayout()
-        self.setLayout(layout)
+        if not isinstance(data, la.Session):
+            data = la.Session(data)
```
gdementen: Unsure we want this. This might make using the viewer as a debugger harder. We need to this think through.

larray/viewer.py
alixdamman: It's working on my machine. Which version  of Qt do you have ? But, anyway, I will change for explicit shortcuts.

larray/viewer.py
```diff
@@ -1961,32 +1965,42 @@ def __init__(self, parent=None):
...
         self.data = data
 
         icon = self.style().standardIcon(QStyle.SP_ComputerIcon)
         if icon is not None:
             self.setWindowIcon(icon)
 
         if not title:
-            title = _("Session viewer") if readonly else _("Session editor")
+            title = 'Session' if self.__class__ == MappingEditor else 'Array'
```
gdementen: not necessary anymore

larray/viewer.py
```diff
@@ -2252,22 +2281,135 @@ def set_widget_array(self, array, title):
...
+    def _isDataModified(self):
+        if self.arraywidget.model.readonly:
+            return False
+        else:
+            return len(self.arraywidget.model.changes) > 0 or self._appliedchanges
+
+    def _askToSaveIfDataModified(self):
+        if self._isDataModified():
+            ret = QMessageBox.warning(self, "Warning", "The data have been modified.\nDo you want to save your changes?",
+                               QMessageBox.Save | QMessageBox.Discard | QMessageBox.Cancel)
```
gdementen: strange indent

larray/viewer.py
```diff
@@ -2252,22 +2281,135 @@ def set_widget_array(self, array, title):
...
 
+    def _isDataModified(self):
+        if self.arraywidget.model.readonly:
+            return False
+        else:
+            return len(self.arraywidget.model.changes) > 0 or self._appliedchanges
+
+    def _askToSaveIfDataModified(self):
+        if self._isDataModified():
+            ret = QMessageBox.warning(self, "Warning", "The data have been modified.\nDo you want to save your changes?",
```
gdementen: s/have/has

larray/viewer.py
```diff
@@ -2252,22 +2281,135 @@ def set_widget_array(self, array, title):
...
+            QMessageBox.critical(self, "Error", "Current session could not be saved")
+            return False
+        else:
+            self._saveData(dialog.selectedFiles()[0])
+            return True
+
+    @Slot()
+    def about(self):
+        QMessageBox.about(self, _("About"), _("Larray's viewer. "
+                        "See <a href=\"http://larray.readthedocs.io/en/stable/\">larray_documentation</a> "
```
gdementen: I am not against an "about" window per se, but this one would be better as a direct "online documentation" link in the "help" menu (which is a great thing to have btw).

larray/viewer.py
```diff
@@ -2252,22 +2281,135 @@ def set_widget_array(self, array, title):
...
+            settings = QSettings()
+            files = settings.value("recentFileList")
+            if filepath in files:
+                files.remove(filepath)
+            files = [filepath] + files[:self.MAX_RECENT_FILES-1]
+            settings.setValue("recentFileList", files)
+        self.updateRecentFileActions()
+
+    def updateRecentFileActions(self):
+        # XXX : use QSettings instead of self.recentOpenedFiles attribute?
```
gdementen: remove XXX?

gdementen:
still crashes :( I suspect an exception is raised somewhere and this crashes the viewer.

alixdamman:
I guess I need to create a specific conda environment with Qt 5 to check why it crashes.

alixdamman:
OK. I've got:

Traceback (most recent call last):
  File "C:/Users/ald/Projects/larray/larray/viewer.py", line 2236, in ipython_cell_executed
    self.update_mapping(clean_ns)
  File "C:/Users/ald/Projects/larray/larray/viewer.py", line 2148, in update_mapping
    del self.data[k]
AttributeError: __delitem__

when I tried to do stuff in the qtconsole. I guess this is because I transformed self.data into a Session.

gdementen:
I found the cause of the remaining crash... Here (I guess because of Qt5), QFileDialog.getOpenFileName(self) returns a tuple (filepath, '') instead of a single string, which makes Session(this_tuple) raise an exception, which in turn crash the viewer.

larray/viewer.py
```diff
@@ -2626,11 +2777,11 @@ def edit(obj=None, title='', minvalue=None, maxvalue=None, readonly=False, depth
 
     dlg = MappingEditor(parent) if hasattr(obj, 'keys') else ArrayEditor(parent)
     if dlg.setup_and_check(obj, title=title, minvalue=minvalue, maxvalue=maxvalue, readonly=readonly):
-        if parent:
-            dlg.show()
-        else:
-            dlg.exec_()
+        if isinstance(obj, str):
+            dlg.updateRecentFileActions(obj)
```
gdementen: This code will never be called (because if obj WAS str, it should be converted to a Session by now).

larray/viewer.py
```diff
@@ -1949,10 +1940,20 @@ def get_value(self):
 DISPLAY_IN_GRID = (la.LArray, np.ndarray)
 
 
-class MappingEditor(QDialog):
+class MappingEditor(QMainWindow):
     """Session Editor Dialog"""
+
+    MAX_RECENT_FILES = 5
```
gdementen: 5 seems like too few, I would set this to at least 10.

larray/viewer.py
alixdamman: Old useless dead code. Sorry.

alixdamman:
I made a few changes in session.py 
If you have better ideas concerning XLWingsHandler, I'm open.

larray/viewer.py
```diff
@@ -2261,22 +2287,145 @@ def set_widget_array(self, array, title):
...
+        self.setCurrentFile(filepath)
+        self.statusBar().showMessage("File {} loaded".format(os.path.basename(filepath)), 4000)
+
+    @Slot()
+    def open(self):
+        if self._askToSaveIfDataModified():
+            filepath = QFileDialog.getOpenFileName(self)
+            # Qt5 returns a tuple (filepath,) instead of a string
+            if isinstance(filepath, (list, tuple)):
+                filepath, = filepath
```
alixdamman: Not yet but I will.

larray/session.py
```diff
@@ -190,6 +192,8 @@ def _dump(self, key, value, *args, **kwargs):
         self.handle[key] = value.dump(*args, **kwargs)
 
     def save(self):
+        if self.new_file:
+            self.handle['Sheet1'].delete()
```
alixdamman: You fixed this in LArray.to_excel method.

gdementen: btw, if you want to tackle this, it would be better to avoid creating a sheet in the first place (if possible), instead of deleting it afterwards. If you do not want to tackle this, please open an issue for it.

larray/viewer.py
```diff
@@ -1985,8 +1990,13 @@ def setup_and_check(self, data, title='', readonly=False,
             title += ' (' + _('read only') + ')'
         self.setWindowTitle(title)
 
+        self.statusBar().showMessage("Welcome to the LArray Viewer", 4000)
+
+        widget = QWidget()
+        self.setCentralWidget(widget)
```
gdementen: why do we need the extra widget? is it because QMainWindow instead of QDialog?

larray/session.py
alixdamman: No, it is not possible to open a Workbook without a sheet. I updated XLWingsHandler to rename the first sheet when I first call the _dump() method.

larray/viewer.py
alixdamman: http://doc.qt.io/qt-4.8/qmainwindow.html

Below image in section 'Qt Main Window Framework'

Note: Creating a main window without a central widget is not supported. You must have a central widget even if it is just a placeholder.

doc/source/changes/version_0_22.rst.inc
```diff
@@ -208,3 +211,9 @@ Fixes
 
 * fixed array creation functions not raising an exception when called with wrong syntax func(axis1, axis2, ...)
   instead of func([axis1, axis2, ...]) (closes :issue:`203`).
+
+* fixed adding new sheets in a Workbook: a new sheet is appended instead of prepended. (closes :issue:`229`)
+
+* fixed dump of a Session as Excel file: result file contained an empty sheet `Sheet1` (closes :issue:`230`)
```
gdementen: you fixed more than this now :) When creating a file via open_excel...

larray/excel.py
```diff
@@ -158,14 +167,14 @@ def __setitem__(self, key, value):
                     value.xw_sheet.api.Copy(xw_sheet.api)
                     xw_sheet.delete()
                 else:
-                    xw_sheet = self[-1]
+                    xw_sheet = self[-1].xw_sheet
```
gdementen: need changelog

larray/session.py
```diff
@@ -209,17 +209,15 @@ def _open_for_write(self):
 
     def list(self):
         # strip extension from files
-        # FIXME: only take .csv files
         # TODO: also support fname pattern, eg. "dump_*.csv" (using glob)
-        return [os.path.splitext(fname)[0] for fname in os.listdir(self.fname)]
+        return [os.path.splitext(fname)[0] for fname in os.listdir(self.fname) if '.csv' in fname]
```
gdementen: need changelog

larray/session.py
```diff
@@ -333,6 +331,9 @@ def __setitem__(self, key, value):
     def _ipython_key_completions_(self):
         return list(self.keys())
 
+    def __delitem__(self, key):
```
gdementen: need changelog

larray/session.py
```diff
@@ -368,14 +369,15 @@ def load(self, fname, names=None, engine='auto', display=False, **kwargs):
         # TODO: support path + *.csv
         if engine == 'auto':
             _, ext = os.path.splitext(fname)
-            engine = ext_default_engine[ext.strip('.')]
+            engine = ext.strip('.') if '.' in ext else 'csv'
```
gdementen: name of variable ('engine') is misleading. Should be ext or stripped_ext or normalized_ext or whatever.

larray/session.py
```diff
@@ -406,7 +409,7 @@ def dump(self, fname, names=None, engine='auto', display=False, **kwargs):
             arrays = [(k, v) for k, v in arrays if k in names_set]
         handler.dump_arrays(arrays, display=display, **kwargs)
 
-    def dump_hdf(self, fname, names=None, *args, **kwargs):
+    def to_hdf(self, fname, names=None, *args, **kwargs):
```
gdementen: needs changelog, a dummy method with the old name with a deprecation warning and a mention of the #217 issue in the commit message + changelog :)

larray/viewer.py
```diff
@@ -1383,8 +1379,8 @@ def set_data(self, data, xlabels=None, ylabels=None, current_filter=None,
                 filters_layout.addWidget(self.create_filter_combo(axis))
             filters_layout.addStretch()
         self.filtered_data = self.la_data
-        if data.size == 0:
-            QMessageBox.critical(self, _("Error"), _("Array is empty"))
+        #if data.size == 0:
```
gdementen: remove the lines

larray/viewer.py
gdementen: Ok, thanks for the info.

larray/session.py
alixdamman: Done

larray/session.py
alixdamman: Done

larray/excel.py
alixdamman: Done

doc/source/changes/version_0_22.rst.inc
```diff
@@ -127,6 +127,14 @@
 
 * viewer: added possibility to delete an array by pressing Delete on keyboard (closes :issue:`116`).
 
+* viewer: added a menu bar 'File' with options 'New', 'Open', 'Save', 'SaveAs', 'Open Recent' and 'Quit'.
+
+  WARNING: Only LArray objects are saved.
```
gdementen: I didn't check, but is this clear enough in the viewer itself?

doc/source/changes/version_0_22.rst.inc
```diff
@@ -208,3 +218,16 @@ Fixes
 
 * fixed array creation functions not raising an exception when called with wrong syntax func(axis1, axis2, ...)
   instead of func([axis1, axis2, ...]) (closes :issue:`203`).
+
+* fixed adding new sheets in a workbook: a new sheet is appended instead of prepended (closes :issue:`229`).
```
gdementen: this is a tad misleading: if you read up to the ":", you think it did not work at all. I would phrase this like: "fixed position of added sheets in excel workbook: new sheets are appended instead of prepended."

doc/source/changes/version_0_22.rst.inc
```diff
@@ -208,3 +218,16 @@ Fixes
...
+
+* fixed __setitem__ of Workbook when value is an instance of `Sheet` and key is new.
+
+* fixed Workbook behavior in case of new workbook: the first added sheet replaces the default sheet `Sheet1`
+  (closes :issue:`230`).
+
+* fixed load session: read arrays in the same order as they are stored.
```
gdementen: something like: "fixed order of arrays in sessions loaded from files. The order is now the same as in the file"?

doc/source/changes/version_0_22.rst.inc
```diff
@@ -208,3 +218,16 @@ Fixes
 
 * fixed array creation functions not raising an exception when called with wrong syntax func(axis1, axis2, ...)
   instead of func([axis1, axis2, ...]) (closes :issue:`203`).
+
+* fixed adding new sheets in a workbook: a new sheet is appended instead of prepended (closes :issue:`229`).
+
+* fixed __setitem__ of Workbook when value is an instance of `Sheet` and key is new.
```
gdementen: a tiny (or even partial) example would make this clearer.

alixdamman: Well, it crashes on my side. Could you test it?

doc/source/changes/version_0_22.rst.inc
alixdamman: What do you mean? Tooltip?

PR #218, opened 2017-04-12
fix #116 : (viewer) added possibility to delete an array from the list

### LINKED ISSUE #116: delete arrays in viewer using del key in addition to via the console
See point 3 of #114.

alixdamman:
Well, this PR allows to delete arrays one by one and not several at once. 
It is not possible to select multiple items in the QListWidget right now. 

We could start with this PR and remove 'fix #116' from the commit and really fix #116 later ?

### DIFF
--- a/doc/source/changes/version_0_22.rst.inc
+++ b/doc/source/changes/version_0_22.rst.inc
@@ -127,7 +127,11 @@
 
   (closes :issue:`170`)
 
+* viewer: added possibility to delete an array by pressing Delete on keyboard.
+  (closes :issue:`116`)
+
 .. _misc:
+
 
 Miscellaneous improvements
 --------------------------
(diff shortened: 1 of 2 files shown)


gdementen:
one by one is enough for now, let's close the issue. This PR needs a changelog entry though.

alixdamman:
I removed 'fix #116' from the commit and updated the changelog

gdementen:
LGTM, could you just rebase so that this can be commited nicely? I tried to fix the conflict using github web interface but this does not seem to fix it. Sorry about this. If you could eradicate it while rebasing, that would be nice.

doc/source/changes/version_0_22.rst.inc
```diff
@@ -164,8 +168,6 @@ Miscellaneous improvements
 
 (closes :issue:`152`)
 
-* improved documentation of plot method (closes :issue:`169`)
```
gdementen: bad merge, it seems :)

PR #220, opened 2017-04-13
fix #203 : raise error in array creation functions 

### LINKED ISSUE #203: zeros(axis1, axis2) should raise an error
Other creation functions (ones, ndrange, ...) are also affected obviously.

```python
>>> a = Axis('a', 'a0..a1')
>>> b = Axis('b', 'b0..b2')
>>> zeros(a, b)
a |  a0 |  a1
  | 0.0 | 0.0
>>> zeros((a, b))
a\b |  b0 |  b1 |  b2
 a0 | 0.0 | 0.0 | 0.0
 a1 | 0.0 | 0.0 | 0.0
```

alixdamman:
fix #203

### DIFF
--- a/larray/core.py
+++ b/larray/core.py
@@ -91,6 +91,7 @@ import os
 import re
 import sys
 import warnings
+import functools
 
 try:
     import builtins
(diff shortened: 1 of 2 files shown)


larray/core.py
```diff
@@ -10343,12 +10355,13 @@ def empty_like(array, title='', dtype=None, order='K'):
     return LArray(np.empty_like(array.data, dtype, order), array.axes, title)
 
 
+# No need to add @_check_axes_argument, `axes` argument will be checked in empty
```
gdementen: I think it **is** needed, because what needs to be checked is fill_value.

larray/core.py
```diff
@@ -10130,12 +10130,22 @@ def read_sas(filepath, nb_index=None, index_col=None, na=np.nan, sort_rows=False
     return df_aslarray(df, sort_rows=sort_rows, sort_columns=sort_columns, fill_value=na)
 
 
+def _check_axes_argument(func):
+    def wrapper(*args, **kwargs):
+        if len(args) > 1 and isinstance(args[1], (int, Axis)):
+            raise ValueError("If you want to pass several axes or dimension lengths to {}, "
+                             "you must pass them as a list (using []) or tuple (using()).".format(func.__name__))
+        return func(*args, **kwargs)
```
gdementen: when wrapping a function, you need to copy its __name__ and __doc__ and possibly other stuff, or preferably, use functools.wraps on the wrapper.

larray/core.py
alixdamman: It will return an error if fill_value is an integer. But I can test if fill_value is not an Axis before to call empty inside full

larray/core.py
```diff
@@ -10381,6 +10396,8 @@ def full(axes, fill_value, title='', dtype=None, order='C'):
          BE | 0 | 1
          FO | 0 | 1
     """
+    if isinstance(fill_value, Axis):
+        raise TypeError("`fill_value` must be a numeric value")
```
gdementen: wouldn't it be better to raise the same error than other creation functions?

larray/core.py
gdementen: ok, so rather than "no need to add...", the comment should be something like: "we cannot use ... because an integer fill_value would be considered as an error"

gdementen:
LGTM

PR #221, opened 2017-04-13
fix #192 : made percentile([multiple, percentiles], axis) working.

### LINKED ISSUE #192: percentile([multiple, percentiles], axis) break
```python
>>> from larray import *
>>> a = ndtest((2, 3))
# this works (but we might want to return an LArray instead)
>>> a.percentile([10, 20])
array([ 0.5,  1. ])
>>> a.percentile([10, 20], 'a')
ValueError                                Traceback (most recent call last)
...
ValueError: number of axes (1) does not match number of dimensions of data (2)
```
This is because the generic aggregate code does not expect an aggregation function to "add" an extra axis.

alixdamman:
Result array contains a new Axis with name='percentile' and labels=[multiple, percentiles]

### DIFF
--- a/larray/core.py
+++ b/larray/core.py
@@ -7692,6 +7692,10 @@ class LArray(object):
         """
         pass
 
+    # XXX : for performance reasons, we should use the fact that the
+    #       underlying numpy function handles multiple percentiles in one call.
+    #       This is easy to implement in _axis_aggregate() but not in _group_aggregate()
+    #       since in this case np.percentile() may be called several times.
     # percentile needs an explicit method because it has not the same
     # signature as other aggregate functions (extra argument)
     def percentile(self, q, *args, **kwargs):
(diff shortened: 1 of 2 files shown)


larray/core.py
```diff
@@ -7661,8 +7661,11 @@ def percentile(self, q, *args, **kwargs):
         if skipna is None:
             skipna = True
         func = np.nanpercentile if skipna else np.percentile
-        return self._aggregate(func, args, kwargs, keepaxes=keepaxes,
-                               commutative=True, extra_kwargs={'q': q})
+        if isinstance(q, (list, tuple)):
+            return stack([(v, self._aggregate(func, args, kwargs, keepaxes=keepaxes, commutative=True,
+                                              extra_kwargs={'q': v})) for v in q], 'percentile')
```
gdementen: This is the wrong way to fix it because the underlying numpy function handles multiple percentiles in one call (which is much faster than multiple calls with one percentile each).

alixdamman: OK in case of axes but in _group_aggregate(), op(arr, axis=axis_idx, out=out, **kwargs) may be called several times and creates a new dimension each time.

gdementen:
thanks!

Issue #190, opened 2017-04-03
should std ddof default to 0 or 1?


gdementen:
numpy defaults to 0, pandas defaults to 1

I have no idea what is a better default value. I would tend to favor pandas over numpy since pandas is more popular these days, but I have no clue, really.

Issue #189, opened 2017-04-03
ddof argument of std does not work, possibly others


gdementen:
```python
In [1]: from larray import *
In [2]: a = ndtest((2, 3))
In [3]: a.std(ddof=1)
KeyError                                  Traceback (most recent call last)
...
KeyError: "axis 'ddof' not found in {a, b}"
```
This is probably the case for all "extra" keyword arguments handled by the underlying numpy function and not within larray.

PR #222, opened 2017-04-13
Fix #41, #189 and #190 : updated agg methods docstring + solved ddof arg 

### LINKED ISSUE #41: Change signature of functions sum, mean, ...

alixdamman:


alixdamman:
Once this PR is merged (or just before the next release), we should build documentation with tag='latest' on readthedocs and see if everything is OK.

gdementen:
The change in ddof needs a changelog entry. It also closes issue #190. I imagine (doc)tests will fail though because the numbers will not be the same.

gdementen:
Also on what ground did you choose ddof=1 over ddof=0? Just because of Pandas or for another reason?

larray/core.py
```diff
@@ -6625,29 +6626,29 @@ def percent(self, *axes):
...
+        def decorated(func):
+            def wrapper(self, *args, **kwargs):
+                keepaxes = kwargs.pop('keepaxes', False)
+                skipna = kwargs.pop('skipna', None)
+                if skipna is None:
+                    skipna = nanfunc is not None
+                if skipna and nanfunc is None:
+                    raise ValueError("skipna is not available for {}".format(func.__name__))
+                _npfunc = nanfunc if skipna else npfunc
+                extra_kwargs = {k: kwargs.pop(k) for k in list(kwargs.keys()) if k in _kwarg_agg.keys()}
```
gdementen: where is _kwarg_agg defined?

larray/core.py
```diff
@@ -6625,29 +6626,29 @@ def percent(self, *axes):
...
+                if 'percentile' in func.__name__:
+                    extra_kwargs['q'] = args[0]
```
gdementen: This seems too specific. Wouldn't an some kind of extra argument (eg. fill_extra_kwargs) to _decorate_agg_method be more generic here? which would be given as a dict: {argname: argposition}. For example for percentile, that would be:
{'q': 0}?

larray/core.py
```diff
@@ -6713,9 +6714,11 @@ def method(self, *args, **kwargs):
         >>> # or equivalently
         >>> # barr.all('a0,a1>>a01;a2,a3>>a23')
         """.format(_doc_agg_method("AND reduction", kwargs="out,skipna,keepaxes"))
+        pass
```
gdementen: This is very strange to use a decorator over a function and not use the decorated function at all, except its name and signature. Is there no other way?

alixdamman:
It seems that using S_{n-1} is a common practice in statistics. 
https://en.wikipedia.org/wiki/Standard_deviation#Corrected_sample_standard_deviation

I am not sure why Numpy set ddof to 0 be default.

larray/core.py
alixdamman: just before finction _doc_agg_method which is defined above class LArray

larray/core.py
alixdamman: Yes, there is another way. Need some additional work.

larray/core.py
gdementen: It did not appear in the diff, so I thought it was some broken code. Sorry about this!

larray/core.py
alixdamman: I discover that explicit arguments like ddof=1 are not passed to wrapper. I will try to add extra_args=() and extra_kwargs={} as arguments to _decorate_agg_method.

alixdamman:
Need review before going further. 
Using decorator was a bad idea. It is a nightmare to handle optional arguments. 

I propose something that is somewhat close to what Pandas does --> see end of file 
https://github.com/pandas-dev/pandas/blob/v0.19.2/pandas/core/generic.py

It is also a nightmare to find simple doctests for var, std and describe with ddof=1 by default...

alixdamman:
Do not review... I'm working on it.

alixdamman:
Overriding signature in doctstring is working with Sphinx --> 
http://larray-test.readthedocs.io/en/fix_agg_methods_docstring_and_ddof_arg_41_189/api.html#la-agg
but not sure that PyCharm is smart enough to show 'new' signature with Ctrl+Q and Ctrl+P

gdementen:
The result is great. You just need to include both the "extra arguments fix" and default ddof change in the changelog.

gdementen:
LGTM

PR #227, opened 2017-04-24
misc quality of life improvements

gdementen:


doc/source/changes/version_0_22.rst.inc
```diff
@@ -151,42 +159,48 @@ Miscellaneous improvements
...
+  var, ...). Closes :issue:`41`.
 
-    >>> anonymous = Axis('0..100')
-    >>> age = Axis('age=0..100')
-    >>> gender = Axis('M,F', 'gender')
+* improved documentation of plot method (closes :issue:`169`).
+
+* improved auto-completion in interactive consoles for both LArray and Session objects. LArray objects can now
+  complete keys within [].
+
```
gdementen: yes! It is a relatively new feature in ipython. I should probably specify that it's ipython specific.

gdementen:
I did not test sessions with numpy arrays. Thinking of which, the pickle part needs tests. As for #217, I do not see what is the relationship with this PR?!

doc/source/changes/version_0_22.rst.inc
alixdamman: My question was not about `tab` but about to specify `using tab` after `within []`.

alixdamman:
Well, for me #217 is just a small misc. improvement. It does not need a specific PR.

doc/source/changes/version_0_22.rst.inc
gdementen: hehe, ok :)

alixdamman:
In Session.py - class FileHandler - method read_arrays(), res shouldn't be declared as an OrderedDict instead of a simple dict (line 87)?

gdementen:
It should indeed.

PR #233, opened 2017-05-01
fix #232 : replaced nose by pytest

### LINKED ISSUE #232: Use pytest instead of nose

alixdamman:


doc/source/changes/version_0_22.rst.inc
```diff
@@ -190,6 +190,8 @@ Miscellaneous improvements
 
   will autocomplete to `s.a_nice_test_array` and `s['a_<tab>` will be completed to `s['a_nice_test_array`
 
+* tests: replaced nose by pytest (closes :issue:`232`)
```
gdementen: Unsure we should mention this. So far, the changelog has been user-oriented and I do not think any of our current users cares about this.

gdementen:
except potentially for the changelog, this LGTM

doc/source/changes/version_0_22.rst.inc
alixdamman: Since we do not have external contributors currently, I agree.

PR #235, opened 2017-05-02
misc stuff

gdementen:


larray/utils.py
```diff
@@ -438,3 +439,14 @@ def find_closing_chr(s, start=0):
...
+
+
+def float_error_handler_factory(stacklevel):
+    def error_handler(error, flag):
+        if error == 'invalid value':
+            error = 'invalid value (NaN)'
+            extra = ' (this is typically caused by a 0 / 0)'
+        else:
+            extra = ''
+        warnings.warn("{} encountered during operation{}".format(error, extra), RuntimeWarning, stacklevel=stacklevel)
```
alixdamman: Need an extra withespace after 'operation' in warning.warn

gdementen: this is intentional (so that if extra is '', we do not get a trailing space).

gdementen: the "missing" space is within "extra"

alixdamman: I missed that, sorry. Everything fine then.

larray/core.py
```diff
@@ -4363,6 +4363,14 @@ def __getattr__(self, key):
...
             raise AttributeError("'{}' object has no attribute '{}'".format(self.__class__.__name__, key))
 
+    # needed to make *un*pickling work (because otherwise, __getattr__ is called before .axes exists, which leads to
+    # an infinite recursion)
+    def __getstate__(self):
+        return self.__dict__
+
+    def __setstate__(self, d):
+        self.__dict__ = d
+
```
alixdamman: Are there some tests associated?

gdementen: indirectly, yes. the Session pickling test includes Axis and LArray objects, which means they are pickled and unpickled too.

alixdamman:
Except for the crash with Python 2.7, this LGTM

gdementen:
the fix for 2.7 is coming in a few seconds :)

PR #240, opened 2017-05-03
Viewer : filter extension when opening a file + update kernel when adding/deleting arrays

### LINKED ISSUE #238: new and open do not clear arrays from kernel

alixdamman:
fix #238 and #239

### DIFF
--- a/larray/session.py
+++ b/larray/session.py
@@ -202,23 +202,28 @@ class PandasCSVHandler(FileHandler):
         pass
 
     def _open_for_write(self):
-        try:
-            os.makedirs(self.fname)
-        except OSError:
-            if not os.path.isdir(self.fname):
-                raise
+        if self.fname is not None:
+            try:
+                os.makedirs(self.fname)
+            except OSError:
+                if not os.path.isdir(self.fname):
+                    raise ValueError("Path {} must represent a directory".format(self.fname))
 
     def list(self):
         # strip extension from files
         # TODO: also support fname pattern, eg. "dump_*.csv" (using glob)
-        return sorted([os.path.splitext(fname)[0] for fname in os.listdir(self.fname) if '.csv' in fname])
+        if self.fname is not None:
+            return sorted([os.path.splitext(fname)[0] for fname in os.listdir(self.fname) if '.csv' in fname])
+        else:
+            return []
 
     def _read_array(self, key, *args, **kwargs):
-        fpath = os.path.join(self.fname, '{}.csv'.format(key))
+        fpath = os.path.join(self.fname, '{}.csv'.format(key)) if self.fname is not None else key
         return read_csv(fpath, *args, **kwargs)
 
     def _dump(self, key, value, *args, **kwargs):
-        value.to_csv(os.path.join(self.fname, '{}.csv'.format(key)), *args, **kwargs)
+        fpath = os.path.join(self.fname, '{}.csv'.format(key)) if self.fname is not None else key
+        value.to_csv(fpath, *args, **kwargs)
 
     def close(self):
         pass
--- a/larray/session.py
+++ b/larray/session.py
@@ -92,7 +92,10 @@ class FileHandler(object):
         for key in keys:
             if display:
                 print("loading", key, "...", end=' ')
-            dest_key = key.strip('/')
+            if os.path.isfile(key):
+                dest_key = os.path.splitext(os.path.basename(key))[0]
+            else:
+                dest_key = key.strip('/')
             res[dest_key] = self._read_array(key, *args, **kwargs)
             if display:
                 print("done")
@@ -400,8 +403,8 @@ class Session(object):
         fname : str
             Path to the file.
         names : list of str, optional
-            List of arrays to load. Defaults to all valid objects present in
-            the file/directory.
+            List of arrays to load. If `fname` is None, list of paths to CSV files.
+            Defaults to all valid objects present in the file/directory.
         engine : str, optional
             Load using `engine`. Defaults to 'auto' (use default engine for
             the format guessed from the file extension).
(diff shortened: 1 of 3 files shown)


alixdamman:
Well, I should test if qtconsole_available before to play with kernel...

alixdamman:
Done.

larray/viewer.py
```diff
@@ -2197,7 +2197,10 @@ def update_mapping(self, value):
     @Slot()
```
gdementen: I do not think the `@Slot()` is necessary for new-style signal-slots.

larray/viewer.py
```diff
@@ -2293,12 +2296,17 @@ def _add_arrays(self, arrays):
...
+            self.kernel.shell.push(dict(arrays))
 
     def _clear_arrays(self):
         arrays = [k for k, v in self.data.items() if self._display_in_grid(k, v)]
         for name in arrays:
             del self.data[name]
             self.delete_list_item(name)
+        if qtconsole_available:
+            for name in arrays:
+                self.kernel.shell.del_var(name)
```
gdementen: I wonder if it wouldn't be possible to refactor delete_current_item, _add_arrays, update_mapping, etc. to reduce duplication of code.

larray/viewer.py
```diff
@@ -2100,12 +2100,12 @@ def void_formatter(array, *args, **kwargs):
...
 
-        file_menu.addAction(create_action(self, _('New'), shortcut=QKeySequence("Ctrl+N"), triggered=self.new))
-        file_menu.addAction(create_action(self, _('Open'), shortcut=QKeySequence("Ctrl+O"), triggered=self.open,
+        file_menu.addAction(create_action(self, _('&New'), shortcut="Ctrl+N", triggered=self.new))
+        file_menu.addAction(create_action(self, _('&Open'), shortcut="Ctrl+O", triggered=self.open,
                                           statustip=_('Load session from file')))
-        file_menu.addAction(create_action(self, _('Save'), shortcut=QKeySequence("Ctrl+S"), triggered=self.save,
+        file_menu.addAction(create_action(self, _('&Save'), shortcut="Ctrl+S", triggered=self.save,
                                           statustip=_('Save all arrays as a session in a file')))
         file_menu.addAction(create_action(self, _('Save As'), triggered=self.save_as,
```
gdementen: Usually, the "A" from "As" is underlined (&As) in file menus.

larray/viewer.py
```diff
@@ -2332,34 +2358,50 @@ def new(self):
...
-                self._open_file(filepath)
+                filepaths = QFileDialog.getOpenFileNames(self, filter=filter)
+            if len(filepaths) >= 1:
+                if all(['.csv' in filepath for filepath in filepaths]):
+                    self._open_file(filepaths)
+                elif len(filepaths) == 1:
+                    self._open_file(filepaths[0])
+                else:
+                    QMessageBox.critical(self, "Error",
+                                         "It possible to load several CSV files or one Excel or HDF file")
```
gdementen: It is only possible?

larray/viewer.py
```diff
@@ -2332,34 +2358,50 @@ def new(self):
...
         # XXX : clear console history ?
         self._clear_arrays()
-        session = la.Session(filepath)
+        session = la.Session()
+        if '.csv' in filepath:
+            filepath = [filepath]
+        if isinstance(filepath, (list, tuple)):
+            directory = os.path.dirname(filepath[0])
+            names = [os.path.splitext(os.path.basename(fp))[0] for fp in filepath]
+            session.load(directory, names)
```
gdementen: we probably want to add "something" to open several .csv files in Session. In Session.load maybe?

larray/viewer.py
```diff
@@ -2332,34 +2358,50 @@ def new(self):
...
+            for name in names:
+                self.set_current_file('{}/{}.csv'.format(directory, name))
```
gdementen: This smells like a small refactor is necessary: move most of the functionality of set_current_file to a method called something like "update_recent_files". Then make set_current_file call that method and call that method here instead of using set_current_file.

larray/viewer.py
```diff
@@ -2332,34 +2358,50 @@ def new(self):
...
+            self.statusBar().showMessage("Arrays {} from CSV files loaded".format(' ,'.join(names)), 4000)
+        else:
+            session.load(filepath)
+            self.set_current_file(filepath)
+            self.statusBar().showMessage("File {} loaded".format(os.path.basename(filepath)), 4000)
```
gdementen: I wonder if the statusbar update shouldn't be done in set_current_File?

larray/viewer.py
gdementen: Also, the "R" from "Open Recent" could also have a shortcut.

alixdamman:
I think I do have another problem: new() does not 'clear' the session, it only removes all arrays displayed in the QListWidget.

larray/viewer.py
alixdamman: In Session.load or in PandasCSVHandler.

gdementen: good question :)

alixdamman:
Still need to implement _reset_session + add pickle in open()

larray/session.py
```diff
@@ -92,7 +92,10 @@ def read_arrays(self, keys, *args, **kwargs):
         for key in keys:
             if display:
                 print("loading", key, "...", end=' ')
-            dest_key = key.strip('/')
+            if os.path.isfile(key):
+                dest_key = os.path.splitext(os.path.basename(key))[0]
```
gdementen: I don't really like this particular change because it is too specific. It breaks the generic contract that keys are the actual destination (session) keys. I will fix this myself though.

PR #249, opened 2017-05-09
miscexcel

gdementen:
reuse global instance in Excel I/O

### DIFF
--- a/larray/excel.py
+++ b/larray/excel.py
@@ -1,4 +1,5 @@
 import os
+import atexit
 
 import numpy as np
 try:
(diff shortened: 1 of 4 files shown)


larray/excel.py
```diff
@@ -83,8 +111,9 @@ def __init__(self, filepath=None, overwrite_file=False, visible=None, silent=Non
                     app = xw_wkb.app
 
             # active workbook use active app by default
-            if filepath == -1 and app is None:
-                app = -1
+            # XXX: I wonder if we have a choice here (ie maybe the check for app == "global" is superfluous)
+            if filepath == -1 and app == "global":
+                app = "active"
```
alixdamman: Yes, I think it is superfluous. We can return an error with an explicit message if filepath == -1 and app != "global"?

gdementen:
I rewrote parts of this to make it actually useful in practice :)

gdementen:
hang on... one more fix incoming :)

PR #254, opened 2017-05-11
join

gdementen:
WIP implementation of array.align and all that was necessary
(AxisCollection.align, Axis.align)

### DIFF
--- a/doc/source/changes/version_0_25.rst.inc
+++ b/doc/source/changes/version_0_25.rst.inc
@@ -5,6 +5,77 @@
   It contains a link to open the documentation, a shortcut to launch the user interface in edition mode
   and a shortcut to update larrayenv. Closes issue:`281`.
 
+* implemented array.align(other_array) which makes two arrays compatible with each other (by making all common axes
+  compatible). This is done by adding, removing or reordering labels for each common axis according to the join method
+  used:
+
+    - outer: will use a label if it is in either arrays axis (ordered like the first array). This is the default as it
+             results in no information loss.
+    - inner: will use a label if it is in both arrays axis (ordered like the first array)
+    - left: will use the first array axis labels
+    - right: will use the other array axis labels
+
+  The fill value for missing labels defaults to nan.
+
+    >>> arr1 = ndtest((2, 3))
+    >>> arr1
+    a\b  b0  b1  b2
+     a0   0   1   2
+     a1   3   4   5
+    >>> arr2 = -ndtest((3, 2))
+    >>> # reorder array to make the test more interesting
+    >>> arr2 = arr2[['b1', 'b0']]
+    >>> arr2
+    a\\b  b1  b0
+     a0  -1   0
+     a1  -3  -2
+     a2  -5  -4
+
+  Align arr1 and arr2
+
+    >>> aligned1, aligned2 = arr1.align(arr2)
+    >>> aligned1
+    a\b   b0   b1   b2
+     a0  0.0  1.0  2.0
+     a1  3.0  4.0  5.0
+     a2  nan  nan  nan
+    >>> aligned2
+    a\b    b0    b1   b2
+     a0   0.0  -1.0  nan
+     a1  -2.0  -3.0  nan
+     a2  -4.0  -5.0  nan
+
+  After aligning all common axes, one can then do operations between the two arrays
+
+    >>> aligned1 + aligned2
+    a\b   b0   b1   b2
+     a0  0.0  0.0  nan
+     a1  1.0  1.0  nan
+     a2  nan  nan  nan
+
+  The fill value for missing labels defaults to nan but can be changed to any compatible value.
+
+    >>> aligned1, aligned2 = arr1.align(arr2, fill_value=0)
+    >>> aligned1
+    a\b  b0  b1  b2
+     a0   0   1   2
+     a1   3   4   5
+     a2   0   0   0
+    >>> aligned2
+    a\b  b0  b1  b2
+     a0   0  -1   0
+     a1  -2  -3   0
+     a2  -4  -5   0
+    >>> aligned1 + aligned2
+    a\b  b0  b1  b2
+     a0   0   0   2
+     a1   1   1   5
+     a2  -4  -5   0
+
+* implemented Axis.align(other_axis) and AxisCollection.align(other_collection) which makes two axes / axis collections
+  compatible with each other, see LArray.align above.
+
+
 Miscellaneous improvements
 --------------------------
 
@@ -14,6 +85,7 @@ Miscellaneous improvements
 * added icon to display in Windows start menu and editor windows.
 
 * added 'Getting Started' section in documentation.
+
 
 Fixes
 -----
--- a/larray/core/array.py
+++ b/larray/core/array.py
@@ -1479,8 +1479,9 @@ class LArray(ABCLArray):
         Parameters
         ----------
         key : scalar or tuple or Group
-            Key along which to sort.
-            Must have exactly one dimension less than ndim.
+            Key along which to sort. Must have exactly one dimension less than ndim.
+        reverse : bool, optional
+            Sort values in descending order. Defaults to False (ascending order).
 
         Returns
         -------
--- a/larray/core/array.py
+++ b/larray/core/array.py
@@ -1473,6 +1473,178 @@ class LArray(ABCLArray):
         else:
             return res
 
+    def align(self, other, join='outer', fill_value=nan, axes=None):
+        """Align two arrays on their axes with the specified join method.
+
+        In other words, it ensure all common axes are compatible. Those arrays can then be used in binary operations.
+
+        Parameters
+        ----------
+        other : LArray-like
+        join : {'outer', 'inner', 'left', 'right'}, optional
+            Join method. For each axis common to both arrays:
+              - outer: will use a label if it is in either arrays axis (ordered like the first array).
+                       This is the default as it results in no information loss.
+              - inner: will use a label if it is in both arrays axis (ordered like the first array)
+              - left: will use the first array axis labels
+              - right: will use the other array axis labels.
+        fill_value : scalar or LArray, optional
+            Value to use for missing values. Defaults to NaN.
+        axes : AxisReference or sequence of them, optional
+            Axes to align. Need to be valid in both arrays. Defaults to None (all common axes). This must be specified
+            when mixing anonymous and non-anonymous axes.
+
+        Returns
+        -------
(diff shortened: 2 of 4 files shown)


gdementen:
also the choice of default join method is not innocent and should be considered carefully, as this should be consistent with what we will do for binary ops (which should be almost trivial to implement once this PR is merged).

larray/core.py
```diff
@@ -1830,6 +1830,45 @@ def difference(self, other):
...
+        >>> axis1.align(axis2)
+        Axis(['a0', 'a1', 'a2', 'a3'], 'a')
+        >>> axis1.align(axis2, join='inner')
+        Axis(['a1', 'a2'], 'a')
+        >>> axis1.align(axis2, join='left')
+        Axis(['a0', 'a1', 'a2'], 'a')
+        >>> axis1.align(axis2, join='right')
+        Axis(['a1', 'a2', 'a3'], 'a')
+        """
+        assert join in {'outer', 'inner', 'left', 'right'}
```
gdementen: I should raise a ValueError exception instead of using an assert

larray/core.py
```diff
@@ -3762,6 +3801,63 @@ def split_axis(self, axis, sep='_', names=None, regex=None):
...
+        ])
+        >>> aligned2
+        AxisCollection([
+            Axis(['a0', 'a1', 'a2'], 'a'),
+            Axis(['b0', 'b1', 'b2'], 'b')
+        ])
+        """
+        assert join in {'outer', 'inner', 'left', 'right'}
+        other = other if isinstance(other, AxisCollection) else AxisCollection(other)
+        join_axes = list(OrderedSet(self.names) & OrderedSet(other.names)) if axes is None else axes
```
alixdamman: What about anonymous axes? Did you try what happen in case of anonymous axes?

larray/core.py
```diff
@@ -5069,6 +5165,76 @@ def get_labels(self_axis):
...
+        a1 |  b0 | 5.0
+        a1 |  b1 | 7.0
+        a1 |  b2 | nan
+        a2 |  b0 | nan
+        a2 |  b1 | nan
+        a2 |  b2 | nan
+        """
+        other = aslarray(other)
+        left_axes, right_axes = self.axes.align(other.axes, join=join, axes=axes)
+        return self.reindex(left_axes, fill_value=fill_value), other.reindex(right_axes, fill_value=fill_value)
```
alixdamman: If I understand well, AxisCollection.align returns the same AxisCollection `new_axes` twice. So why not returning only one `new_axes` collection and replace here left_axes, right_axes by common_axes?

larray/core.py
gdementen: Not yet. I will certainly break at this point. The way to handle them is relatively easy when there are only anonymous axes (ie do it by axis position) but when mixing anonymous and non-anonymous, this gets tricky and I don't really know what we should do.

larray/core.py
gdementen: No. Please see the comment at https://github.com/liam2/larray/pull/254/files#diff-d2e8c0a8b16062fcfdf68b1de7e842caR3857

alixdamman:
should be referenced in changelog 0.23

gdementen:
* I am still unsure about the default value for join here
* it still does not work for anonymous axes, but I am unsure we should care at this point

doc/source/changes/version_0_25.rst.inc
```diff
@@ -5,6 +5,67 @@
   It contains a link to open the documentation, a shortcut to launch the user interface in edition mode
   and a shortcut to update larrayenv. Closes issue:`281`.
 
+* implemented array.align(other_array) which makes two arrays compatible with each other (by making all common axes
+  compatible). This is done by adding, removing or reordering labels for each common axis according to the join method:
+
+    - outer: will use a label if it is in either arrays axis (ordered like the first array). This is the default.
```
alixdamman: This is the default option as it results in zero information loss.

doc/source/changes/version_0_25.rst.inc
```diff
@@ -5,6 +5,67 @@
   It contains a link to open the documentation, a shortcut to launch the user interface in edition mode
   and a shortcut to update larrayenv. Closes issue:`281`.
 
+* implemented array.align(other_array) which makes two arrays compatible with each other (by making all common axes
+  compatible). This is done by adding, removing or reordering labels for each common axis according to the join method:
```
alixdamman: ... compatible). The fill value for missing labels defaults to nan. This is done by ...

alixdamman: according to the join methodS?

doc/source/changes/version_0_25.rst.inc
```diff
@@ -5,6 +5,67 @@
...
+    - left: will use the first array axis labels
+    - right: will use the other array axis labels
+
+    >>> arr1 = ndtest((2, 3))
+    >>> arr1
+    a\b  b0  b1  b2
+     a0   0   1   2
+     a1   3   4   5
+    >>> arr2 = -ndtest((3, 2))
+    >>> # reorder array to make the test more interesing
```
alixdamman: interesTing

doc/source/changes/version_0_25.rst.inc
```diff
@@ -5,6 +5,67 @@
...
+    >>> arr2 = arr2[['b1', 'b0']]
+    >>> arr2
+    a\\b  b1  b0
+     a0  -1   0
+     a1  -3  -2
+     a2  -5  -4
```
alixdamman: /blankline
align arrays arr1 and arr2
/blankline

larray/core/array.py
```diff
@@ -1473,14 +1473,175 @@ def get_labels(self_axis):
         else:
             return res
 
+    def align(self, other, join='outer', fill_value=nan, axes=None):
```
alixdamman: I would keep 'outer' as default option but maybe add a warning in the doctrsing.

larray/core/array.py
```diff
@@ -1473,14 +1473,175 @@ def get_labels(self_axis):
...
+
+        Parameters
+        ----------
+        other : LArray-like
+        join : {'outer', 'inner', 'left', 'right'}, optional
+            Join method. For each axis common to both arrays:
+              - outer (default): will use a label if it is in either arrays axis (ordered like the first array)
+              - inner: will use a label if it is in both arrays axis (ordered like the first array)
+              - left: will use the first array axis labels
+              - right: will use the other array axis labels.
```
alixdamman: Defaults to 'outer' as it results in zero information loss.

larray/core/array.py
```diff
@@ -1473,14 +1473,175 @@ def get_labels(self_axis):
...
+        fill_value : scalar or LArray, optional
+            Value to use for missing values. Defaults to NaN.
```
alixdamman: nan?

larray/core/array.py
```diff
@@ -1473,14 +1473,175 @@ def get_labels(self_axis):
...
+
+        Examples
+        --------
+        >>> arr1 = ndtest((2, 3))
+        >>> arr1
+        a\\b  b0  b1  b2
+         a0   0   1   2
+         a1   3   4   5
+        >>> arr2 = -ndtest((3, 2))
+        >>> # reorder array to make the test more interesing
```
alixdamman: same as in changelog

larray/core/array.py
```diff
@@ -1473,14 +1473,175 @@ def get_labels(self_axis):
...
+        >>> arr2 = arr2[['b1', 'b0']]
+        >>> arr2
+        a\\b  b1  b0
+         a0  -1   0
+         a1  -3  -2
+         a2  -5  -4
```
alixdamman: same as in changelog

larray/core/array.py
alixdamman: Need also a comment about anonymous axes (maybe a doctest that raises an Error).

larray/core/axis.py
```diff
@@ -998,6 +998,45 @@ def difference(self, other):
...
+        ----------
+        other : Axis or label sequence
+        join : {'outer', 'inner', 'left', 'right'}, optional
+            Defaults to 'outer'.
+
+        Returns
+        -------
+        Axis
+            Aligned axis
+
```
alixdamman: See Also
--------
LArray.align

larray/core/axis.py
```diff
@@ -2200,6 +2242,65 @@ def split_axis(self, axis, sep='_', names=None, regex=None):
...
+            Axes to align. Need to be valid in both arrays. Defaults to None (all common axes).
+
+        Returns
+        -------
+        (left, right) : (AxisCollection, AxisCollection)
+            Aligned collections
+
+        Examples
+        --------
+
```
alixdamman: No need for this blankline after Examples

larray/core/axis.py
```diff
@@ -2200,6 +2242,65 @@ def split_axis(self, axis, sep='_', names=None, regex=None):
...
+        join : {'outer', 'inner', 'left', 'right'}, optional
+            Defaults to 'outer'.
+        axes : AxisReference or sequence of them, optional
+            Axes to align. Need to be valid in both arrays. Defaults to None (all common axes).
+
+        Returns
+        -------
+        (left, right) : (AxisCollection, AxisCollection)
+            Aligned collections
+
```
alixdamman: See Also
--------
LArray.align

alixdamman:
Couldn't we consider the fact that it doesn't work for anonymous axes as a feature and not a bug? 
I think we should keep it like this for moment and update align if users ask (need).

doc/source/changes/version_0_25.rst.inc
gdementen: there is only one method used at a time

larray/core/array.py
gdementen: What do you mean by a warning? This is already documented...

alixdamman: where? I missed it.

alixdamman: where? I missed it.

alixdamman: Sorry, for the warning, I was talking about the anonymous axes.

larray/core/array.py
gdementen: see #335

alixdamman:
Could you add an entry for align in api.rst? Thanks

larray/core/axis.py
alixdamman: This docstring should be only read by developers or also users? If developers only, no need for section "See Also".

alixdamman:
LGTM

Issue #256, opened 2017-05-11
unsaved changes indicator in viewer


gdementen:
using an * next to the filename in the window title is sometimes used.

Issue #181, opened 2017-03-30
custom title in a MappingEditor (session viewer) is ignored


gdementen:
It is overwritten by the selected array name, instead of being prepended to it.

alixdamman:
So, displayed title = '{custom title} - {current file} - {current array info}'?

PR #257, opened 2017-05-11
miscviewer

gdementen:
misc fixes/code improvements in the viewer

### DIFF
--- a/larray/viewer.py
+++ b/larray/viewer.py
@@ -2367,11 +2367,9 @@ class MappingEditor(QMainWindow):
     def open(self):
         if self._ask_to_save_if_unsaved_modifications():
             filter = "All (*.xls *xlsx *.h5 *.csv);;Excel Files (*.xls *xlsx);;HDF Files (*.h5);;CSV Files (*.csv)"
+            res = QFileDialog.getOpenFileNames(self, filter=filter)
             # Qt5 returns a tuple (filepaths, '') instead of a string
-            if PYQT5:
-                filepaths, _ = QFileDialog.getOpenFileNames(self, filter=filter)
-            else:
-                filepaths = QFileDialog.getOpenFileNames(self, filter=filter)
+            filepaths = res[0] if PYQT5 else res
             if len(filepaths) >= 1:
                 if all(['.csv' in filepath for filepath in filepaths]):
                     self._open_file(filepaths)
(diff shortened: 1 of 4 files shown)


gdementen:
once this PR is merged and the issue of showing the current file in the viewer is fixed, I think we are good to go for the release

alixdamman:
I would like to fix #181 (custom title) and #256 (put * if unsaved change) before the release. If you agree.

larray/viewer.py
```diff
@@ -2418,7 +2426,8 @@ def open_documentation(self):
 
     def set_current_file(self, filepath):
         self.update_recent_files([filepath])
-        self.currentFile = filepath
+        self.current_file = filepath
+        # TODO: update window title
```
alixdamman: or update window title in set_widget_array?

larray/viewer.py
```diff
@@ -2433,17 +2442,17 @@ def update_recent_files(self, filepaths):
...
+
+        # zip will iterate up to the shortest of the two
+        for filepath, action in zip(recent_files, self.recent_file_actions):
+            action.setText(os.path.basename(filepath))
+            action.setStatusTip(filepath)
+            action.setData(filepath)
+            action.setVisible(True)
+        # if we have less recent recent files than actions, hide the remaining actions
+        for action in self.recent_file_actions[len(recent_files):]:
+            action.setVisible(False)
```
alixdamman: bad copy/paste from C++ example. Sorry.

larray/viewer.py
```diff
@@ -2176,8 +2176,12 @@ def update_mapping(self, value):
         changed_keys = [k for k in keys_after if value[k] is not self.data.get(k)]
```
gdementen: Indeed, that's a good point/catch. But I do not think there is any practical way to do this. We would need to keeps copies of every array in memory just to check them... :(

larray/viewer.py
gdementen: no worries

larray/viewer.py
gdementen: this is already the case, but the trick is that set_current_file can be called in cases where set_widget_array is not called (eg on save), and the title should still be update in that case.

gdementen:
I would rather wait for 0.23 to fix #256, because I would like to make the release today and I fear fixing #256 properly will take too much time. Besides it is not urgent. Even Excel has no explicit indicator of modified file. I would like to have it but it is a nice-to-have not a must-have IMO.

PR #258, opened 2017-05-11
fix #181 : custom title in MappingEditor

### LINKED ISSUE #181: custom title in a MappingEditor (session viewer) is ignored

alixdamman:
displayed title = 'associated file - name : axes_info - custom part' if LArray object, else 'name - custom part'

I will include #256 after your review on current commit (fix #181)

### DIFF
--- a/larray/viewer.py
+++ b/larray/viewer.py
@@ -2163,11 +2163,11 @@ class MappingEditor(QMainWindow):
         # if the currently selected item (value) need to be refreshed (e.g it was modified)
         if prev_selected and prev_selected[0] == changed_items[0]:
             # we need to update the array widget explicitly
-            self.set_widget_array(self.data[to_display], to_display)
+            self.set_current_array(self.data[to_display], to_display)
         else:
             # for some reason, on_item_changed is not triggered when no item was selected
             if not prev_selected:
-                self.set_widget_array(self.data[to_display], to_display)
+                self.set_current_array(self.data[to_display], to_display)
             self._listwidget.setCurrentItem(changed_items[0])
 
     def update_mapping(self, value):
@@ -2230,7 +2230,7 @@ class MappingEditor(QMainWindow):
 
     def view_expr(self, array, *args, **kwargs):
         self._listwidget.clearSelection()
-        self.set_widget_array(array, '<expr>')
+        self.set_current_array(array, '<expr>')
 
     def _display_in_grid(self, k, v):
         return not k.startswith('__') and isinstance(v, DISPLAY_IN_GRID)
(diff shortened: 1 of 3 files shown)


larray/viewer.py
```diff
@@ -1983,11 +1983,13 @@ def setup_and_check(self, data, title='', readonly=False, minvalue=None, maxvalu
...
-            title = _("Session viewer") if readonly else _("Session editor")
+        if title:
+            self.title = title
+        else:
+            self.title = _("Session viewer") if readonly else _("Session editor")
         if readonly:
-            title += ' (' + _('read only') + ')'
-        self.setWindowTitle(title)
+            self.title += ' (' + _('read only') + ')'
+        self.setWindowTitle(self.title)
```
gdementen: in this kind of case, I prefer to use simply "title" and set self.title at the end.

larray/viewer.py
```diff
@@ -2292,14 +2294,24 @@ def on_item_changed(self, curr, prev):
...
             axes = array.axes
             axes_info = ' x '.join("%s (%d)" % (display_name, len(axis))
                                    for display_name, axis
                                    in zip(axes.display_names, axes))
-            title = (title + ': ' + axes_info) if title else axes_info
-        self.setWindowTitle(title)
+            title += [(name + ': ' + axes_info) if name else axes_info]
+        elif name:
+            title = [name]
+        title += [self.title]
```
gdementen: we might want to move this whole "title construction" to its own method

gdementen:
as I said in #257, I would postpone fixing #256 to 0.23 (unless you have done it already)

alixdamman:
OK to postpone #256 to 0.23

alixdamman:
rebase done + updated my commit "fix 181..."

larray/viewer.py
```diff
@@ -2295,16 +2296,34 @@ def on_item_changed(self, curr, prev):
             else:
                 self.eval_box.setText(expr)
 
-    # TODO: rename to set_current_array
-    def set_widget_array(self, array, title):
+    def update_title(self):
+        current_item = self._listwidget.currentItem()
+        name = str(current_item.text()) if current_item else ''
+        array = self.data[name] if name else None
```
gdementen: does this work for `<expr>`? (computed in the console)?

alixdamman: If qtconsole_available, yes.

gdementen:
LGTM then

larray/viewer.py
gdementen: well... It does not... I will fix this.

alixdamman: I tried with something like 
[1] arr4 = arr4.reshape(...)
which updated the title correctly. 
I guess I misunderstood when view_expr() is called.

gdementen: it is called if you try something like:
`arr4.reshape(...)` (without assigning it to anything)

PR #267, opened 2017-05-22
implemented as_table(light=True)

gdementen:
this will be useful to implement optional lighter display

if we do not want to make this the default, I ma unsure about the final user-facing API, but this can come later

### DIFF
--- a/larray/core/array.py
+++ b/larray/core/array.py
@@ -104,7 +104,8 @@ from larray.core.expr import ExprNode
 from larray.core.group import Group, PGroup, LGroup, remove_nested_groups, _to_key, _to_keys, _range_to_slice
 from larray.core.axis import Axis, AxisCollection, x, _make_axis
 from larray.util.misc import (table2str, size2str, basestring, izip, rproduct, ReprString, duplicates,
-                              float_error_handler_factory, csv_open, skip_comment_cells, strip_rows, _isnoneslice)
+                              float_error_handler_factory, csv_open, skip_comment_cells, strip_rows, _isnoneslice,
+                              light_product)
 
 nan = np.nan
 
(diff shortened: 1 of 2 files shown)


larray/core.py
```diff
@@ -5965,6 +5964,22 @@ def as_table(self, maxlines=None, edgeitems=5):
...
+         ['a0', 'b0', 0, 1, 2],
+         ['a0', 'b1', 3, 4, 5],
+         ['a1', 'b0', 6, 7, 8],
+         ['a1', 'b1', 9, 10, 11]]
+        >>> list(arr.as_table(light=True))  # doctest: +NORMALIZE_WHITESPACE
+        [['a', 'b\\\\c', 'c0', 'c1', 'c2'],
+         ['a0', 'b0', 0, 1, 2],
+         ['', 'b1', 3, 4, 5],
+         ['a1', 'b0', 6, 7, 8],
+         ['', 'b1', 9, 10, 11]]
```
alixdamman: Why using list in `as_table` examples?

larray/utils.py
```diff
@@ -212,6 +212,35 @@ def rproduct(*i):
     return product(*[x[::-1] for x in i])
 
 
+def light_product(*iterables, repeat=1):
+    """Cartesian product of input iterables, replacing repeated values by empty strings.
```
alixdamman: signature not accepted in Python 2.7

gdementen: yes... Will fix this...

larray/core.py
gdementen: because it currently return a generator

gdementen:
having something like set_printoptions was one of the ideas. We could also have our own function to display arrays to the console, instead of print(), so that we can pass it arguments. See other issue about this.

PR #270, opened 2017-05-22
fix #265 : added flag argument `new_file` to open_excel function

### LINKED ISSUE #265: using open_excel("inexisting file") yields ununderstandble error
The problem is that:
```python
wb = open_excel("inexisting_file.xlsx")
```
does NOT fail, but rather creates a new blank workbook, and then:
```python
toto = wb['my_sheet']
```
fails. That error is not understandable either, but I will open another issue for this.

The solution to the first problem is to add an argument to open_excel and only create a new workbook if explicitly asked for.

alixdamman:


larray/excel.py
```diff
@@ -453,9 +453,58 @@ def load(self, header=True, convert_float=True, nb_index=None, index_col=None):
...
+        >>> arr
+        a\\b  b0  b1  b2
+         a0   0   1   2
+         a1   3   4   5
+         a2   6   7   8
+        
+        create and save an array
+        
+        >>> # argument new_file of open_excel must be set to True if you want to create a new Excel file
+        >>> wb = open_excel(abspath('excel_file.xlsx'), True)
```
gdementen: I would rather use a with statement

larray/excel.py
```diff
@@ -453,9 +453,58 @@ def load(self, header=True, convert_float=True, nb_index=None, index_col=None):
...
+        >>> wb['arr'] = arr.dump()
+        >>> wb.save()
+        >>> wb.close()
+        
+        read array from Excel file
+        
+        >>> wb = open_excel(abspath('excel_file.xlsx'))
```
gdementen: also a with statement

larray/excel.py
```diff
@@ -453,9 +453,58 @@ def load(self, header=True, convert_float=True, nb_index=None, index_col=None):
...
+        >>> wb.close()
+        """
+        if not os.path.isfile(filepath) and not new_file:
+            raise ValueError("File {} does not exist. Please give the path to an existing file "
+                             "or set argument `new_file` to True to create a new file.".format(filepath))
+        else:
+            return Workbook(filepath, **kwargs)
 else:
-    def open_excel(filepath=None, **kwargs):
+    def open_excel(filepath=None, new_file=False, **kwargs):
```
gdementen: the new_file argument should rather be handled in Workbook. Also I think a mode argument  ("w" | "r" | "a") would be more consistent with the rest of Python. Opening a file read-only is currently not supported by xlwings (https://github.com/ZoomerAnalytics/xlwings/issues/318), so we will need to add a warning in the doc that the file can be modified anyway.

FWIW, using win32com, something like this might work:
```
wb = excel.Workbooks.Open(excelPath, ReadOnly=1)
# OR
wb = excel.Workbooks.Open(excelPath, ReadOnly=1)
```

larray/excel.py
```diff
@@ -453,9 +453,58 @@ def load(self, header=True, convert_float=True, nb_index=None, index_col=None):
...
+        """
+        Open an Excel workbook.  
+        
+        Parameters
+        ----------
+        filepath: str
+            Path to an Excel file. The file must exist if `new_file` is not set to True.
+        new_file: bool, optional
+            If True and filepath represents a not existing Excel file, creates a new file.
+        \**kwargs
```
gdementen: other kwargs from Workbook should be copied here explicitly (OR open_excel should be deprecated telling users to use Workbook directly)

larray/excel.py
alixdamman: a mode like "a" does make sens and "r" is not a read-only mode. Note sure that a mode argument is suitable for open_excel function. 
Why not replace `overwrite_file` argument by something like `create_or_overwrite_file` ?

larray/tests/test_array.py
```diff
@@ -2651,7 +2651,7 @@ def test_open_excel(self):
         # 1) with headers
         # ===============
 
-        with open_excel(abspath('test_open_excel.xlsx')) as wb:
+        with open_excel(abspath('test_open_excel.xlsx'), True) as wb:
```
gdementen: I would rather use an explicit keyword argument here

larray/tests/test_array.py
```diff
@@ -2731,7 +2731,7 @@ def test_open_excel(self):
 
         # 2) without headers
         # ==================
-        with open_excel(abspath('test_open_excel_no_headers.xlsx')) as wb:
+        with open_excel(abspath('test_open_excel_no_headers.xlsx'), True) as wb:
```
gdementen: idem

larray/excel.py
gdementen: In a broad sense mode="a" makes sense. It means "modify existing", and mode="r" is a read-only mode for normal files open(filename, mode="r") and *could* be implemented in open_excel in the long term

larray/tests/test_array.py
```diff
@@ -2648,10 +2648,31 @@ def test_to_excel_xlwings(self):
 
     @pytest.mark.skipif(xw is None, reason="xlwings is not available")
     def test_open_excel(self):
-        # 1) with headers
-        # ===============
+        # 1) Mode
+        # =======
+        # 1a) Read and read+ modes crash if file does not exist
+        with pytest.raises(ValueError):
+            open_excel(abspath('test_open_excel.xlsx'), mode='r')
```
gdementen: In these kind of case I prefer when the filename makes it obvious that it does not exist (it would also make sure we do not accidentally break the test with another test creating this file). eg. 'this_file_does_not_exist.xlsx'

larray/io/excel.py
```diff
@@ -455,9 +431,77 @@ def load(self, header=True, convert_float=True, nb_index=None, index_col=None):
...
+    Use None for a new blank workbook, -1 for the last active
+    workbook. Defaults to None.
+mode : {'r', 'w', 'a'}, optional
+    the mode to open the file. It can be one of the following:
+
+    * 'r': Read; similar to 'a', but the file must already exist 
+      (the read-only mode is currently not supported by xlwings).
+    * 'w': Write; a new file is created (an existing file with the same name would be deleted).
+    * 'a': Append; an existing file is opened for reading and writing, and if the file does not exist it is created.
+    * 'r+': currently idem as 'r'.
```
gdementen: Unsure we need to support 'r+'. This will probably confuse our users more than anything. But maybe changing overwrite_file to mode was a bad idea after all if we do not implement all modes/do not match normal behaviour. I am ashamed to admit I did not realize mode='a' is supposed to be 'write only'. Sorry about this! Do you we need to discuss this some more or would you prefer to revert to what you had done earlier and be done with it?

As a side note, if we do support it I think r+ is closer to 'a' than to 'r'.

alixdamman: OK to revert. 

Question: Do you think xlwings will have a read-only mode one day?

gdementen: I guess it will. No idea when though. At worse (if we actually need the functionality before it is implemented), we could implement it ourselves and submit a PR to them.

larray/io/excel.py
```diff
@@ -105,6 +79,9 @@ def __init__(self, filepath=None, overwrite_file=False, visible=None, silent=Non
                     if not ext.startswith('.xl'):
                         raise ValueError("'%s' is not a supported file "
                                          "extension" % ext)
+                    if not os.path.isfile(filepath) and not overwrite_file:
+                        raise ValueError("File {} does not exist. Please give the path to an existing file "
+                                         "or set mode to 'w' or 'a'".format(filepath))
```
gdementen: this needs to be changed too (yes I know, this is all my fault, please do not hate me too much ;-))

alixdamman: Oops.

PR #275, opened 2017-05-23
fix #271 : changed from_string sep default to ' '

### LINKED ISSUE #271: change from_string sep default to ' '
It would be very convenient to be able to use from_string directly on the output of an array (modulo \\ doubling)

Currently we need to specify sep manually:
```python
>>> from_string('''a\\b  b0  b1  b2
 a0   0   1   2
 a1   3   4   5''', sep=' ').info
2 x 3
 a [2]: 'a0' 'a1'
 b [3]: 'b0' 'b1' 'b2'
```

alixdamman:


larray/io/array.py
```diff
@@ -436,27 +436,28 @@ def from_string(s, nb_index=None, index_col=None, sep=',', **kwargs):
 
     Examples
     --------
-    >>> from_string("sex,M,F\\n,0,1")
+    >>> # if one dimension array and default separator ' ', a - must be added in front of each data line
```
gdementen: of the data line? (there is only one after all)

alixdamman:
Still need to update the changelog. Do not merge yet.

gdementen:
my comment meant to replace "each" by "the" since there is only one line in case of a 1D array

PR #278, opened 2017-05-24
fix #191 : replaced all occurences of `VG` by `LGroup`

### LINKED ISSUE #191: eradicate VG

alixdamman:


gdementen:
LGTM

PR #279, opened 2017-05-24
miscfix

gdementen:
a first batch of misc improvements and fixes


doc/source/changes/version_0_23.rst.inc
```diff
@@ -12,19 +12,57 @@ Miscellaneous improvements
```
alixdamman: Remove New features section if we decide to release very soon?

PR #290, opened 2017-05-31
fix #285 : prefer import from top package larray instead of larray.(...)

### LINKED ISSUE #285: imports in tests
shouldn't tests import functions from larray "top" package instead of from subpackage to check they are exported in larray?

alixdamman:


larray/tests/test_session.py
```diff
@@ -5,9 +5,8 @@
 import numpy as np
 import pytest
 
-from larray import Session, Axis, LArray, ndrange, isnan, larray_equal
 from larray.tests.common import assert_array_nan_equal, abspath
-from larray.util.misc import pickle
```
gdementen: I do not think pickle should be part of the public API so it should be imported from the sub module (and we should prevent it from ending up in larray top-level namespace

gdementen:
LGTM

PR #294, opened 2017-05-31
Refactor array model (issue 24)

alixdamman:
Need a preliminary review to know if I'm going to the right direction or not.

### DIFF
--- a/larray/viewer/model.py
+++ b/larray/viewer/model.py
@@ -176,6 +176,7 @@ class ArrayModel(QAbstractTableModel):
         self.minvalue = minvalue
         self.maxvalue = maxvalue
         # TODO: check that data respects minvalue/maxvalue
+        self._set_changes()
         self._set_data(data, bg_gradient=bg_gradient, bg_value=bg_value)
 
     def get_format(self):
@@ -188,12 +189,17 @@ class ArrayModel(QAbstractTableModel):
         return self._data2D
 
     def set_data(self, data, changes=None, bg_gradient=None, bg_value=None):
-        self._set_data(data, changes, bg_gradient, bg_value)
+        self._set_data(data, bg_gradient, bg_value)
+        self._set_changes(changes)
         self.reset()
 
-    def _set_data(self, data, changes=None, bg_gradient=None, bg_value=None):
+    def _set_changes(self, changes=None):
         if changes is None:
             changes = {}
+        assert isinstance(changes, dict)
+        self.changes = changes
+
+    def _set_data(self, data, bg_gradient=None, bg_value=None):
         if data is None:
             data = np.empty((0, 0), dtype=np.int8)
         la_data = la.aslarray(data)
@@ -220,9 +226,6 @@ class ArrayModel(QAbstractTableModel):
             self.color_func = np.real
         self.bg_gradient = bg_gradient
         self.bg_value = bg_value
-
-        assert isinstance(changes, dict)
-        self.changes = changes
 
         # get 2D shape + xlabels + ylabels
         if la_data.ndim == 0:
@@ -312,6 +315,16 @@ class ArrayModel(QAbstractTableModel):
         labels = self._position_to_labels(position)
         axes_ids = list(self.la_data.axes.ids)
         return dict(zip(axes_ids, labels))
+
+    def _dict_axes_ids_labels_to_position(self, dkey):
+        # transform (axis:label) dict key to positional ND key
+        try:
+            index_key = self.la_data._translated_key(dkey)
+        except ValueError:
+            return None
+        # transform positional ND key to positional 2D key
+        strides = np.append(1, np.cumprod(self.la_data.shape[1:-1][::-1]))[::-1]
+        return (index_key[:-1] * strides).sum(), index_key[-1]
 
     def columnCount(self, qindex=QModelIndex()):
         """Return array column number"""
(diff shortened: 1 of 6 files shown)


larray/viewer/view.py
```diff
@@ -131,14 +131,26 @@
...
-                                 larray_to_array_and_labels, ndarray_to_array_and_labels)
+from larray.viewer.model import ArrayModel, _, get_font, from_qvariant, to_qvariant, is_float, is_number
 
 from larray.viewer.combo import FilterComboBox, FilterMenu
 import larray as la
 
 PY2 = sys.version[0] == '2'
 
+
+# def _table_to_larray_indexes(self, row, column):
```
gdementen: shouldn't this block of commented code be removed?

larray/viewer/model.py
```diff
@@ -227,14 +223,27 @@ def _set_data(self, data, xlabels, ylabels, changes=None, bg_gradient=None, bg_v
...
-            xlabels = [[], []]
-        self.xlabels = xlabels
-        if ylabels is None:
-            ylabels = [[]]
+
+        # convert input data to LArray if not
+        # add extra dimension if 1D
+        la_data = la.aslarray(data)
+        if la_data.ndim == 1:
+            la_data = la_data.reshape([la.Axis(1), la_data.axes[0]])
```
gdementen: I am a bit nervous about this. I would prefer to keep the larray as it was given and reshape the underlying numpy data (in the "Convert LArray data to a 2D Numpy array" step below) so that we can export it directly without it having an extra dimension

larray/viewer/model.py
```diff
@@ -190,30 +186,33 @@ def get_data(self):
...
         if changes is None:
             changes = {}
         if data is None:
-            data = np.empty(0, dtype=np.int8).reshape(0, 0)
-        if data.dtype.names is None:
-            dtn = data.dtype.name
+            data = np.empty((0, 0), dtype=np.int8)
+
+        # convert input data to LArray if not
+        if not isinstance(data, (np.ndarray, pd.DataFrame, la.LArray)):
```
gdementen: Can't we just let aslarray handle this (ie remove the explicit test)? just say that we accept anything which can be converted to an larray (basically anything I think)?

larray/viewer/model.py
```diff
@@ -277,6 +288,19 @@ def set_format(self, format):
         self._format = format
         self.reset()
 
+    def _index_to_position(self, index):
+        """
+        Cell at position (0, 0) contains the first data.
```
gdementen: the first data cell? (data is not either an adjective or uncountable noun in English)

larray/viewer/model.py
```diff
@@ -227,16 +226,28 @@ def _set_data(self, data, xlabels, ylabels, changes=None, bg_gradient=None, bg_v
...
-        self.xlabels = xlabels
-        if ylabels is None:
-            ylabels = [[]]
-        self.ylabels = ylabels
-
-        self.total_rows = self._data.shape[0]
-        self.total_cols = self._data.shape[1]
+
+        if la_data.ndim == 0:
+            axes = la.AxisCollection((1, 1))
```
gdementen: modifying axes for 0d and 1d arrays like this means they will display exactly like 2d arrays, which will be ugly and confusing, especially because of the extra filter button(s). This wasn't the case previously.

larray/viewer/model.py
```diff
@@ -227,16 +226,28 @@ def _set_data(self, data, xlabels, ylabels, changes=None, bg_gradient=None, bg_v
...
+        # set xlabels
+        self.xlabels = [axes.display_names, axes.labels[-1]]
+
+        # set ylabels
+        otherlabels = axes.labels[:-1]
+        prod = Product(otherlabels)
+        self.ylabels = [_LazyNone(len(prod) + 1)] + [_LazyDimLabels(prod, i) for i in range(len(otherlabels))]
+
+        # set data (reshape to a 2D array if not)
+        self._data = la_data.data.reshape(new_shape)
```
gdementen: rename _data to _data2d for more clarity?

larray/viewer/model.py
```diff
@@ -118,15 +119,11 @@ class ArrayModel(QAbstractTableModel):
 
     Parameters
     ----------
-    data : 2D NumPy array, optional
-        Input data (2D array).
+    data : LArray or NumPy array, optional
```
gdementen: or dataframe, or any array-like? actually anything which can be converted to an LArray

larray/viewer/model.py
```diff
@@ -222,21 +219,47 @@ def _set_data(self, data, xlabels, ylabels, changes=None, bg_gradient=None, bg_v
...
-        self.ylabels = ylabels
-
-        self.total_rows = self._data.shape[0]
-        self.total_cols = self._data.shape[1]
+    @property
+    def filtered_data(self):
+        return self.la_data[self.current_filter]
+
+    def _set_labels_and_data_to_display(self):
+        la_data = self.filtered_data if self.current_filter else self.la_data
```
gdementen: using self.filtered_data in all cases seems simpler

larray/viewer/model.py
```diff
@@ -277,6 +310,134 @@ def set_format(self, format):
...
+                return None
+        # transform local dictionary key to local positional 2D key
+        return self._dict_axes_ids_labels_to_position(dkey)
+
+    def change_filter(self, axis, indices):
+        # must be done before changing self.current_filter
+        self.update_global_changes()
+        cur_filter = self.current_filter
+        axis_id = self.la_data.axes.axis_id(axis)
+        # if index == 0:
```
gdementen: kill comment?

alixdamman:
requires a "squash and merge"

PR #298, opened 2017-06-01
implemented Session.to_globals (closes #276)

### LINKED ISSUE #276: implement function load_arrays() (to use in viewer only)
It appears that some users would like to have a function that does same thing as edit() does: 
search for all existing LArray objects and create a Session + save it.

Same with a 'load_arrays' function : load a Session and put all arrays in the current namespace.

gdementen:


alixdamman:
Is it possible to check if we are in a IPython environment inside the function? 
I would like to raise a warning to discourage users to use it outside a IPython environment. 
I guess this is not possible.

gdementen:
I think it is possible, but I do not like to arbitrarily limit users. It is important to warn them in the docstring but if they prefer to use that technique anyway I would let them. There might be cases where the advantages outweigh the drawbacks.

alixdamman:
OK then.

PR #300, opened 2017-06-02
fix #296 : to_excel creates a new file by default if it does not exist

### LINKED ISSUE #296: to_excel does not create file by default if it does not exist in 0.23
```python
>>> ndtest(3).to_excel('out.xlsx')
```
does not work.

by default, to_excel should create the file if it does not exist, update the file (add or replace the sheet) if it exist. 

We also need a test for this (file does not exist) because. We did not notice we broke it and broke the doctests (which are marked as skip).

alixdamman:
+ updated test_to_excel_xlwings

### DIFF
--- a/larray/core/array.py
+++ b/larray/core/array.py
@@ -5673,8 +5673,6 @@ class LArray(ABCLArray):
         if engine == 'xlwings':
             from larray.io.excel import open_excel
 
-            wb = open_excel(filepath, overwrite_file=overwrite_file)
-
             close = False
             new_workbook = False
             if filepath is None:
@@ -5685,6 +5683,10 @@ class LArray(ABCLArray):
                     if not os.path.isfile(filepath):
                         new_workbook = True
                     close = True
+            if new_workbook or overwrite_file:
+                new_workbook = overwrite_file = True
+
+            wb = open_excel(filepath, overwrite_file=overwrite_file)
 
             if new_workbook:
                 sheet = wb.sheets[0]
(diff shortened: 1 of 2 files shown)


larray/core/array.py
```diff
@@ -5685,17 +5683,22 @@ def to_excel(self, filepath=None, sheet_name=None, position='A1',
                     if not os.path.isfile(filepath):
                         new_workbook = True
                     close = True
+            if overwrite_file:
```
gdementen: It seems more intuitive to me to invert the test:
```python
if new_workbook:
    overwrite_file = True
```
As an added benefit (assuming it works), in that case you can write a single
```python
wb = open_excel(filepath, overwrite_file=overwrite_file)
```
line and avoid modifying the rest of the code. I **think** it would yield the same result.

alixdamman: In that case, the test 'test_to_excel_xlwings' fails at line 
a2.to_excel(fpath, overwrite_file=True, engine='xlwings')
because the file already exists and new_workbook == False. 
What happens in that line is the creation a second sheet Sheet2 in which a2 is dumped.

alixdamman: Solution is to replace 
if new_workbook:
    sheet = wb.sheets[0]
by
if overwrite_file:
    sheet = wb.sheets[0]

gdementen: but in that case new_workbook is no longer used...

alixdamman: And should be fully replaced by overwrite_file, that's what you mean?

gdementen:
LGTM

PR #305, opened 2017-06-08
fix #288 + #302 : create_sequential + LArray attributes autocomplete

### LINKED ISSUE #288: create_sequential fails for array init, scalar inc
```python
>>> create_sequential('b=b0..b2', ndtest(3), inc=1)
TypeError                                 Traceback (most recent call last)

c:\users\gdm\devel\larray\larray\core\array.py in create_sequential(axis, initial, inc, mult, func, axes, title)
   6954             # stop is not included
   6955             stop = initial + inc * len(axis)
-> 6956             data = np.arange(initial, stop, inc)
   6957             res[:] = LArray(data, axis)
   6958         # inc only (other scalar)

c:\users\gdm\devel\larray\larray\core\array.py in __float__(self)
   5137 
   5138     def __float__(self):
-> 5139         return self.data.__float__()
   5140 
   5141     def divnot0(self, other):

TypeError: only length-1 arrays can be converted to Python scalars
```

### LINKED ISSUE #302: LArray attributes are not autocompleted
`LArray.__dir__()` should contain them (ie axes, data, title)

We should add `self.__dict__.keys()`

alixdamman:
contains only fix for issue #302 for now.

### DIFF
--- a/larray/core/array.py
+++ b/larray/core/array.py
@@ -967,7 +967,7 @@ class LArray(ABCLArray):
 
     def __dir__(self):
         names = set(axis.name for axis in self.axes if axis.name is not None)
-        return list(set(dir(self.__class__)) | names)
+        return list(set(dir(self.__class__)) | set(self.__dict__.keys()) | names)
 
     def _ipython_key_completions_(self):
         return list(chain(*[list(labels) for labels in self.axes.labels]))
--- a/larray/core/array.py
+++ b/larray/core/array.py
@@ -814,11 +814,7 @@ class LArray(ABCLArray):
           F  10  11  12
     """
 
-    def __init__(self,
-                 data,
-                 axes=None,
-                 title=''   # type: str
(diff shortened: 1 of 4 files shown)


larray/core/array.py
```diff
@@ -6841,6 +6837,18 @@ def create_sequential(axis, initial=0, inc=None, mult=1, func=None, axes=None, t
     >>> initial
     sex  M  F
          3  4
+    >>> create_sequential(year, initial, 1)
+    sex\\year  2016  2017  2018  2019
+           M     3     4     5     6
+           F     4     5     6     7
+    >>> create_sequential(year, initial, 1.0)
```
gdementen: this is one example too much IMO. It does not really help understanding what the function does. We probably want to have this example (using a float argument) as a unit test though.

gdementen:
LGTM

PR #308, opened 2017-06-09
compare names (issues 306 & 307)

gdementen:


larray/viewer/api.py
```diff
@@ -65,25 +65,33 @@ def get_title(obj, depth=0, maxnames=3):
...
-        Defaults to the collection of all local variables where
-        the function was called.
+        Object to visualize. If string, array(s) will be loaded from the file given as argument.
+        Defaults to the collection of all local variables where the function was called.
     title : str, optional
-        Title for the current object.
-        A default one is generated if not provided.
+        Title for the current object. Defaults to the name of the first object found in the caller namespace which
+        corresponds to `obj` (it will use a combination of the 3 first names if several names correspond to the same
+        object).
```
alixdamman: Remember that, if a MappingEditor is created, a new title will be generated by the method `update_title` of MappingEditor which looks like 'filepath - array's name : axes_info - passed title'.

larray/viewer/api.py
```diff
@@ -155,7 +155,26 @@ def view(obj=None, title='', depth=0):
 
 
 def compare(*args, **kwargs):
+    """
```
alixdamman: I wonder if it would not be a good idea to allow compare(a1=a1, a2=a2) ?

larray/viewer/api.py
gdementen: What you describe is IMO the window title. This is only the "object" title (ie the "passed title" part), so I do not think this is incorrect. It could be confusing for the user, but I do not want to spend too much time for such a detail. Do you have an alternative proposal which is not too complex?

alixdamman: Seems OK like this for me.

larray/viewer/api.py
gdementen: That's nice indeed, but that cannot work in all cases (spaces or special character in names etc.) so that cannot be the only way (and I am a bit tired of our habit to provide 6 different ways to do the same thing).

maybe a compromise would be compare({'a1': a1, 'a2': a2})

alixdamman: If users ask, we will think about it. In the meantime, let's go with this PR as it is.

gdementen:
rebased

PR #309, opened 2017-06-09
miscexcel

gdementen:
fixes for a few obscure bugs in excel Sheet handling


larray/io/excel.py
```diff
@@ -253,14 +253,34 @@ def __repr__(self):
...
 
 
+    def _fill_slice(s, length):
+        """
+        replaces a slice None bounds by actual bounds.
+
+        Parameters
+        ----------
+        k : slice
+            slice to replace
```
alixdamman: 's' instead of 'k'

Issue #293, opened 2017-05-31
Session.save should overwrite file by default


gdementen:
but we should keep the possibility to "update" the file like we have now. Probably as a different method: Session.update_file()?

Issue #153, opened 2017-03-16
dump/load sessions with non-arrays


gdementen:
I am thinking mostly of groups and scalars.

This would be especially important for users of the standalone interface (see #88).

When using the standalone interface, they will probably want to also dump functions defined within the console in the same file, so that they can start working exactly where they left of, but we could use a specific format for that (e.g pickle the whole session), but in that case, using `from larray import *` will become problematic.

gdementen:
This might be a crazy idea, but, if we want to save python functions inside an .xlsx, we could place them inside macros, potentially wrapped inside xlwings RunPython so that they are even executable from within Excel. For maintainability and collaboration it is probably better to save the python file externally (so that you can put it on a VCS, but there might be cases where this could be useful.

gdementen:
We might want to use something like https://github.com/h5io/h5io or https://github.com/telegraphic/hickle

gdementen:
Or, we could roll our own stuff. If we go that route and want to continue using Pandas HDFStore, this might be useful:

http://stackoverflow.com/questions/29129095/save-additional-attributes-in-pandas-dataframe/29130146#29130146

```python
In [172]: df = pd.DataFrame(np.random.randn(8,3))
In [173]: store = pd.HDFStore('test.h5')
In [174]: store.put('df',df)
# you can store an arbitrary python object via pickle
In [175]: store.get_storer('df').attrs.my_attribute = dict(A = 10)
In [176]: store.get_storer('df').attrs.my_attribute
Out[176]: {'A': 10}
```

gdementen:
Here are links to stuff that might help (or not). This is a dump of all the browser windows I have been keeping open for months. ;-)

* standard pickle (which, if using a low enough protocol are guaranteed to work across Python versions). Marshal (used to create .pyc) on the other hand is not. 

* to pickle tracebacks: https://github.com/ionelmc/python-tblib
* better pickles (pickle more kinds of objects than standard pickle):
  - cloudpickle: https://github.com/cloudpipe/cloudpickle
  - dill: 
     * https://pypi.python.org/pypi/dill
     * http://trac.mystic.cacr.caltech.edu/project/pathos/wiki/dill.html
     * of note is: `dill also provides the capability to save and load python interpreter sessions`

* jsonpickle (serialize most python objects as json -- but with similar security implications than pickle!). The format is nicer than pickle, but I am unsure it brings any value to the table, if we do not plan to exchange sessions across languages.
  https://jsonpickle.github.io/

[...]

PR #312, opened 2017-06-13
fix #291 + #293 + #313 : Session.save (0D arrays + Excel + overwrite file by default) 

### LINKED ISSUE #291: Session.save fails when it contains 0d arrays
because `LArray(0).to_frame()` fails

alixdamman:
until #153 is fixed.

### DIFF
--- a/doc/source/changes/version_0_24.rst.inc
+++ b/doc/source/changes/version_0_24.rst.inc
@@ -18,8 +18,26 @@
      a0   0   1
      a1   2   3
 
-* added new boolean argument 'overwrite' to Session.save method. If True, overwrite existing file if any.
-  Otherwise, update the file. Defaults to True. Closes issue:`293`.
+* added new boolean argument 'overwrite' to Session.save, Session.to_hdf, Session.to_excel and Session.to_pickle
+  methods (closes issue:`293`). If overwrite=True, the file is removed and replaced by a new one if it already existed.
+  This is the default behavior. If overwrite=False, its content is updated :
+
+    >>> arr1, arr2, arr3 = ndtest((2, 2)), ndtest(4), ndtest((3, 2))
+    >>> s = Session([('arr1', arr1), ('arr2', arr2), ('arr3', arr3)])
+
+    >>> # save arr1, arr2 and arr3 in file output.h5
+    >>> s.save('output.h5')
+
+    >>> # replace arr1 and create arr4 + put them in an second session
+    >>> arr1, arr4 = ndtest((3, 3)), ndtest((2, 3))
+    >>> s2 = Session([('arr1', arr1), ('arr4', arr4)])
+
+    >>> # replace arr1 and add arr4 in file output.h5
+    >>> s2.save('output.h5', overwrite=False)
+
+    >>> # erase content of 'output.h5' and save only arrays contained in the second session
+    >>> s2.save('output.h5')
+
 
 Miscellaneous improvements
 --------------------------
--- a/doc/source/changes/version_0_24.rst.inc
+++ b/doc/source/changes/version_0_24.rst.inc
@@ -89,3 +89,5 @@ Fixes
 
 * fixed Session.save and Session.to_excel failing to create new Excel file
   (it only worked if the file already existed). Closes issue:`313`.
+
+* fixed Session.load(file, engine='pandas_excel') : axes were considered as anonymous.
--- a/doc/source/changes/version_0_24.rst.inc
+++ b/doc/source/changes/version_0_24.rst.inc
@@ -18,6 +18,8 @@
      a0   0   1
      a1   2   3
 
+* added new boolean argument 'overwrite' to Session.save method. If True, overwrite existing file if any.
+  Otherwise, update the file. Defaults to True. Closes issue:`293`.
 
 Miscellaneous improvements
 --------------------------
@@ -63,3 +65,9 @@ Fixes
 * fixed getting float data instead of int when converting an Excel Sheet or Range to an larray or numpy array.
 
 * fixed some warning messages to point to the correct line in user code.
+
+* fixed crash of Session.save method when it contains a 0D array.
+  0D arrays are now skipped when saving a session (closes issue:`291`).
+
+* fixed Session.save and Session.to_excel failing to create new Excel file
+  (it only worked if the file already existed). Closes issue:`313`.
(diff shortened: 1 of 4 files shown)


larray/io/session.py
```diff
@@ -114,6 +115,10 @@ def dump_arrays(self, key_values, *args, **kwargs):
         display = kwargs.pop('display', False)
         self._open_for_write()
         for key, value in key_values:
+            if isinstance(value, ABCLArray) and value.ndim == 0:
+                if display:
+                    print('Cannot dump {}. Dumping 0D arrays is not supported currently.'.format(key))
```
gdementen: currently not supported

alixdamman:
Did you run pytest after commit https://github.com/liam2/larray/pull/309/commits/1286d6715fb7244f60762c2d75c86078319e1064 ? Because you removed the "__larray__" method from Range class but the call from "__getattr__" a dew lines below.

gdementen:
Oops, nice catch! I thought I did run the tests. Maybe I will need to add another one.

doc/source/changes/version_0_24.rst.inc
```diff
@@ -57,3 +57,9 @@ Fixes
 * fixed the array() method on excel.Sheet returning float labels when int labels are expected.
 
 * fixed getting float data instead of int when converting an Excel Sheet or Range to an larray or numpy array.
+
+* fixed crash of Session.save method when it contains a 0D array.
+  0D arrays are skiped when saving a session (closes issue:`291`).
+
+* fixed crash of Session.save method when trying to save a session in an Excel that does not exist.
+  A new Excel file is created if it does not exist yet (closes issue:`313`).
```
gdementen: isn't crash a too strong word for something which failed with a relatively nice error message? I would write something like:

    fixed Session.save and Session.to_excel failing to create new Excel files (it only worked if the file already existed). Closes ...

larray/tests/test_session.py
```diff
@@ -161,6 +162,8 @@ def test_xlsx_pandas_io(self):
     @pytest.mark.skipif(xw is None, reason="xlwings is not available")
     def test_xlsx_xlwings_io(self):
         fpath = abspath('test_session_xw.xlsx')
+        if os.path.isfile(fpath):
```
gdementen: I think we also need an explicit test where the file already existed.

alixdamman:
Do you want me to include issue #293 in this PR or another?

gdementen:
as you prefer.

alixdamman:
Not sure for issue #293

larray/tests/test_session.py
```diff
@@ -134,24 +136,39 @@ def test_names(self):
...
         s = Session()
         s.load(fpath, engine='pandas_excel')
         self.assertEqual(list(s.keys()), ['e', 'g', 'f'])
 
+        # update an array (overwrite=False)
+        s['e'] = self.e2
+        s.save(fpath, engine='pandas_excel', overwrite=False)
+        s.load(fpath, engine='pandas_excel')
+        self.assertEqual(list(s.keys()), ['e', 'g', 'f'])
+        #assert_array_nan_equal(s['e'], self.e2)
```
gdementen: why is this line commented?

larray/tests/test_session.py
```diff
@@ -161,12 +178,21 @@ def test_xlsx_pandas_io(self):
     @pytest.mark.skipif(xw is None, reason="xlwings is not available")
     def test_xlsx_xlwings_io(self):
         fpath = abspath('test_session_xw.xlsx')
+        # test save when Excel does not exist
```
gdementen: Excel file

larray/tests/test_session.py
```diff
@@ -134,24 +136,39 @@ def test_names(self):
...
+
         s = Session()
         s.load(fpath)
         # HDF does *not* keep ordering (ie, keys are always sorted)
         self.assertEqual(list(s.keys()), ['e', 'f', 'g'])
 
+        # update an array (overwrite=False)
+        s['e'] = self.e2
+        s.save(fpath, overwrite=False)
+        s.load(fpath)
```
gdementen: This test (and all similar tests below) is a bit misleading. It will pass even if the file is completely wiped out, because the arrays still exist in the session s and s.load(path) does not clear the session. Maybe we want to change that last point, but that would be for another PR/release.

larray/core/session.py
```diff
@@ -245,12 +245,14 @@ def save(self, fname, names=None, engine='auto', display=False, **kwargs):
         names : list of str or None, optional
             List of names of objects to dump. If `fname` is None, list of paths to CSV files.
             Defaults to all objects present in the Session.
-        engine : str, optional
+        engine : {'auto', 'pandas_csv', 'pandas_hdf', 'xlwings_excel', 'pickle'}, optional
             Dump using `engine`. Defaults to 'auto' (use default engine for
             the format guessed from the file extension).
+        overwrite: bool, optional
+            Whether or not to overwrite an existing file, if any. Only for non CSV files.
```
gdementen: Ignored for CSV files?

doc/source/changes/version_0_24.rst.inc
```diff
@@ -18,6 +18,8 @@
      a0   0   1
      a1   2   3
 
+* added new boolean argument 'overwrite' to Session.save method. If True, overwrite existing file if any.
```
gdementen: * we should also mention other "save" methods: to_excel, etc. which are also impacted by this change.
* I don't think it is clear enough that the default behaviour changed from overwrite=False to overwrite=True

larray/io/session.py
```diff
@@ -164,7 +169,7 @@ def list(self):
 
     def _read_array(self, key, *args, **kwargs):
         df = self.handle.parse(key, *args, **kwargs)
-        return df_aslarray(df)
+        return df_aslarray(df, raw=True)
```
gdementen: What's the relation with the rest? Is this an extra fix? If so it needs a test & changelog :)

alixdamman: This is an extra fix for an issue discovered when I tried to update the IO tests for Session ;)

gdementen: In fact, this is not entirely correct. We should use: raw=index_col is None. But I admit the chance users ever hit this case is close to zero :)

alixdamman: forgot changelog, sorry.

gdementen:
LGTM

Issue #212, opened 2017-04-11
rename create_sequential() to sequence()


gdementen:
something like `create_seq()` or `sequence()` or even `seq()`

gdementen:
@jehanneman suggests using sequential(), but he is not against sequence(). He mentions that zeros, ones etc also create an array and do not contain "create". 

My current favourite goes to sequence(). @alixdamman : what do you think?

alixdamman:
Using an adjective as function name is weird for me. I would go for sequence() also.

PR #316, opened 2017-06-14
rename create_sequential to sequence (#212) +small misc

gdementen:


larray/io/excel.py
```diff
@@ -446,7 +446,7 @@ def __dir__(self):
 
         def __getattr__(self, key):
             if hasattr(LArray, key):
-                return getattr(self.__larray__(), key)
+                return getattr(LArray(self.__array__()), key)
             else:
```
alixdamman: Could you do the same in the method __str__ below?

gdementen:
doh!

alixdamman:
Did you run pytest this time? (to test xlwings)

gdementen:
I run pytest everytime! But the flip side of the coin is that I rely too much on them. When they pass, I think it is ok. The Excel tests have a bad coverage so far...

alixdamman:
OK to merge

PR #317, opened 2017-06-14
prepare for release 0.24

gdementen:


gdementen:
I guess we could. I never tried but that would be a good thing to have if it works (one less file to update)

PR #323, opened 2017-06-16
fix #322 (viewer): documentation of function edit, view and compare

### LINKED ISSUE #322: viewer functions documentation does not appear in the rendered documentation
I guess this is because the doc is picked up from the "dummy" versions which are used when Qt is not available (because pyqt it is not installed on the readthedocs server?).

One way to fix this would be to copy the docstring to those functions (and change their signature) from the "real" functions.

alixdamman:
I moved the doc in init.py. The only problem now is that CTRL + Q will not show the doc in PyCharm unless we copy the doc of the functions edit, view and compare in both init.py and api.y (similar problem for open_excel).

### DIFF
--- a/doc/environment.yml
+++ b/doc/environment.yml
@@ -12,3 +12,5 @@ dependencies:
   - numpydoc
   - nbsphinx
   - pandoc
+  - pyqt
+  - qtpy
--- a/larray/viewer/api.py
+++ b/larray/viewer/api.py
@@ -67,6 +67,8 @@ def edit(obj=None, title='', minvalue=None, maxvalue=None, readonly=False, depth
     """
     Opens a new editor window.
 
+    Parameters
+    ----------
     obj : np.ndarray, LArray, Session, dict or str, optional
         Object to visualize. If string, array(s) will be loaded from the file given as argument.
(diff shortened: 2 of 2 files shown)


alixdamman:
Another solution would be to update the file environment.yml in directory doc/ and include Qt (+ xlwings).

gdementen:
This is the usual issue with PyCharm... It is a known issue but I do not think it will be fixed anytime soon.
https://youtrack.jetbrains.com/issue/PY-19384

Anyway, I think it is more important to have the manual correct than the Ctrl+Q of PyCharm.

alixdamman:
As you wish.

gdementen:
having it in both cases is obviously better, so adding Qt to the environment would probably be better. I doubt xlwings is available on linux though (I guess readthedocs server is linux-based).

alixdamman:
Preview of the documentation for this PR : http://larray-test.readthedocs.io/en/documentation/api.html

alixdamman:
Seems OK : http://larray-test.readthedocs.io/en/documentation/api.html#viewer

gdementen:
LGTM, thanks!

PR #327, opened 2017-06-21
miscellaneous changes

alixdamman:


larray/viewer/view.py
```diff
@@ -1316,8 +1316,9 @@ def setup_menu_bar(self):
                                           triggered=self.close))
 
         help_menu = menu_bar.addMenu('&Help')
-        help_menu.addAction(create_action(self, _('Online documentation'), shortcut="Ctrl+H",
-                                          triggered=self.open_documentation))
+        help_menu.addAction(create_action(self, _('Online &Documentation'), shortcut="Ctrl+H", triggered=self.open_documentation))
+        help_menu.addAction(create_action(self, _('&Getting Started'), triggered=self.open_tutorial))
```
gdementen: why not name this "Tutorial" or "Online Tutorial" instead of "Getting Started"?

larray/viewer/view.py
```diff
@@ -1316,8 +1316,9 @@ def setup_menu_bar(self):
...
+        help_menu.addAction(create_action(self, _('Objects and &Functions'), triggered=self.open_api_documentation))
```
gdementen: Objects and Functions (API) &Reference?

gdementen: Online?

larray/viewer/view.py
alixdamman: IMO, Getting Started is the equivalent of 10 minutes to Pandas which does not correspond to the "tutorial". As I see it, the getting started should be read in maximum 10 minutes and just present the main features of LArray

alixdamman: Well, here open_tutorial will open the tutorial. So you're right. Sorry

larray/viewer/view.py
alixdamman: Our main users don't know what API means. That's why I didn't mentioned it. But OK, I could add in ( ).

larray/viewer/view.py
alixdamman: I mixed up with PR #329

alixdamman:
updated

gdementen:
LGTM

gdementen:
I know... But for those who know what that means, it is clearer. Or at
least I usually scan for both "API" and "Reference" keywords. Here there
was neither.

PR #333, opened 2017-07-03
converted LArray_intro.ipynb into LArray_intro.rst

alixdamman:
This PR is only dedicated on the conversion of the notebook tutorial into a rst file. 
I know that the content of the tutorial should be changed but IMO this should be done in another PR.

Result can be seen here: http://larray-test.readthedocs.io/en/misc/tutorial/LArray_intro.html
It seems there is a problem between pyqt5 (maybe pyqt4) and matplotlib (at least on readthedocs)

Examples of how to use the ipython directive are here: https://github.com/pydata/xarray/tree/master/doc
Documentation: http://matplotlib.org/sampledoc/ipython_directive.html


alixdamman:
Once you agree on the result, let me clean the history before to merge. thanks.

alixdamman:
Note: don't be afraid if I used 

  > In [1]: ...

with verbatim. The "1" is automatically replaced by the correct number when the doc is compiled.

PR #334, opened 2017-07-05
fixed arr['axis_name[ambiguouslabel]'] (#331)

gdementen:


larray/core/axis.py
```diff
@@ -1299,7 +1305,7 @@ def __eq__(self, other):
         if not isinstance(other, list):
             other = list(other)
         return len(self._list) == len(other) and \
-               all(a.equals(b) for a, b in zip(self._list, other))
+            all(a.equals(b) for a, b in zip(self._list, other))
```
alixdamman: why not join the 2 lines?

gdementen: good question :)

PR #344, opened 2017-07-18
implemented Session.apply

gdementen:


larray/core/session.py
```diff
@@ -781,6 +780,50 @@ def compact(self, display=False):
             new_items.append((k, compacted))
         return Session(new_items)
 
+    def apply(self, func, *args, **kwargs):
+        """
+        Apply function `func` on all elements of the session and return a new session.
```
alixdamman: We do not have to think about that right now but one day elements of a session could not be (L)arrays.
Here, you could replace elements by arrays, no?

larray/core/session.py
```diff
@@ -781,6 +780,50 @@ def compact(self, display=False):
...
+
+        Parameters
+        ----------
+        func : function
+            Function to apply to each element of the session. It should take a single `element` argument and return
```
alixdamman: s/element/array ?

larray/core/session.py
```diff
@@ -781,6 +780,50 @@ def compact(self, display=False):
...
+        ...     return element + 1
+        >>> sess2 = sess1.apply(increment)
+        >>> sess2.arr1
+        a  a0  a1
+            1   2
+        >>> sess2.arr2
+        a  a0  a1  a2
+            1   2   3
+        """
+        return Session([(k, func(v, *args, **kwargs)) for k, v in self.items()])
```
alixdamman: Session([ ... if isinstance(v, LArray)]) ?

larray/core/session.py
gdementen: Well, it is already the case right now.

alixdamman: Yes but our current users do not know that, right?

larray/core/session.py
gdementen: Hmmm, that's an interesting option too. I cannot really tell which is better. That's the usual dilemma between an API which works in all cases and one that is most practical in the usual case.

larray/core/session.py
gdementen: ;-)

larray/core/session.py
alixdamman: Or we could an optional argument `filter` (containing a tuple of accepted types) to `apply`?

gdementen: I wouldn't do that, because if you want filter, you can do so using the filter method:
`sess.filter(kind=LArray).apply()`

Oh, now I realize I misread what you wrote. I thought you meant:
`return Session([(k, func(v, *args, **kwargs) if isinstance(v, LArray) else v) for k, v in self.items()])`
maybe having an array=True keyword which apply to arrays only by default would be the best solution?

alixdamman: Yes, something like that

gdementen:
rebased and added "kind" keyword argument which defaults to LArray

larray/core/session.py
```diff
@@ -808,6 +807,54 @@ def compact(self, display=False):
...
+        >>> def increment(array):
+        ...     return array + 1
+        >>> sess2 = sess1.apply(increment)
+        >>> sess2.arr1
+        a  a0  a1
+            1   2
+        >>> sess2.arr2
+        a  a0  a1  a2
+            1   2   3
+        """
```
alixdamman: add an example with a function with arguments?

alixdamman:
Already approved. Must be updated.

gdementen:
rebased & added example

PR #347, opened 2017-07-18
implemented unary ops on Session

gdementen:


doc/source/changes/version_0_25.rst.inc
```diff
@@ -102,6 +102,30 @@
     arr1 -> b, c, a
     arr2 -> b, a
 
+* implemented unary operations on Session, which means one can negate all arrays in a Session or take the absolute
+  value of all arrays in a Session without writing an explicit loop for that.
+
```
alixdamman: Available unary operations are `neg`, `pos`, `abs` and `invert`.

gdementen: I have never seen any use for pos so I would rather not mention it, as for invert it has almost no use for now (until I change Session.__eq__ to work element-wise instead of array-wise).

alixdamman: Sessions are not widely used for now (except to do Session(...).to_globals() + I/O), so I guess the example with abs is enough.

PR #352, opened 2017-08-08
implemented ipfp axes argument

gdementen:


doc/source/changes/version_0_25.rst.inc
```diff
@@ -137,13 +138,36 @@
...
+    >>> result = zeros(initial.axes)
+    >>> for year in initial.year:
+    ...     current = initial[year]
+    ...     # assume you have some targets for each year
+    ...     current_targets = [current.sum(x.a) + 1, current.sum(x.b) + 1]
+    ...     result[year] = ipfp(current_targets, current)
+
+  Now you can apply the procedure on all years at once, by telling you want to do the fitting procedure on the other
+  axes. This is a bit shorter to type, but this is also *much* faster.
+
```
alixdamman: "much faster" but bellow you say that you made ipfp 10 to 20% faster. This not "much" or are you talking about 2 different things?

doc/source/changes/version_0_25.rst.inc
```diff
@@ -137,13 +138,36 @@
...
+    >>> all_targets = [initial.sum(x.a) + 1, initial.sum(x.b) + 1]
+    >>> result = ipfp(all_targets, initial, axes=(x.a, x.b))
+
```
alixdamman: is it possible to write something like 

> ipfp(all_targets, initial, axes='a, b') 

I think some will try.

larray/extra/ipfp.py
```diff
@@ -39,34 +55,36 @@ def ipfp(target_sums, a=None, maxiter=1000, threshold=0.5, stepstoabort=10,
...
         'warn': return values computed up to that point and print a warning
         'raise': raise an exception (default)
     display_progress : False, True or 'condensed', optional
-        whether or not to display progress. Defaults to False.
-        if 'condensed' will display progress using a denser template (using one
+        Whether or not to display progress. Defaults to False.
+        If 'condensed' will display progress using a denser template (using one
         line per iteration).
 
     Returns
```
alixdamman: You should an example with axes argument.

doc/source/changes/version_0_25.rst.inc
gdementen: That's indeed 2 different things. It's 10-20% faster in the "normal" case.

doc/source/changes/version_0_25.rst.inc
gdementen: This is not supported at the moment. We could support this, but then I think there are other places (which ask for a list/collection of axes) where we should implement it.

gdementen: I will open an issue to do this as a followup.

doc/source/changes/version_0_25.rst.inc
alixdamman: Do not change. It's clear enough.

doc/source/changes/version_0_25.rst.inc
alixdamman: Thanks. Some users prefer (love) the string syntax.

larray/extra/ipfp.py
gdementen: Fair enough.

doc/source/changes/version_0_25.rst.inc
gdementen: opened issue #358

gdementen:
fixed

doc/source/changes/version_0_25.rst.inc
alixdamman: OK.

PR #353, opened 2017-08-09
more session I/O features

gdementen:
implemented Session().load(..., ignore_exceptions=True) and Session().load('data/*.csv')

### DIFF
--- a/larray/core/session.py
+++ b/larray/core/session.py
@@ -182,7 +182,7 @@ class Session(object):
 
     def load(self, fname, names=None, engine='auto', display=False, **kwargs):
         """
-        Loads array objects from a file.
+        Loads array objects from a file, or several .csv files.
 
         WARNING: never load a file using the pickle engine (.pkl or .pickle) from an untrusted source, as it can lead
         to arbitrary code execution.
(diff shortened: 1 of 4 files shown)


larray/core/session.py
```diff
@@ -216,6 +216,15 @@ def load(self, fname, names=None, engine='auto', display=False, **kwargs):
...
         >>> arr1, arr2, arr3 = s.values()                 # doctest: +SKIP
+
+        Using .csv files (assuming the same session as above)
+
+        >>> s.save('data')                                # doctest: +SKIP
+        >>> s = Session()                                 # doctest: +SKIP
+        >>> # load all .csv files starting with "output" in the data directory
+        >>> s.load('data')                                # doctest: +SKIP
+        >>> # or equivalently in this case
+        >>> s.load('data/arr*.csv')                       # doctest: +SKIP
```
alixdamman: Since doctests are skipped, corresponding tests should be updated.

gdementen: Fair enough... I wanted to get away without writing new tests ;-), but you are right... I will add them...

gdementen:
rebased and added tests

PR #363, opened 2017-08-23
Check labels in setitem 269

alixdamman:


larray/core/array.py
```diff
@@ -2201,13 +2199,12 @@ def __setitem__(self, key, value, collapse_slices=True):
...
-                        else Axis(1, axis.name)
-                        for axis, axis_key in zip(self.axes, translated_key)]
+                        else Axis(1, axis.name) for axis, axis_key in zip(self.axes, translated_key)]
             else:
-                axes = [axis.subaxis(axis_key)
-                        for axis, axis_key in zip(self.axes, translated_key)
+                axes = [axis.subaxis(axis_key) for axis, axis_key in zip(self.axes, translated_key)
                         if not np.isscalar(axis_key)]
             value = value.broadcast_with(axes)
+            value.axes.check_compatible(AxisCollection(axes))
```
gdementen: `value.axes.check_compatible(axes)` should work

alixdamman:
A new changelog file is also needed. "0.25.1" or "0.26"?

gdementen:
0.25.1. We will do a bug-fix only release

gdementen:
we also need to mention the viewer fixes

doc/source/changes/version_0_25_1.rst.inc
```diff
@@ -0,0 +1,33 @@
...
+    a  a0  a1  a2  a3
+        0   1   2   3
+    >>> arr['a2:'] = arr[':a1']
+    ...
+    ValueError: incompatible axes:
+    Axis(['a2', 'a3'], 'a')
+    vs
+    Axis(['a0', 'a1'], 'a')
+
+  is not allowed and raises a ValueError (closes :issue:`269`).
```
gdementen: * I consider not failing when it should have a (grave) bug (i.e. I'd move this to the fixes section). It is probably not obvious that it is an issue in this particular case (*) and I don't know how to make it obvious, but I have seen several occurrences of this silently corrupting user data. (*) and this is the reason I initially allowed this to work on purpose. 
* mentioning that it raises a ValueError does not seem useful (we see it just a few lines before). I would move the "issue closed mention" on the first line and kill this line entirely.
* I usually strip the traceback completely (i.e. remove even the "...")
* I think it might help users if you also mention how to circumvent this check (using drop_labels('a') or drop_labels() seems to be the cleanest way).

doc/source/changes/version_0_25_1.rst.inc
```diff
@@ -0,0 +1,33 @@
...
+
+
+Fixes
+-----
+
+* fixed error raised in viewer when switching between arrays.
```
gdementen: ... when a filter was set

gdementen:
also need to mention your other viewer improvement (new)

PR #374, opened 2017-08-25
fixed array.points[key] = value when broadcasting is necessary (closes #368)

### LINKED ISSUE #368: arr.points[key] = value does not broadcast LArray values
```python
>>> arr = ndtest((2, 3, 4))
>>> arr.points['b0,b1,b2', 'c0,c1,c2'] = arr['b0', 'c0']
ValueError: shape mismatch: value array of shape (2,) could not be broadcast to indexing result of shape (3,2)
```

gdementen:


alixdamman:
LGTM

PR #375, opened 2017-08-25
Make int point point int works in selection 350

alixdamman:
Need a new unittest?


alixdamman:
updated and added a simple test

Issue #376, opened 2017-08-28
make combine_axes and boolean selection use same sep


gdementen:
combine_axes defaults to `_` and boolean selection defaults to `,` and this makes them incompatible by default. Unsure which we should choose (probably `_` to not piss off Bernard ;-)). Unsure if issuing a warning first (for a few release) will be possible/practical.

This is required for #369.

alixdamman:
I would go for `_` (I think `_` has no particular meaning in the case of the "string" syntax of LArray). 
I guess boolean filtering will return a "MultiIndex" version of an LArray in the future.

gdementen:
For axes names `_` is mostly equivalent to `,` (at least at this point -- until we implement #358). But for actual labels, I think `,` was superior since it had the nice property that you could (\*) use a label of the combined axis as a key in the split array and it worked. 

(\*) actually, it did not work but it could be made to work, so that does not change my point. Depending on how fast we implement #28, spending much time on this might not be worth it.

PR #378, opened 2017-08-28
fixed array.split_axis when combined labels are not sorted by the first part then second part (closes #364)

### LINKED ISSUE #364: split_axis broken when combined labels are not sorted by the first part then second part

gdementen:


larray/core/array.py
```diff
@@ -6591,7 +6591,7 @@ def split_axis(self, axis, sep='_', names=None, regex=None):
          a0   0   1   2
          a1   3   4   5
         """
-        return self.reshape(self.axes.split_axis(axis, sep, names, regex))
+        return self.sort_axis(axis).reshape(self.axes.split_axis(axis, sep, names, regex))
```
alixdamman: why not handle this as an option which is True by default. Users may unsorted axis after split?

gdementen: because it mangles the data and I do not see that as an interesting option ;-)

alixdamman: OK

gdementen:
my fix is broken...

gdementen:
Here is the proper fix. The remaining problem concerns #376.

larray/core/axis.py
```diff
@@ -274,6 +274,64 @@ def extend(self, labels):
...
 
+    def split(self, sep='_', names=None, regex=None, return_labels=False):
+        """Split axis and returns a list of Axis.
+
+        Parameters
+        ----------
+        sep : str, optional
+            Delimiter to use for splitting. Defaults to '_'.
+
+            When `regex` is provided, the delimiter is only used on `names` if given as one string or on axis name if
```
alixdamman: why a blank line? To force line return?

gdementen: no idea. I think it was that way where I copy-paste that block from :) (or I introduced it by mistake)

larray/core/axis.py
```diff
@@ -274,6 +274,64 @@ def extend(self, labels):
...
+            `names` is None.
+        names : str or list of str, optional
+            Names of resulting axes. Defaults to None.
+        regex : str, optional
+            Use regex instead of delimiter to split labels. Defaults to None.
+        labels : bool, optional
+            Whether or not split labels must be returned (as a tuple of tuples). These labels are suitable for indexing
+            via array.points[labels]. Defaults to False.
```
alixdamman: Do you want to open a new issue to add that option to LArray.split_axes?

gdementen: LArray.split_axes? I don't see a use case. In AxisCollection.split_axes, that **might** make the code of LArray.split_axes nicer/smaller, but I doubt the array.points[] = value with multiple labels (for the several axes) will work. I haven't thought much about that though.

alixdamman:
except for the extra blank line in docstring of `Axis.split`, LGTM

PR #379, opened 2017-08-28
updated /viewer/__init__.py + /doc/environment.yml file ( fix #322 )

### LINKED ISSUE #322: viewer functions documentation does not appear in the rendered documentation

alixdamman:


alixdamman:
This PR could wait for 0.25.2 or 0.26

gdementen:
why the do not merge?

alixdamman:
To be sure to not break something. But indeed could be solved in 0.25.2 if any problem.

PR #381, opened 2017-08-29
Fix from string 372

alixdamman:


doc/source/changes/version_0_25_2.rst.inc
```diff
@@ -0,0 +1,15 @@
+Miscellaneous improvements
+--------------------------
+
+* ...
+
+Fixes
+-----
+
+* fixed extracting LArray object using `from_string` function when second axis is implicit
```
gdementen: It is much larger than just from_string, which most users do not know/care about. My test case used from_string so that it was self-contained, but the initial bug report was using a .csv file and I suspect excel would suffer from this too (and depending on the backend, this fix might not affect it because the code path is very different with the xlwings backend).

larray/io/array.py
```diff
@@ -77,22 +77,17 @@ def df_aslarray(df, sort_rows=False, sort_columns=False, raw=False, parse_header
...
-            index_columns = [df.columns[i] for i in range(pos_last + 1)]
-            # TODO: we should pass a flag to df_aslarray so that we can use inplace=True here
-            # df.set_index(index_columns, inplace=True)
-            df = df.set_index(index_columns)
+        # This is required to handle int column names (otherwise we can simply use column positions in set_index).
+        # This is NOT the same as df.columns[list(range(...))] !
+        index_columns = [df.columns[i] for i in range(pos_last + 1)]
+        # TODO: we should pass a flag to df_aslarray so that we can use inplace=True here
+        # df.set_index(index_columns, inplace=True)
+        df = df.set_index(index_columns)
```
gdementen: I am surprised this works for 1D arrays. I wonder if the result wouldn't be subtly different, so I am nervous to introduce this in a bugfix release. At a minimum, we will need to check how user models behave with this change, so maybe it would be wise to postpone this to 0.26?

alixdamman:
Need to move added lines in changelog 0.25.1 file to changelog 0.26 file. Do not merge!

alixdamman:
This PR may be affected by the version of Pandas. Needs more investigation.

gdementen:
is the do-not-merge still applicable?

alixdamman:
I would like to add more tests in test_df_aslarray()

alixdamman:
New tests explain why #427 fails (aslarray(1x1 or 1xN Dataframe)).

alixdamman:
Still need to update changelog

doc/source/changes/version_0_26.rst.inc
```diff
@@ -254,5 +254,14 @@ Fixes
 
 * fixed array creation with axis(es) given as string containing only one label (axis name and label were inversed).
 
+* fixed extracting LArray object using `from_string` function when second axis is implicit
```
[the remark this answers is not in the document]
alixdamman: why mentioning `dialect=classic` if it's not implemented?

doc/source/changes/version_0_26.rst.inc
```diff
@@ -254,5 +254,20 @@ Fixes
 
 * fixed array creation with axis(es) given as string containing only one label (axis name and label were inversed).
 
+* fixed extracting LArray object from a CSV or Excel file when second axis is implicit (closes :issue:`372`).
+  For example, let say we want to read a CSV file 'pop.csv' with content ::
```
gdementen: s/let say/let us say/ or /let us suppose/ or /let's say/...
s/with content/with the following content (indented for clarity)/

doc/source/changes/version_0_26.rst.inc
```diff
@@ -254,5 +254,20 @@ Fixes
 
 * fixed array creation with axis(es) given as string containing only one label (axis name and label were inversed).
 
+* fixed extracting LArray object from a CSV or Excel file when second axis is implicit (closes :issue:`372`).
```
gdementen: s/extracting LArray object/reading an array/?
s/second axis is implicit/the columns axis is not explicitly named (via )/?

doc/source/changes/version_0_26.rst.inc
```diff
@@ -254,5 +254,20 @@ Fixes
...
+  For example, let say we want to read a CSV file 'pop.csv' with content ::
+
+  sex,2015,2016
```
gdementen: It would be more readable if you indented the table, as in:
```
sex, 2015, 2016
  M,   12,   10
  F,   11,   13
```
I am unsure whether or not we should keep the commas. I guess we should leave them in. See also comment above.

doc/source/changes/version_0_26.rst.inc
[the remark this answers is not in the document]
gdementen: the to_csv(dialect='classic') **is** implemented, read_csv(dialect='classic') is not. In other words, you fixed reading back a file created using to_csv(dialect='classic').

Issue #385, opened 2017-08-30
label arg of with_total ignored when using a group  


alixdamman:
```python
        >>> arr = ndtest((3, 3))
        >>> arr
        a\b  b0  b1  b2
         a0   0   1   2
         a1   3   4   5
         a2   6   7   8
```

Expected behavior
```python
      >>> arr.with_total('a0,a1', label='tot_a0a1')
        a\b  b0  b1  b2
          a0   0   1   2
          a1   3   4   5
          a2   6   7   8
    tot_a0a1   3   5   7
```

What we got
```python
      a\b  b0  b1  b2
       a0   0   1   2
       a1   3   4   5
       a2   6   7   8
    a0,a1   3   5   7
```

gdementen:
AFAICT, it works fine when using an axis.

alixdamman:
That's why I edited the title of the issue.

gdementen:
sorry, we just edited it simultaneously...

alixdamman:
Close this issue. Users should pass named groups. See discussion in PR #387

PR #387, opened 2017-08-30
Issues 89 : update docstring of method with_total

alixdamman:


larray/core/array.py
```diff
@@ -2942,12 +2942,15 @@ def with_total(self, *args, **kwargs):
...
-        kwargs
-        op : aggregate function
-            Defaults to `sum()`.
-        label : scalar value
-            label to use for the total. Defaults to "total".
+        \*args : int or str or Axis or Group or any combination of those, optional
+            Axes or groups along which to compute the aggregates. Defaults to aggregate over the whole array.
+        \**kwargs
+            * op : aggregate function
+
```
gdementen: useless blank line

larray/core/array.py
```diff
@@ -2942,12 +2942,15 @@ def with_total(self, *args, **kwargs):
...
-        label : scalar value
-            label to use for the total. Defaults to "total".
+        \*args : int or str or Axis or Group or any combination of those, optional
+            Axes or groups along which to compute the aggregates. Defaults to aggregate over the whole array.
+        \**kwargs
+            * op : aggregate function
+                
+                Defaults to `sum()`.
+            * label : scalar value
+
```
gdementen: useless blank line

larray/core/array.py
```diff
@@ -2942,12 +2942,15 @@ def with_total(self, *args, **kwargs):
...
-            Defaults to `sum()`.
-        label : scalar value
-            label to use for the total. Defaults to "total".
+        \*args : int or str or Axis or Group or any combination of those, optional
+            Axes or groups along which to compute the aggregates. Defaults to aggregate over the whole array.
+        \**kwargs
+            * op : aggregate function
+                
+                Defaults to `sum()`.
+            * label : scalar value
```
gdementen: , optional

larray/core/array.py
```diff
@@ -3008,6 +3023,7 @@ def with_total(self, *args, **kwargs):
                 lgkey = axis
                 axis = lgkey[0].axis
                 value = res._aggregate(npop[op], (lgkey,))
+                value.set_axes(value.axes[axis], Axis([label], axis.name), inplace=True)
```
gdementen: use value.set_labels instead?
besides, this will break when you use several groups on the same axis, but Travis already told you that :)

alixdamman:
I wonder if we shouldn't limit label to axes and ask users to provide named groups if they want to use groups?

gdementen:
That was the initial intention :). It should be documented though.

alixdamman:
OK. So, can I remove the last 2 commits and close the issue #385 ?

gdementen:
sure

alixdamman:
updated

larray/core/array.py
```diff
@@ -2942,12 +2942,14 @@ def with_total(self, *args, **kwargs):
...
-        label : scalar value
-            label to use for the total. Defaults to "total".
+        *args : int or str or Axis or Group or any combination of those, optional
+            Axes or groups along which to compute the aggregates. Passed groups should be named. 
+            Defaults to aggregate over the whole array.
+        **kwargs:
+            * op : aggregate function, optional 
+                Defaults to `sum()`.
+            * label : scalar value, optional 
+                label to use for the total. Applies only on aggregated axes, not groups. Defaults to "total".
```
gdementen: please do not use this syntax (`**kwargs` : then listing the kwargs). It does not render nicely in sphinx.

alixdamman: Do I remove `**kwargs` or unident kwargs below?

gdementen:
unindent and document it by itself, because "extra" kwargs are one way to
disambiguate ambiguous labels. eg: arr.with_total(a='a0,a4')

alixdamman:
updated

larray/core/array.py
```diff
@@ -2942,12 +2942,15 @@ def with_total(self, *args, **kwargs):
 
         Parameters
         ----------
-        args
-        kwargs
-        op : aggregate function
+        *args : int or str or Axis or Group or any combination of those, optional
+            Axes or groups along which to compute the aggregates. Passed groups should be named. 
+            Defaults to aggregate over the whole array.
+        **kwargs : Axis or Group, optional
```
gdementen: move below

larray/core/array.py
```diff
@@ -2942,12 +2942,15 @@ def with_total(self, *args, **kwargs):
...
+            Axes or groups given as keywords. 
+        op : aggregate function, optional 
             Defaults to `sum()`.
-        label : scalar value
-            label to use for the total. Defaults to "total".
+        label : scalar value, optional 
+            label to use for the total. Applies only on aggregated axes, not groups. Defaults to "total".
```
gdementen: Applies only to

alixdamman:
updated

larray/core/array.py
```diff
@@ -2942,12 +2942,15 @@ def with_total(self, *args, **kwargs):
...
+        *args : int or str or Axis or Group or any combination of those, optional
+            Axes or groups along which to compute the aggregates. Passed groups should be named. 
+            Defaults to aggregate over the whole array.
+        op : aggregate function, optional 
             Defaults to `sum()`.
-        label : scalar value
-            label to use for the total. Defaults to "total".
+        label : scalar value, optional 
+            label to use for the total. Applies only to aggregated axes, not groups. Defaults to "total".
+        **kwargs : Axis or Group, optional
```
gdementen: Sorry, to be picky, but this is wrong again. kwargs can also be scalars or combinations of groups/scalars and cannot be Axis.

larray/core/array.py
```diff
@@ -2942,12 +2942,15 @@ def with_total(self, *args, **kwargs):
...
+            Axes or groups given as keywords.
```
gdementen: idem

larray/core/array.py
alixdamman: I am often confused about arguments of methods like getitem and _aggregate. If one day we only support Python 3, it would be great to use type hints in method signatures (you already proposed that but it's tedious to have something working on both Python 2 and 3).

gdementen: I don't see how type hints would help your confusion. That's another syntax for the same thing. The reason it could be interesting is that tools exist to check the args values actually used when called the function respect the type mentioned in the type hints, whereas no such AFAIK for numpy docstrings (but we could implement one ;-))

alixdamman: kwargs can be scalars and groups but also string or not?

gdementen: You raise an interesting point. Whenever I speak about scalars I always use the numpy definition of scalar, which includes strings. I admit this is not obvious... Unsure what we should do about this. Avoid the term or???

alixdamman:
updated

alixdamman:
Thanks for the review

gdementen:
Thanks for your patience :)

PR #388, opened 2017-08-30
added changelog file for versions 0.25.2 and 0.26

alixdamman:


gdementen:
why not add them with the first feature targeted to those versions?

alixdamman:
Indeed

PR #392, opened 2017-08-31
Fixes syntax .. and spaces (issue 389)

alixdamman:


larray/tests/test_array.py
```diff
@@ -254,7 +254,7 @@ def test_getitem(self):
         assert_array_equal(la[..., lipro159], raw[..., [0, 4, 8]])
 
         # string 'int..int'
-        assert_array_equal(la['10..15'], la['10,11,12,13,14,15'])
+        assert_array_equal(la['8, 10..13, 15'], la['8,10,11,12,13,15'])
```
gdementen: I would rather have a test for both (with and without space), we never know how we will break things in the future :)

gdementen:
I would prefer if such simple features would be squashed in the future...

alixdamman:
I make separate commits when I update the changelog. But OK, I will stop to do that.

PR #395, opened 2017-08-31
issue 268 : bad error message from LArray.setitem 

alixdamman:
I may misunderstood what you had in mind. In that case, I can withdraw this PR.

### DIFF
--- a/doc/source/changes/version_0_25_1.rst.inc
+++ b/doc/source/changes/version_0_25_1.rst.inc
@@ -7,7 +7,7 @@ Miscellaneous improvements
 * updated documentation of method `with_total` (closes :issue:`89`)
 
 * trying to set values of a subset by passing an array with incompatible axes
-  displays a better error message in case of incompatible axes (closes :issue:`268`)
+  displays a better error message (closes :issue:`268`)
 
 Fixes
 -----
--- a/doc/source/changes/version_0_25_1.rst.inc
+++ b/doc/source/changes/version_0_25_1.rst.inc
@@ -6,8 +6,8 @@ Miscellaneous improvements
 
 * updated documentation of method `with_total` (closes :issue:`89`)
 
-* trying to set values of a subset by passing an LArray object displays
-  a better error message in case of incompatible axes (closes :issue:`268`)
+* trying to set values of a subset by passing an array with incompatible axes
+  displays a better error message in case of incompatible axes (closes :issue:`268`)
 
 Fixes
 -----
--- a/doc/source/changes/version_0_25_1.rst.inc
+++ b/doc/source/changes/version_0_25_1.rst.inc
@@ -6,6 +6,9 @@ Miscellaneous improvements
 
 * updated documentation of method `with_total` (closes :issue:`89`)
 
+* trying to set values of a subset by passing an LArray object displays
+  a better error message in case of incompatible axes (closes :issue:`268`)
+
 Fixes
 -----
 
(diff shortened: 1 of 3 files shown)


larray/core/array.py
```diff
@@ -2219,11 +2219,22 @@ def __setitem__(self, key, value, collapse_slices=True):
...
+            try:
+                data[cross_key] = value
+            except ValueError as err:
+                new_message = "Incompatible axes. Values from input array with axes\n" \
+                              "{!r}\ncannot be set to subset with axes\n{!r}".format(value.axes, self[key].axes)
+                if PY2:
+                    err.message = new_message
+                    raise
+                else:
+                    raise ValueError(new_message) from err
```
gdementen: The problem is only present if there are more axes on the value than the target, so I would have written something like:

```python
extra_axes = value.axes - axes
if extra_axes:
    text = 'axes are' if len(extra_axes) > 1 else 'axis is'
    raise ValueError("Value {} {} not present in target subset {}. A value can only have the same axes or fewer axes than the subset being targeted".format(extra_axes, text, axes))
```
which would yield, for example:

> Value {c} axis is not present in target subset {a, b}. A value can only have the same axes or fewer axes than the subset being targeted

doc/source/changes/version_0_25_1.rst.inc
```diff
@@ -6,8 +6,8 @@ Miscellaneous improvements
 
 * updated documentation of method `with_total` (closes :issue:`89`)
 
-* trying to set values of a subset by passing an LArray object displays
-  a better error message in case of incompatible axes (closes :issue:`268`)
+* trying to set values of a subset by passing an array with incompatible axes
+  displays a better error message in case of incompatible axes (closes :issue:`268`)
```
gdementen: raaaah...
s/ in case of incompatible axes//

larray/tests/test_array.py
```diff
@@ -855,18 +855,6 @@ def test_setitem_larray(self):
         # same raw as previous test
         assert_array_equal(la, raw)
 
-        # c) value has an extra length-1 axis
```
gdementen: I wonder if keeping this behaviour would be worth it. Adapating the test would be trivial (exclude length 1 axes from extra_axes). A setitem with extra length-1 axes is non-ambiguous after all. I don't see a use case for it though, so it's your call.

larray/tests/test_array.py
```diff
@@ -918,6 +906,17 @@ def test_setitem_larray(self):
         la[:] = 0
         assert_array_equal(la, np.zeros_like(raw))
 
+        # 6) incompatible axes
+        la = self.small.copy()
+        la2 = la.copy()
+
+        with pytest.raises(ValueError):
+           la['P01'] = la2
```
gdementen: check for the error message, otherwise we have no idea whether or not the correct error was raised (and we could go back to the initial situation of not understandable message without noticing).

gdementen:
squashing those commits would be nice

alixdamman:
For the error message, I use the method `__str__` of AxisCollection to not display a huge message (only names of axes are displayed). Do you agree with that?

alixdamman:
Yes, squashing was in my mind.

gdementen:
Error message only containing axes names: sure, that was the idea.

Strangely, here, this is what I get by default when putting an AxisCollection in a formatted string (without any format specifier):

```python
>>> arr = ndtest((2, 3))
>>> arr.axes
AxisCollection([
    Axis(['a0', 'a1'], 'a'),
    Axis(['b0', 'b1', 'b2'], 'b')
])
>>> '{}'.format(arr.axes)
'{a, b}'
```

alixdamman:
To be sure, I will let the '{!s}' as it. OK to squash and merge?

gdementen:
yes

PR #396, opened 2017-08-31
Misc : temporary patch for issue 393 (na, sort_columns, sort_rows)

alixdamman:


doc/environment.yml
```diff
@@ -8,8 +8,6 @@ dependencies:
   - pandas
   - matplotlib
   - pytables
-  - jupyter
-  - jupyter_client
   - sphinx
   - numpydoc
   - nbsphinx
```
gdementen: the commit message tells about nbsphinx not being required either???

doc/source/changes/version_0_25_1.rst.inc
```diff
@@ -6,6 +6,9 @@ Miscellaneous improvements
 
 * updated documentation of method `with_total` (closes :issue:`89`)
 
+* `read_excel` displays an error message when arguments `na`, `sort_rows` or `sort_columns` are used
+  with `xlwings` engine
```
gdementen: with the default `xlwings` engine?

larray/io/array.py
```diff
@@ -324,6 +324,12 @@ def read_excel(filepath, sheetname=0, nb_index=None, index_col=None, na=np.nan,
         if kwargs:
             raise TypeError("'{}' is an invalid keyword argument for this function when using the xlwings backend"
                             .format(list(kwargs.keys())[0]))
+        if not np.isnan(na):
+            raise NotImplementedError("na argument is not currently supported with xlwings engine")
```
gdementen: "with the default engine (xlwings)" (or something similar)?

doc/environment.yml
alixdamman: nbsphinx is an extension of sphinx (see conf.py). I should change the commit message.

larray/io/array.py
alixdamman: "with the (default) xlwings engine"?

doc/environment.yml
alixdamman: what about
"updated doc/conf.py (remove nbsphinx extension) and doc/environment.yml (removed jupyter and jupyter_client packages): jupyter scripts no longer used in documentation" for the commit message?

larray/io/array.py
gdementen: Sure.

doc/environment.yml
gdementen: Whatever. If it's only the commit message wrong, no need to change it.

larray/core/array.py
```diff
@@ -1380,7 +1380,8 @@ def reindex(self, axes_to_reindex=None, new_axis=None, fill_value=np.nan, inplac
         new_axis : int, str, list/tuple of str or Axis, optional
             List of new labels or new axis if `axes_to_replace` contains a single axis reference.
         fill_value : scalar or LArray, optional
-            Value set to data corresponding to added labels. Defaults to NaN.
+            Value used to fill cells corresponding to label combinations which were not present before reindexing.
+            Defaults to NumPy NaN.
```
gdementen: I wouldn't mention numpy, as the result would be the same if we used float('nan'). "Defaults to NaN" seems enough.

larray/io/array.py
```diff
@@ -128,7 +128,7 @@ def df_aslarray(df, sort_rows=False, sort_columns=False, raw=False, parse_header
     return LArray(data, axes)
 
 
-def read_csv(filepath_or_buffer, nb_index=None, index_col=None, sep=',', headersep=None, na=np.nan,
+def read_csv(filepath_or_buffer, nb_index=None, index_col=None, sep=',', headersep=None, fill_value=np.nan,
```
gdementen: We need to still support na= and output a warning message.

alixdamman:
Have you seen the update of this PR?

alixdamman:
Once OK, let me clean my commits before to merge.

larray/io/array.py
```diff
@@ -196,6 +196,7 @@ def read_csv(filepath_or_buffer, nb_index=None, index_col=None, sep=',', headers
          FO  2  3
     """
     if not np.isnan(na):
+        fill_value = na
         warnings.warn("Argument na will be replace by fill_value in a future version of larray.",
```
gdementen: I would rather use the "deprecated" or "renamed" wording, like we did for deprecated functions. Here it is only an argument but the concept is the same. Something like: 

    read_csv `na` argument has been renamed to `fill_value`. Please use that instead.

alixdamman:
Didn't change anything, just rebased. OK to merge?

PR #398, opened 2017-08-31
fixed an obscure bug Geert stumbled upon

gdementen:


larray/core/axis.py
```diff
@@ -720,7 +720,7 @@ def translate(self, key, bool_passthrough=True):
             # path is only used if the key was given in "non normalized form"
             assert np.isscalar(key), "%s (%s) is not scalar" % (key, type(key))
             # key is scalar (integer, float, string, ...)
-            if np.dtype(type(key)).kind == self.labels.dtype.kind:
```
alixdamman: I guess this line was written before _is_key_type_compatible() ;)

gdementen: indeed!

PR #401, opened 2017-09-01
prepare for release on monday

gdementen:


alixdamman:
I wonder if we shouldn't upgrade the required version for numpy and pandas in setup.py?

gdementen:
Why? Is there any feature we use that is not present in the old versions?

PR #414, opened 2017-09-06
fix #376 : make combine_axes and boolean selection use both _ as sep

### LINKED ISSUE #376: make combine_axes and boolean selection use same sep

alixdamman:


gdementen:
I need to test this with Bernard's model. I don't remember why he wanted to use `_` instead of `,` but there was a good reason. I guess I should ask him, maybe he remembers something.

gdementen:
could you rebase this?

alixdamman:
rebased. I'm not convinced that choosing `,` is a good idea.

gdementen:
Why? Honestly, I am unsure too, but I don't know why and I don't have a better idea. :)

gdementen:
@alixdamman ping. You never answered why using `,` is not a good idea. What do you suggest we do? I only know that to keep using two different/incompatible separators is not an option.

alixdamman:
`,` is used to separate labels in the "string" syntax (when you extract subset). 
Using `,` in labels would be dangerous and messy, don't you think?

gdementen:
Right, this would prevent one to select multiple such combined labels using the string syntax:
`a['a0,b0,a1,b1']` 
would not work. The question is: what separator do we use?
a = ndtest((2, 2))

* `_` like we did in combine_axes?

```python
>>> a[a % 3 == 0]
a_b	a0_b0	a1_b1
	0	3
>>> a.combine_axes()
a_b	a0_b0	a0_b1	a1_b0	a1_b1
	0	1	2	3
>>> a['a0_b0,a1_b1']
a_b	a0_b0	a1_b1
	0	3
```

* `&`

```python
>>> a[a % 3 == 0]
a&b	a0&b0	a1&b1
	0	3
>>> a.combine_axes()
a&b	a0&b0	a0&b1	a1&b0	a1&b1
	0	1	2	3
>>> a['a0&b0,a1&b1']
a&b	a0&b0	a1&b1
	0	3
```
* other separator which might make sense: `/` `~` `+`
* `;` would have mostly the same problem than `,` if we do #434 
* separators I would **not** use: parentheses, brackets, `@`, `=`, `|` (because it will be confusing when NDGroups are implemented).

alixdamman:
`_`?

gdementen:
ok. Could you tackle that?

alixdamman:
updated

doc/source/changes/version_0_26.rst.inc
```diff
@@ -318,6 +318,18 @@ Miscellaneous improvements
     sex  M  F
          0  1
 
+* getting a subset using a boolean selection returns an array with labels combined with underscore by defaults
```
gdementen: move this at the top in a "backward incompatible changes" section?

PR #420, opened 2017-09-11
rewrap all code to 120 chars (closes #94)

### LINKED ISSUE #94: rewrap all code to 120 chars
To avoid having 100 small commits doing it, let's do it all at once.

gdementen:


alixdamman:
This PR makes me think that it should also be done for larray-editor...

gdementen:
I know... In fact that was one of the comments I intended to add. Here are the others:
* this was a marathon. A lot more time-consuming and tiring (I would even say painful) than I expected. I will not start on larray-editor just yet. I would appreciate if you did that part.
* I probably missed a ton, but we will find them in time. I hereby authorise you to fix them in unrelated PR when you see one :)
* I left many occurrences where readability would be impacted. Most of them in list comprehensions:
I find:

```python
toto = [complex_expr
        for a, b in blabla]
```
much more readable than:
```python
toto = [complex_expr for a, b in blabla]
```
e.g.
```python
        key = tuple(axis.translate(axis_key, bool_passthrough=not bool_stuff)
                    for axis, axis_key in zip(self.axes, key))
```

This is especially true when complex_expr is a ternary expression, as for example in:

```python
        key = [key[axis] if axis in key else slice(None)
               for axis in self.axes]
```
compare with:
```python
        key = [key[axis] if axis in key else slice(None) for axis in self.axes]
```

but for "simple enough" exprs I use the one line form.  The criterion is that I can tell a first glance where the "value" ends and where the "for" part starts. This might be due to syntax highlighting. If the first keyword I find while reading the expr is not the "for" keyword, I must actually "read" the full expr to find it, which annoys me.

gdementen:
One more thing: I tried hard to restrain myself to line changes. I did fix a few things along the way (typos, etc.) but I restrained myself to uncontroversial changes.

larray/example.py
```diff
@@ -8,14 +8,15 @@
...
 }
 
+
 def load_example_data(name):
     """Load arrays used in the tutorial so that all examples in it can be reproduced.
 
     Parameters
     ----------
-    example_data : str
+    name : str
```
alixdamman: good catch

gdementen:
done

PR #422, opened 2017-09-12
fix #328 : Method to_excel/to_hdf accept Group with as sheet_name/key

### LINKED ISSUE #328: Methods to_excel/to_hdf should accept a Group with a single element as sheet_name/key
Users may want to do something like this:

```python
for label in a.row:
    a[label].to_excel('file.xlsx', label)
```

alixdamman:


doc/source/changes/version_0_26.rst.inc
```diff
@@ -44,6 +44,24 @@ Miscellaneous improvements
 * replaced `na` argument of `read_csv`, `read_excel`, `read_hdf` and `read_sas` functions by `fill_value`
   (closes :issue:`394`)
 
+* allowed to pass a label or a named group as `sheet_name` argument of the method `to_excel` and
+  `key` argument of the method `to_hdf` (closes :issue:`328`):
+
+    >>> arr = ndtest((4, 4, 4))
+    >>> # iterate over labels of a given axis
+    >>> for label in arr.a:
+    ...    arr[label].to_excel('my_file.xlsx', label)
```
gdementen: use 4 spaces indent instead of 3 spaces?

larray/core/array.py
```diff
@@ -5681,6 +5690,15 @@ def to_excel(self, filepath=None, sheet_name=None, position='A1', overwrite_file
         >>> # add to existing sheet starting at position A15
         >>> a.to_excel('test.xlsx', 'Sheet1', 'A15')  # doctest: +SKIP
         """
+        if isinstance(sheet_name, Group):
```
gdementen: I would rather use `sheet_name = str(_to_tick(sheet_name))` instead of all this code

larray/core/array.py
```diff
@@ -5637,6 +5637,15 @@ def to_hdf(self, filepath, key, *args, **kwargs):
         >>> a = ndtest((2, 3))
         >>> a.to_hdf('test.h5', 'a')  # doctest: +SKIP
         """
+        if isinstance(key, Group):
```
gdementen: I would rather use `sheet_name = str(_to_tick(sheet_name))` instead of all this code

doc/source/changes/version_0_26.rst.inc
```diff
@@ -44,6 +44,24 @@ Miscellaneous improvements
...
+    ...    arr[label].to_hdf('my_file.h5', label)
+    >>> # create and use a named group
+    >>> even = arr.a['a0::2'] >> 'even'
+    >>> arr[even].to_excel('my_file.xlsx', even)
+    >>> arr[even].to_hdf('my_file.h5', even)
+    >>> # passing a group with no name is not allowed
+    >>> odd = arr.a['a1::2']
+    >>> arr[odd].to_excel('my_file.xlsx', odd)
+    ...
+    ValueError: Passed group for `sheet_name` argument must be of length 1 or have a name
```
gdementen: disallowing sheet_names with commas seems arbitrary and counter-productive. See other comments below.

alixdamman:
I agree that `str(_to_tick(...))` is a better option and I will use it in the code. 
But I need to add a test to avoid error like:
`'You typed an invalid name for a sheet or chart. Make sure that:
The name that you type does not exceed 31 characters.
The name does not contain any of the following characters:  :  \\  /  ?  *  [  or  ]\n
You did not leave the name blank.'`

alixdamman:
This will happen often with unnamed groups.

alixdamman:
what about: 
```python
sheet_name = '_'.join(str(l) for l in _to_ticks(sheet_name))
```
?

gdementen:
Could you provide a concrete example where that problem appears?
using to_ticks seems weird as to_tick is meant to produce a single scalar, which is what we need, while to_ticks is meant to produce several.

alixdamman:
New implementation with `_to_tick` crashes with unnamed group defined via slice. 
See changes in test_array.py for an example.

gdementen:
Ok, now I see the example. The problem is mostly with slices but can happen in other cases. I think making it always work would be practical for users. Maybe translate all invalid chars to `_`?

doc/source/changes/version_0_26.rst.inc
```diff
@@ -44,6 +44,27 @@ Miscellaneous improvements
...
   (closes :issue:`394`)
 
+* allowed to pass a label or a named group as `sheet_name` argument of the method `to_excel` and
+  `key` argument of the method `to_hdf` (closes :issue:`328`):
+
+    >>> arr = ndtest((4, 4, 4))
+
+    >>> # iterate over labels of a given axis
+    >>> for label in arr.a:
+    ...     arr[label].to_excel('my_file.xlsx', label)
```
gdementen: I have a problem with this example, because this is not the way users should dump an array to multiple sheets, because it opens & close the file several times, which is a bit costly. They should use open_excel in that case:
```python
with open_excel('my_file.xlsx') as wb:
    for label in arr.a:
        wb[label] = arr[label]
```
so I guess Workbook[Group] = xxx should be implemented too before the release. This can be done in this PR or another, as you prefer, or we might want to implement #423 directly.

doc/source/changes/version_0_26.rst.inc
```diff
@@ -44,6 +44,27 @@ Miscellaneous improvements
 * replaced `na` argument of `read_csv`, `read_excel`, `read_hdf` and `read_sas` functions by `fill_value`
   (closes :issue:`394`)
 
+* allowed to pass a label or a named group as `sheet_name` argument of the method `to_excel` and
```
gdementen: s/a label or a named group/an axis group/ ??

doc/source/changes/version_0_26.rst.inc
```diff
@@ -44,6 +44,27 @@ Miscellaneous improvements
...
+
+    >>> arr = ndtest((4, 4, 4))
+
+    >>> # iterate over labels of a given axis
+    >>> for label in arr.a:
+    ...     arr[label].to_excel('my_file.xlsx', label)
+    ...     arr[label].to_hdf('my_file.h5', label)
+
+    >>> # create and use a group
+    >>> even = arr.a['a0::2'] >> 'even'
```
gdementen: I don't think using ::2 in an example is worth it. It is likely to confuse most users because they do not know that syntax.

```python
mygroup = arr.a['a0,a2,a3'] >> 'mygroup'
```

doc/source/changes/version_0_26.rst.inc
alixdamman: The meaning of "axis group" is not obvious to me, so I guess it is not also for the users. But I can drop the 'named' since it also work for unnamed group.

doc/source/changes/version_0_26.rst.inc
alixdamman: Indeed, wrong example. That makes me think we should explain this (not using `to_excel` in a loop) in the tutorial.

doc/source/changes/version_0_26.rst.inc
alixdamman: They don't know that syntax? I guess you're right. I'll change the example as you suggest.

doc/source/changes/version_0_26.rst.inc
alixdamman: OK to implement Workbook[Group] = xxx in this PR. Seems legit.

alixdamman:
I would like to solve issue #423 in another PR if you don't mind.

doc/source/changes/version_0_26.rst.inc
```diff
@@ -44,6 +44,30 @@ Miscellaneous improvements
 * replaced `na` argument of `read_csv`, `read_excel`, `read_hdf` and `read_sas` functions by `fill_value`
   (closes :issue:`394`)
 
+* allowed to pass a label or group as `sheet_name` argument of the method `to_excel` and as
+  `key` argument of the method `to_hdf` (closes :issue:`328`):
```
gdementen: also mention open_excel here

larray/core/array.py
```diff
@@ -5637,6 +5638,8 @@ def to_hdf(self, filepath, key, *args, **kwargs):
...
+            key = re.sub('[\\/?*\[\]:]', '_', str(_to_tick(key)))
```
gdementen: you reused the same invalid chars than for Excel, but I doubt the set of invalid chars is exactly the same.

larray/io/excel.py
```diff
@@ -171,12 +173,16 @@ def _ipython_key_completions_(self):
             return list(self.sheet_names())
 
         def __getitem__(self, key):
+            if isinstance(key, Group):
```
gdementen: I think it would be better to create an helper function "sanitize_excel_sheet_name" or something like that than copy-pasting the same code all over the place.

larray/io/excel.py
```diff
@@ -171,12 +173,16 @@ def _ipython_key_completions_(self):
...
+                key = re.sub('[\\/?*\[\]:]', '_', str(_to_tick(key)))
```
gdementen: for such a "fixed" regex, I prefer to compile the regexp and store it as a module variable

gdementen:
about issue #423. Sure.

larray/core/array.py
alixdamman: I just wanted to be consistent with `to_excel`.

larray/io/excel.py
alixdamman: "translate_excel_sheet_name"?

larray/io/excel.py
alixdamman: OK.

alixdamman:
Don't worry. I will squash and merge at the end.

larray/io/excel.py
```diff
@@ -1,12 +1,14 @@
 import os
 import atexit
+import re
```
gdementen: not needed anymore

larray/core/array.py
```diff
@@ -5637,6 +5639,7 @@ def to_hdf(self, filepath, key, *args, **kwargs):
         >>> a = ndtest((2, 3))
         >>> a.to_hdf('test.h5', 'a')  # doctest: +SKIP
         """
+        key = str(_to_tick(key))
```
gdementen: there aren't ANY invalid characters in HDF array names?

alixdamman: Well, using / seems to be a bad idea (not crashing but leading to weird bug) : https://stackoverflow.com/questions/30070834/in-names-in-hdf5-files-confusion

alixdamman: I could be a good idea to forbid `/` and maybe `\`

gdementen: ok

alixdamman:
OK for the changelog?

Issue #382, opened 2017-08-30
combine several axes lists at once in combine_axes 


gdementen:
See also #366 

Possible syntaxes:
```python
>>> arr.combine_axes({('a', 'b'): 'ab', 'c,d': 'cd'})

>>> # or the opposite
>>> arr.combine_axes({'ab': ('a', 'b'), 'cd': 'c,d'})
>>> # which would allow the nicer syntax:
>>> arr.combine_axes(ab=('a', 'b'), cd='c,d')
```
The problem with this "opposite" syntax is that it is not consistent with (it is the inverse of) how arr.rename, arr.set_axes and arr.set_labels work, which use from, to instead of to, from:

```python
>>> arr = ndtest((2, 3))
>>> arr
a\b  b0  b1  b2
 a0   0   1   2
 a1   3   4   5
>>> arr.rename('a', 'c')
>>> arr.rename({'a': 'c'})

>>> arr.set_axes(a='c=0..6')
>>> arr.set_axes({'a': 'c=0..6'})

>>> arr.set_labels({'a': '0,1'})
```

alixdamman:
The syntax
```python 
>>> arr.combine_axes(ab=('a', 'b'), cd='c,d')
```
do not allow to use special characters in axis names. 

To be consistent with `rename`, `set_axes`, ... I would choose the first syntax (from, to).

gdementen:
Indeed, but
```python
>>> arr.combine_axes({'ab': ('a', 'b'), 'cd': 'c,d'})
```
would allow special characters, and in rename and set_axes we support both a dict and **kwargs too (for that reason).

The question was in fact whether or not to change the syntax for rename, set_axes and set_labels because with the keyword syntax, the "inverted" syntax reads better/is a lot more logical than the current syntax: you **assign** something to the new name, in the same way you assign a variable.

```python
>>> arr.rename(c='a')
```
but I just realized this only works for rename and not for set_axes and set_labels, so the question is closed.

**Conclusion**: let us just implement the first syntax:

```python
>>> arr.combine_axes({('a', 'b'): 'ab'}
```

alixdamman:
to simplify the implement of this, can I add a `name` argument in `AxisCollection.combine_axes` (assuming we use a dict in `LArray.combine_axes`)?

alixdamman:
You want both `LArray.combine_axes` and `AxisCollection.combine_axes` to combine several axes lists at once or just `LArray.combine_axes`?

[...]

Issue #366, opened 2017-08-24
split_axis -> split_axes


gdementen:
Allow split_axis to split several axes at once (easy).

The "problem" is that this calls for combining multiple axes lists at once too and it might be tricky to provide a nice syntax for that (the implementation is easy though).

alixdamman:
Not sure what to do with argument `names`. Nested list?

gdementen:
Ugh... or a dict? I would need to see an example to decide.

alixdamman:
like this?
```python
    >>> arr = ndtest((2, 2, 2, 2))
    >>> arr_comb = arr.combine_axes(['a','b']).combine_axes(['c','d'])
    >>> arr_comb
    a_b\c_d  c0_d0  c0_d1  c1_d0  c1_d1
      a0_b0      0      1      2      3
      a0_b1      4      5      6      7
      a1_b0      8      9     10     11
      a1_b1     12     13     14     15
    >>> arr = arr_comb.split_axes(['a_b', 'c_d'], {'a_b': ['A', 'B'], 'c_d': ['C', 'D']})
    >>> arr
      A   B  C\D  d0  d1
    a0  b0   c0   0   1 
    a0  b0   c1   2   3
    a0  b1   c0   4   5 
    a0  b1   c1   6   7
    a1  b0   c0   8   9
    a1  b0   c1  10  11
    a1  b1   c0  12  13
    a1  b1   c1  14  15
```

alixdamman:
We should stay consistent with #382 :
```python
>>> arr.combine_axes({('a', 'b'): 'ab'})
>>> arr.split_axes({'ab': ('a', 'b')})
```

alixdamman:
Note: accept also `axes=None` as input (all axes with name containing `sep` will be split).

PR #424, opened 2017-09-15
fix #365 + #366 + #382 : split_axes() and combine_axes()

### LINKED ISSUE #365: split_axes -> axes default to all axes containing sep

alixdamman:
I start with `combine_axes`, then when we agree on this, I'll move to `split_axes` (easier).

### DIFF
--- a/doc/source/changes/version_0_26.rst.inc
+++ b/doc/source/changes/version_0_26.rst.inc
@@ -112,6 +112,33 @@ Miscellaneous improvements
     >>> group_even_year
     year[2000, 2002, 2004, 2006, 2008, 2010, 2012, 2014, 2016]
 
+* allowed to perform several axes combinations at once with the `combine_axes()` method (closes :issue:`382`):
+
+    >>> arr = ndtest((2, 2, 2, 2))
+    >>> arr
+     a   b  c\d  d0  d1
+    a0  b0   c0   0   1
+    a0  b0   c1   2   3
+    a0  b1   c0   4   5
+    a0  b1   c1   6   7
+    a1  b0   c0   8   9
+    a1  b0   c1  10  11
+    a1  b1   c0  12  13
+    a1  b1   c1  14  15
+    >>> arr.combine_axes([('a', 'c'), ('b', 'd')])
+    a_c\b_d  b0_d0  b0_d1  b1_d0  b1_d1
+      a0_c0      0      1      4      5
+      a0_c1      2      3      6      7
+      a1_c0      8      9     12     13
+      a1_c1     10     11     14     15
+    >>> # set output axes names by passing a dictionary
+    >>> arr.combine_axes({('a', 'c'): 'ac', ('b', 'd'): 'bd'})
+    ac\bd  b0_d0  b0_d1  b1_d0  b1_d1
+    a0_c0      0      1      4      5
+    a0_c1      2      3      6      7
+    a1_c0      8      9     12     13
+    a1_c1     10     11     14     15
+
 * allowed to use keyword arguments in `set_labels` (closes :issue:`383`):
 
     >>> a = ndrange('nat=BE,FO;sex=M,F')
--- a/doc/source/changes/version_0_26.rst.inc
+++ b/doc/source/changes/version_0_26.rst.inc
@@ -112,6 +112,36 @@ Miscellaneous improvements
     >>> group_even_year
     year[2000, 2002, 2004, 2006, 2008, 2010, 2012, 2014, 2016]
 
+* renamed `split_axis` as `split_axes` which now allows to split several axes at once (closes :issue:`366`):
+
+    >>> combined = ndrange('a_b = a0_b0..a1_b1; c_d = c0_d0..c1_d1')
+    >>> combined
+    a_b\c_d  c0_d0  c0_d1  c1_d0  c1_d1
+      a0_b0      0      1      2      3
+      a0_b1      4      5      6      7
+      a1_b0      8      9     10     11
+      a1_b1     12     13     14     15
+    >>> combined.split_axes(['a_b', 'c_d'])
+     a   b  c\d  d0  d1
+    a0  b0   c0   0   1
+    a0  b0   c1   2   3
+    a0  b1   c0   4   5
+    a0  b1   c1   6   7
+    a1  b0   c0   8   9
+    a1  b0   c1  10  11
+    a1  b1   c0  12  13
+    a1  b1   c1  14  15
+    >>> combined.split_axes({'a_b': ('A', 'B'), 'c_d': ('C', 'D')})
+     A   B  C\D  d0  d1
+    a0  b0   c0   0   1
+    a0  b0   c1   2   3
+    a0  b1   c0   4   5
+    a0  b1   c1   6   7
+    a1  b0   c0   8   9
+    a1  b0   c1  10  11
+    a1  b1   c0  12  13
+    a1  b1   c1  14  15
+
 * allowed to perform several axes combinations at once with the `combine_axes()` method (closes :issue:`382`):
 
     >>> arr = ndtest((2, 2, 2, 2))
--- a/doc/source/changes/version_0_26.rst.inc
+++ b/doc/source/changes/version_0_26.rst.inc
@@ -142,6 +142,27 @@ Miscellaneous improvements
     a1  b1   c0  12  13
     a1  b1   c1  14  15
 
+* argument `axes` of `split_axes` has become optional: defaults to all axes whose name contains the specified delimiter
+  (closes :issue:`365`):
+
+    >>> combined = ndrange('a_b = a0_b0..a1_b1; c_d = c0_d0..c1_d1')
+    >>> combined
+    a_b\c_d  c0_d0  c0_d1  c1_d0  c1_d1
+      a0_b0      0      1      2      3
+      a0_b1      4      5      6      7
+      a1_b0      8      9     10     11
+      a1_b1     12     13     14     15
+    >>> combined.split_axes()
+     a   b  c\d  d0  d1
+    a0  b0   c0   0   1
+    a0  b0   c1   2   3
+    a0  b1   c0   4   5
+    a0  b1   c1   6   7
+    a1  b0   c0   8   9
+    a1  b0   c1  10  11
+    a1  b1   c0  12  13
+    a1  b1   c1  14  15
+
 * allowed to perform several axes combinations at once with the `combine_axes()` method (closes :issue:`382`):
 
     >>> arr = ndtest((2, 2, 2, 2))
--- a/larray/core/array.py
+++ b/larray/core/array.py
@@ -6323,8 +6323,10 @@ class LArray(ABCLArray):
 
         Parameters
         ----------
-        axes : tuple, list or AxisCollection of axes, optional
-            axes to combine. Defaults to all axes.
+        axes : tuple, list, AxisCollection of axes or list of combination of those or dict, optional
+            axes to combine. Tuple, list or AxisCollection will combine several axes into one. To chain several axes 
+            combinations, pass a list of tuple/list/AxisCollection of axes. To set the name(s) of resulting axis(es), 
+            use a {(axes, to, combine): 'new_axis_name'} dictionary. Defaults to all axes.
         sep : str, optional
             delimiter to use for combining. Defaults to '_'.
         wildcard : bool, optional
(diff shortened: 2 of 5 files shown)


larray/core/array.py
```diff
@@ -6301,15 +6301,20 @@ def compact(self):
...
-    def combine_axes(self, axes=None, sep='_', wildcard=False):
+    def combine_axes(self, axes=None, sep='_', name=None, wildcard=False):
         """Combine several axes into one.
 
         Parameters
         ----------
-        axes : tuple, list or AxisCollection of axes, optional
-            axes to combine. Defaults to all axes.
+        axes : tuple, list, AxisCollection of axes or dict, optional
+            axes to combine. Tuple, list or AxisCollection will combine several axes in one. To combine in several
```
gdementen: s/combine in/combine into/g

larray/core/array.py
```diff
@@ -6332,34 +6337,66 @@ def combine_axes(self, axes=None, sep='_', wildcard=False):
...
+           ac  b\\d  d0  d1
+        a0_c0   b0   0   1
+        a0_c0   b1   4   5
+        a0_c1   b0   2   3
+        a0_c1   b1   6   7
+        a1_c0   b0   8   9
+        a1_c0   b1  12  13
+        a1_c1   b0  10  11
+        a1_c1   b1  14  15
+        >>> arr.combine_axes({('a', 'c'): 'ac', ('b', 'd'): None})
```
gdementen: I wonder if we shouldn't support this instead (or in addition to the dict syntax):
```python
>>> arr.combine_axes([('a', 'c'), ('b', 'd')])
```
because the None is kinda annoying to type.

larray/core/axis.py
```diff
@@ -2094,39 +2099,81 @@ def combine_axes(self, axes=None, sep='_', wildcard=False, front_if_spread=False
...
+        >>> axes
+        AxisCollection([
+            Axis(['a0', 'a1', 'a2'], 'a'),
+            Axis(['b0', 'b1', 'b2'], 'b')
+        ])
+        >>> axes.combine_axes()
+        AxisCollection([
+            Axis(['a0_b0', 'a0_b1', 'a0_b2', 'a1_b0', 'a1_b1', 'a1_b2', 'a2_b0', 'a2_b1', 'a2_b2'], 'a_b')
+        ])
+        >>> axes.combine_axes(sep='/')
```
gdementen: I think there are too many examples in here. I would move some of them to unit tests. But wait until we are sure of the syntax before changing them.

larray/core/array.py
alixdamman: I agree to implement this:
```python
>>> arr.combine_axes('a,c;b,d')
>>> arr.combine_axes('a,c;b,d', names='ac;bd')

>>> arr.combine_axes([('a', 'c'), ('b', 'd')])
>>> arr.combine_axes([('a', 'c'), ('b', 'd')], names=['ac', 'bd'])
```
If users want to use `names`, they should provide all names.

larray/core/axis.py
alixdamman: normally, very few users will use `AxisCollection.combine_axes`. So I don't think it is really a problem to have many examples.

larray/core/array.py
gdementen: I wonder if requiring the dict syntax when the user wants to provide names is not an option:

We would have:
```python
>>> arr.combine_axes(('a', 'c'))
>>> arr.combine_axes({('a', 'c'): 'ac'})
>>> arr.combine_axes([('a', 'c'), ('b', 'd')])
>>> arr.combine_axes({('a', 'c'): 'ac', ('b', 'd'): 'bd'})
```

alixdamman: dict are not ordered for Python <= 3.5. This is why I have a loop in `LArray.combine_axes` to compute "transposed_axes". I would prefer to avoid dict.

alixdamman: OK

larray/core/axis.py
```diff
@@ -2082,8 +2082,10 @@ def combine_axes(self, axes=None, sep='_', wildcard=False, front_if_spread=False
 
         Parameters
         ----------
-        axes : tuple, list or AxisCollection of axes, optional
-            axes to combine. Defaults to all axes.
+        axes : tuple, list, AxisCollection of axes or list of combination of those or dict, optional
+            axes to combine. Tuple, list or AxisCollection will combine several axes into one. To chain several axes 
+            combination, pass a list of tuple/list/AxisCollection of axes. To set name(s) of resulting axis(es), use a
```
gdementen: combinations

larray/core/array.py
```diff
@@ -6306,8 +6306,10 @@ def combine_axes(self, axes=None, sep='_', wildcard=False):
 
         Parameters
         ----------
-        axes : tuple, list or AxisCollection of axes, optional
-            axes to combine. Defaults to all axes.
+        axes : tuple, list, AxisCollection of axes or list of combination of those or dict, optional
+            axes to combine. Tuple, list or AxisCollection will combine several axes into one. To chain several axes 
+            combinations, pass a list of tuple/list/AxisCollection of axes. To set name(s) of resulting axis(es), use a
```
gdementen: s/set name(s)/set the name(s)/

larray/core/array.py
```diff
@@ -6306,8 +6306,10 @@ def combine_axes(self, axes=None, sep='_', wildcard=False):
 
         Parameters
         ----------
-        axes : tuple, list or AxisCollection of axes, optional
-            axes to combine. Defaults to all axes.
+        axes : tuple, list, AxisCollection of axes or list of combination of those or dict, optional
+            axes to combine. Tuple, list or AxisCollection will combine several axes into one. To chain several axes 
+            combinations, pass a list of tuple/list/AxisCollection of axes. To set name(s) of resulting axis(es), use a 
+            dictionary with entries {(axes to combine): 'new_axis_name'}. Defaults to all axes.
```
gdementen: maybe: s/use a dictionary with entries {(axes to combine): 'new_axis_name'}./use a {(axes, to, combine): 'new_axis_name'} dictionary./

larray/core/axis.py
```diff
@@ -2097,39 +2099,99 @@ def combine_axes(self, axes=None, sep='_', wildcard=False, front_if_spread=False
         -------
         AxisCollection
             New AxisCollection with combined axes.
+            
+        Examples
+        --------
+        >>> axes = AxisCollection('a=a0..a2;b=b0..b2')
```
gdementen: use a0,a1 instead to make examples shorter and prevent people from thinking axes must have the same length to be combined?

larray/core/axis.py
```diff
@@ -2097,39 +2099,99 @@ def combine_axes(self, axes=None, sep='_', wildcard=False, front_if_spread=False
...
+        ])
+        >>> axes.combine_axes()
+        AxisCollection([
+            Axis(['a0_b0', 'a0_b1', 'a0_b2', 'a1_b0', 'a1_b1', 'a1_b2', 'a2_b0', 'a2_b1', 'a2_b2'], 'a_b')
+        ])
+        >>> axes.combine_axes(sep='/')
+        AxisCollection([
+            Axis(['a0/b0', 'a0/b1', 'a0/b2', 'a1/b0', 'a1/b1', 'a1/b2', 'a2/b0', 'a2/b1', 'a2/b2'], 'a/b')
+        ])
+        >>> axes += AxisCollection('c=c0..c2;d=d0..d2')
```
gdementen: use d0,d1 instead?

doc/source/changes/version_0_26.rst.inc
```diff
@@ -94,6 +94,28 @@ Miscellaneous improvements
     >>> group_even_year
     year[2000, 2002, 2004, 2006, 2008, 2010, 2012, 2014, 2016]
 
+* allowed to perform several axes combinations at once with `combine_axes()` method (closes :issue:`382`):
+
+    >>> axes = AxisCollection('a=a0,a1;b=b0..b2;c=c0..c2;d=d0,d1')
```
gdementen: wouldn't an example using LArray.combine_axes be more telling than an example using AxisCollection? 
to avoid having a too large example I'd use: ndrange('a=a0,a1;b=b0,b1;c=c0,c1;d=d0..d2')

doc/source/changes/version_0_26.rst.inc
```diff
@@ -94,6 +94,28 @@ Miscellaneous improvements
...
+    >>> axes
+    AxisCollection([
+        Axis(['a0', 'a1'], 'a'),
+        Axis(['b0', 'b1', 'b2'], 'b'),
+        Axis(['b0', 'b1', 'b2'], 'b'),
```
gdementen: copy-paste error

doc/source/changes/version_0_26.rst.inc
gdementen: regardless of the example, you should at least mention that it is implemented both for LArrray and AxisCollection

gdementen:
rebase & squash and you are good to go

doc/source/changes/version_0_26.rst.inc
```diff
@@ -112,6 +112,8 @@ Miscellaneous improvements
     >>> group_even_year
     year[2000, 2002, 2004, 2006, 2008, 2010, 2012, 2014, 2016]
 
+* argument `axes` of `split_axes` has become optional: defaults to the last axis (closes :issue:`365`).
```
gdementen: Sorry. I forgot to update the issue... But I wonder if splitting all axes containing sep wouldn't be a better default.

alixdamman:
Still need to change the second commit.

alixdamman:
update: swaped commit 2 and 3 + change commit in which `axes` becomes optional.

doc/source/changes/version_0_26.rst.inc
```diff
@@ -112,6 +112,87 @@ Miscellaneous improvements
...
+* renamed `split_axis` as `split_axes` which now allows to split several axes at once (closes :issue:`366`):
+
+    >>> combined = ndrange('a_b = a0_b0..a1_b1; c_d = c0_d0..c1_d1')
+    >>> combined
+    a_b\c_d  c0_d0  c0_d1  c1_d0  c1_d1
+      a0_b0      0      1      2      3
+      a0_b1      4      5      6      7
+      a1_b0      8      9     10     11
+      a1_b1     12     13     14     15
+    >>> # equivalent to combined.split_axes() which split all axes
```
gdementen: I would either remove the comment here (in the release notes) because you describe that just below and/or combine the release notes of both improvements.

larray/core/array.py
```diff
@@ -6335,43 +6337,88 @@ def combine_axes(self, axes=None, sep='_', wildcard=False):
...
+            transposed_axes = transposed_axes - axes_to_combine
+            transposed_axes = transposed_axes[:min_axis_index] + axes_to_combine + transposed_axes[min_axis_index:]
         transposed = self.transpose(transposed_axes)
 
         new_axes = transposed.axes.combine_axes(axes, sep=sep, wildcard=wildcard)
         return transposed.reshape(new_axes)
 
-    def split_axis(self, axis, sep='_', names=None, regex=None):
-        """Split one axis and returns a new array
+    def split_axes(self, axes=None, sep='_', names=None, regex=None):
```
gdementen: We need to keep the old name for both methods (and have it display a warning...)

doc/source/changes/version_0_26.rst.inc
```diff
@@ -112,6 +112,87 @@ Miscellaneous improvements
...
+    a0  b0   c0   0   1
+    a0  b0   c1   2   3
+    a0  b1   c0   4   5
+    a0  b1   c1   6   7
+    a1  b0   c0   8   9
+    a1  b0   c1  10  11
+    a1  b1   c0  12  13
+    a1  b1   c1  14  15
+
+* allowed to perform several axes combinations at once with `LArray.combine_axes()` and `AxisCollection.combine_axes()`
```
gdementen: Maybe not specifying the class at all, like you did for split_axes above is enough after all. ie "with `combine_axes()`". At least we should be consistent. Either specify that it is for both LArray and AxisCollection in both split_axes & combine_axes or not specify it for both.

larray/core/array.py
```diff
@@ -6407,12 +6454,44 @@ def split_axis(self, axis, sep='_', names=None, regex=None):
...
+        
+        >>> combined = ndrange('a_b = a0_b0..a1_b1; c_d = c0_d0..c1_d1')
+        >>> combined 
+        a_b\\c_d  c0_d0  c0_d1  c1_d0  c1_d1
+          a0_b0      0      1      2      3
+          a0_b1      4      5      6      7
+          a1_b0      8      9     10     11
+          a1_b1     12     13     14     15
+        >>> # equivalent to combined.split_axes() which split all axes 
+        >>> # containing the delimiter defined by the argument `sep`
```
gdementen: all axes whose name contains the `sep` delimiter

larray/core/axis.py
```diff
@@ -2094,48 +2096,110 @@ def combine_axes(self, axes=None, sep='_', wildcard=False, front_if_spread=False
...
+        """Split axes and returns a new collection
 
         Parameters
         ----------
-        axis : int, str or Axis
-            axis to split. All its labels *must* contain the given delimiter string.
+        axes : int, str, Axis or any combination of those, optional 
+            axes to split. All labels *must* contain the given delimiter string. To split several axes at once, pass 
+            a list or tuple of axes to split. To set the names of resulting axes, use a {'axis_to_split': (new, axes)} 
+            dictionary. Defaults to all axes containin the delimeter `sep` or satisfying `regex`.
```
gdementen: The code does not seem to handle the "satisfying `regex`" part.

larray/core/array.py
```diff
@@ -6335,43 +6337,88 @@ def combine_axes(self, axes=None, sep='_', wildcard=False):
...
+        """Split axes and returns a new array
 
         Parameters
         ----------
-        axis : int, str or Axis
-            axis to split. All its labels *must* contain the given delimiter string.
+        axes : int, str, Axis or any combination of those
+            axes to split. All labels *must* contain the given delimiter string. To split several axes at once, pass 
+            a list or tuple of axes to split. To set the names of resulting axes, use a {'axis_to_split': (new, axes)} 
+            dictionary. Defaults to all axes containin the delimeter `sep` or satisfying `regex`.
```
gdementen: The code does not seem to handle the "satisfying `regex`" part.

doc/source/changes/version_0_26.rst.inc
```diff
@@ -112,6 +112,87 @@ Miscellaneous improvements
...
+    a0  b0   c1   2   3
+    a0  b1   c0   4   5
+    a0  b1   c1   6   7
+    a1  b0   c0   8   9
+    a1  b0   c1  10  11
+    a1  b1   c0  12  13
+    a1  b1   c1  14  15
+
+* argument `axes` of `split_axes` has become optional: defaults to all axes with names containing the delimiter
+  defined by the argument `sep` or satisfying the pattern defined by the argument `regex` (closes :issue:`365`):
```
gdementen: * s/defaults to all axes with names containing the delimiter defined by the argument `sep`/defaults to all axes with a name containing the specified delimiter/ or even /defaults to all axes whose name contains the specified delimiter/

* The code does not seem to handle the "satisfying `regex`" part.

larray/core/array.py
alixdamman: `regex` is only for labels. You're right.

PR #425, opened 2017-09-15
fix #408 : set_labels with an Axis as the labels argument

### LINKED ISSUE #408: set_labels with an Axis as the labels argument
This almost work but produces an axis with PGroup labels, which is supposed to never happen.

The correct way to implement this is probably to detect group labels in Axis.__init__ and "evaluate" them.

alixdamman:


gdementen:
also needs a entry in the changelog

PR #428, opened 2017-09-15
fix #341 : allowed to pass an axis/group/int to set the name of group.

### LINKED ISSUE #341: allow setting the name of a Group using another Group
In the code below `>> str(code)` should be `>> code`
```python
>>> arr = ndrange('axis=a,a0..a3,b,b0..b3,c,c0..c3')
>>> arr
axis  a  a0  a1  a2  a3  b  b0  b1  b2  b3   c  c0  c1  c2  c3
      0   1   2   3   4  5   6   7   8   9  10  11  12  13  14
>>> groups = tuple(arr.axis.startswith(code) >> str(code) for code in arr.axis.matches('^.$'))
>>> groups
(axis['a', 'a0', 'a1', 'a2', 'a3'] >> 'a',
 axis['b', 'b0', 'b1', 'b2', 'b3'] >> 'b',
 axis['c', 'c0', 'c1', 'c2', 'c3'] >> 'c')
>>> arr.sum(groups)
axis   a   b   c
      10  35  60
```

alixdamman:


larray/core/group.py
```diff
@@ -727,6 +727,18 @@ def __init__(self, key, name=None, axis=None):
         # the same name. See test_array.py:test_...
         self.axis = axis
 
+    @property
+    def name(self):
+        return self._name
+
+    @name.setter
+    def name(self, name):
```
gdementen: FWIW, this setter approach makes me uneasy. as I always considered the groups to be immutable objects. You could leave that in and we will see how it goes

larray/core/group.py
```diff
@@ -727,6 +727,18 @@ def __init__(self, key, name=None, axis=None):
...
+        if isinstance(name, (Group, ABCAxis)):
+            name = name.name
```
gdementen: This is wrong. It will break for unnamed groups: it will use None as the name, which is not what we want (we want to use the label). Might want to use _to_tick or something like that (needs to be checked though).

larray/tests/test_group.py
```diff
@@ -46,6 +47,17 @@ def test_init(self):
...
+        # passing an axis as name
+        group = Group('1:5', self.age, self.age)
+        assert group.name == self.age.name
+        group = self.age['1:5'] >> self.age
+        assert group.name == self.age.name
+        # passing an group as name
+        group2 = Group('1:5', group, self.age)
+        assert group2.name == group.name
+        group2 = self.age['1:5'] >> group
+        assert group2.name == group.name
```
gdementen: You should also add a check for using a group without name.

doc/source/changes/version_0_26.rst.inc
```diff
@@ -70,6 +70,21 @@ Miscellaneous improvements
     ['_name_with_special___char_']
     >>> # special characters \ or / in labels or groups are replaced by an _ when exporting to HDF file
 
+* allowed setting the name of a Group using another Group or Axis (closes :issue:`341`):
+
+    >>> arr = ndrange('axis=a,a0..a3,b,b0..b3,c,c0..c3')
+    >>> arr
+    axis  a  a0  a1  a2  a3  b  b0  b1  b2  b3   c  c0  c1  c2  c3
+          0   1   2   3   4  5   6   7   8   9  10  11  12  13  14
+    >>> groups = tuple(arr.axis.startswith(code) >> str(code) for code in arr.axis.matches('^.$'))
```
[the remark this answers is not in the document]
alixdamman: Bad copy/paste. I just couldn't find another case in which that feature could be useful.

larray/core/group.py
alixdamman: That was unclear in the issue explanation.

larray/core/group.py
alixdamman: I can move that to init (where `name` is set).

alixdamman:
Passing None to `_to_tick` (idem for `_to_ticks` I guess) return a string 'None'. That's what we want?

gdementen:
_to_tick should never receive None in this context, since what I suggested (or at least wanted to) was to use name = _to_tick(name) if isinstance(name, Group) else name. But what you wrote is fine.

alixdamman:
My question concerned the behavior of `_to_tick` in general, not specifically in this context.

gdementen:
Ah. This comes from np.isscalar(None) returning False (which I find strange). I have yet to see a case where a user wants to use None as a tick, but I suppose returning None instead of "None" would make more sense indeed.

PR #430, opened 2017-09-18
removed useless if statement in _make_axis()  

alixdamman:
This line was also responsible for a nasty bug:
```python
>>> ndrange('age=10')
        10  age
             0
```

### DIFF
--- a/doc/source/changes/version_0_26.rst.inc
+++ b/doc/source/changes/version_0_26.rst.inc
@@ -117,4 +117,4 @@ Miscellaneous improvements
 Fixes
 -----
 
-* fixed array creation with axis(es) containing only one label (axis name and label were inversed).
+* fixed array creation with axis(es) given as string containing only one label (axis name and label were inversed).
--- a/doc/source/changes/version_0_26.rst.inc
+++ b/doc/source/changes/version_0_26.rst.inc
@@ -117,4 +117,4 @@ Miscellaneous improvements
 Fixes
 -----
 
-* fixed something (closes :issue:`1`).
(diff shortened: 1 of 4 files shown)


gdementen:
Nice catch! The bug fix needs a changelog entry though.

doc/source/changes/version_0_26.rst.inc
```diff
@@ -99,4 +99,4 @@ Miscellaneous improvements
 Fixes
 -----
 
-* fixed something (closes :issue:`1`).
+* fixed array creation with axis(es) containing only one label (axis name and label were inversed).
```
gdementen: s/axis(es) containing/axis(es) given as string containing/

alixdamman:
I have modified Axis.labels and _to_ticks because `AxisCollection('age=10')` returned an axis 'age' with a unique STRING label '10' instead of an INT.

gdementen:
I wonder if this latest change does not break creating wildcard axes... Could you test that? eg. `assert len(Axis(10)) == 10`. I don't know if there is already a test for this.

alixdamman:
creation of wilcard axes is already tested in doctest of `Axis.__init__` but added a new test in test_axis.py

alixdamman:
The question is `Axis('axis=10')` should be considered as a wildcard axis or an axis with one label?

gdementen:
I think that ideally, that should be a wildcard axis and `Axis('axis=10,')` should be used to create an axis with a single label but this is such an edge case that I don't really care :)

PR #440, opened 2017-09-21
fix #383 : allowed keyword arguments in set_labels

### LINKED ISSUE #383: allow keyword arguments in set_labels
The goal is mostly to be consistent with rename and set_axes

```python
>>> arr = ndtest((2, 3))
>>> arr
a\b  b0  b1  b2
 a0   0   1   2
 a1   3   4   5
>>> arr.set_labels({'a': 'x,y'})
a\b  b0  b1  b2
  x   0   1   2
  y   3   4   5
>>> arr.set_labels('a', 'x,y')
a\b  b0  b1  b2
  x   0   1   2
  y   3   4   5
>>> arr.set_labels(a='x,y')
TypeError: set_labels() got an unexpected keyword argument 'a'
```

alixdamman:


larray/core/array.py
```diff
@@ -6112,8 +6121,11 @@ def set_labels(self, axis, labels=None, inplace=False):
         """
         if isinstance(axis, dict):
             changes = axis
-        else:
+        elif isinstance(axis, (basestring, Axis, int)):
             changes = {axis: labels}
+        else:
```
gdementen: We probably want to add a something like: `assert axis is None` here. Otherwise errors could pass silently. eg
a user might write: arr.set_labels(['a', 'b'], ['a0,a1', 'b0,b1']) and I think this would silently pass (but do nothing).

PR #442, opened 2017-09-21
fix #439 : allowed to pass a Group to read_excel/read_hdf 

### LINKED ISSUE #439: read_excel(hdf) should accept a Group as sheet(key) argument

alixdamman:


larray/tests/test_array.py
```diff
@@ -2425,10 +2425,17 @@ def test_hdf_roundtrip(self):
...
+        for label in a3.a:
+            subset = read_hdf(fpath, label)
+            assert_array_equal(subset, a3[label])
+
+        # load Session
         from  larray.core.session import Session
         s = Session(fpath)
         assert s.names == sorted(['a0', 'a1', 'a2', 'a3', 'c0,c2', 'c0::2', 'even', ':name?with*special__[characters]'])
 
+
```
gdementen: kill line

doc/source/changes/version_0_26.rst.inc
```diff
@@ -72,6 +72,17 @@ Miscellaneous improvements
     ['_name_with_special___char_']
     >>> # special characters \ or / in labels or groups are replaced by an _ when exporting to HDF file
 
+* allowed to pass a Group to `read_excel`/`read_hdf` as `sheetname`/`key` argument (closes :issue:`439`).
```
gdementen: I would group this with the "output" side of things (the previous change), so that the example can be moved at the end of that change example and simplified (because you no longer need to create the files). Eg somthing like (**untested!!!**):
```python
>>> new_from_excel = zeros((arr.a, arr.b, arr.c))
>>> for label in new_from_excel.a:
...     new_from_excel[label] = read_excel('my_file.xlsx', label)
>>> # But, to avoid loading the file in Excel repeatedly (which is very inefficient), 
>>> # this particular example should rather be written like this:
>>> new_from_excel = zeros((arr.a, arr.b, arr.c))
>>> with open_excel('my_file.xlsx') as wb:
...     for label in new_from_excel.a:
...         new_from_excel[label] = wb[label].load()

>>> # For HDF
>>> new_from_hdf = zeros((arr.a, arr.b, arr.c))
>>> for label in new_from_hdf.a:
...     new_from_hdf[label] = read_hdf('my_file.h5', label)
```
I also wonder if storing the axes in a, b and c variables wouldn't improve the readabilty of the example:
```python
>>> a, b, c = arr.axes
```

alixdamman:
updated (I've tested the example).

PR #445, opened 2017-09-22
containing

gdementen:


doc/source/changes/version_0_26.rst.inc
```diff
@@ -1,19 +1,46 @@
...
+
+* renamed special variable `x` to `X` to let users define an `x` variable in their code without breaking all
+  subsequent code using that special variable (closes :issue:`167`).
+
+* renamed Axis.startswith, endswith and matches to startingwith, endingwith and matching to avoid a possible confusion
+  with str.startswith and endswith which return booleans (closes :issue:`432`).
+
+* renamed `na` argument of `read_csv`, `read_excel`, `read_hdf` and `read_sas` functions to `fill_value` to avoid
+  confusion as to what the argument does and to be consistent with reindex (closes :issue:`394`).
+
```
alixdamman: `reindex`?

doc/source/changes/version_0_26.rst.inc
```diff
@@ -1,19 +1,46 @@
...
   return only arrays defined globally.
 
+* implemented Axis.containing to create a Group with all labels of an axis containing some substring (closes
+  :issue:`402`).
+
+    >>> people = Axis(['Bruce Wayne', 'Bruce Willis', 'Arthur Dent'], 'people')
+    >>> people.containing('Will')
+    people['Bruce Willis']
+
+* implemented Group.containing, startingwith, endingwith and matching to create a group with all labels of a group
```
alixdamman: to create a group with all labels of s/a group/an axis/

larray/core/axis.py
```diff
@@ -441,6 +441,10 @@ def equals(self, other):
                (len(self) == len(other) if self.iswildcard else np.array_equal(self.labels, other.labels))
 
     def matches(self, pattern):
+        warnings.warn("Axis.matching was renamed to Axis.matches, please use that instead", FutureWarning, stacklevel=2)
```
alixdamman: you inverted Axis.matching and Axis.matches in your warning message.

larray/core/group.py
```diff
@@ -12,7 +12,7 @@
 from larray.util.oset import *
 from larray.util.misc import basestring, PY2, unique, find_closing_chr, _parse_bound, _seq_summary
 
-__all__ = ['LGroup', 'LSet', 'PGroup', 'union']
+__all__ = ['Group', 'LGroup', 'LSet', 'PGroup', 'union']
```
alixdamman: why? Group is supposed to be only a base class right?

gdementen: because I needed to do something like if isinstance(v, Group) in some user code (I don't remember where nor why) and it felt cleaner to just export it than use if isinstance(v, (LGroup, PGroup))

alixdamman: OK. Now I understand.

gdementen:
updated

PR #449, opened 2017-09-26
fix #441 : file deletion if crash occurs during file overwriting

### LINKED ISSUE #441: when saving a file under an existing name fails, the file is deleted anyway
This is extremely annoying. I had a large (valid) file, I introduced a few bad values in it, tried to save it. It failed (without error, see larray-project/larray-editor#75) and I then closed the editor to realize my file had disappeared, hence loosing my work!

This happens at least for saving to Excel, but might happen for other backends too.

This is going to be a tad annoying to fix:

```
if destpath exists:
    if overwrite:
        save file as temppath
        delete destpath
        rename temppath to destpath
    else:
       raise "{despath} already exists"
else:
    save file as destpath

```
Or, a tad simpler (but at tad less efficient):    

```
save file as temppath
if destpath exists:
    if overwrite:
        delete destpath
    else:
        raise "{despath} already exists"
rename temppath to destpath
```

alixdamman:


larray/io/session.py
```diff
@@ -109,19 +109,28 @@ def read_arrays(self, keys, *args, **kwargs):
...
+        overwrite_file : bool
+            Whether or not to overwrite the existing file (or just modify the specified sheet).
         kwargs :
             * display: whether or not to display when the dump of each array is started/done.
         """
         display = kwargs.pop('display', False)
-        self._open_for_write()
+        # for XLWingsHandler, the job is already done in the Workbook class
+        if not isinstance(self, (PandasCSVHandler, XLWingsHandler)) and overwrite_file and os.path.isfile(self.fname):
+            fname = self.fname
```
gdementen: * s/fname/final_fname/ or something similar. Having both fname and self.fname have different meaning/values is too confusing.
* when you have to check for specific backends in generic code, this smells like a bad design. I would rather move this functionality in one of the overridden methods (_open_for_write or probably a new one to avoid duplicating code). This method might even be on the generic/abstract FileHandler class, but this way each backend can override it or not.

larray/io/excel.py
```diff
@@ -64,6 +64,7 @@ def __init__(self, filepath=None, overwrite_file=False, visible=None, silent=Non
 
             xw_wkb = None
             self.delayed_filepath = None
+            self.filepath = None
```
gdementen: I wonder if we shouldn't have only one filepath. Ie reuse delayed_filepath in that case?

doc/source/changes/version_0_26.rst.inc
```diff
@@ -253,3 +253,6 @@ Fixes
 -----
 
 * fixed array creation with axis(es) given as string containing only one label (axis name and label were inversed).
+
+* fixed original file deletion when trying to overwrite a file via `Session.save` or `open_excel` but crashing
```
gdementen: * s/deletion/being deleted/ ?
* s/but crashing/failed/

larray/io/session.py
```diff
@@ -40,7 +40,7 @@ def __init__(self, fname):
     def _open_for_read(self):
         raise NotImplementedError()
 
-    def _open_for_write(self):
+    def _open_for_write(self, overwrite_file):
```
gdementen: for other methods, the "state" is passed to __init__ and stored in an attribute instead of being passed as argument to each method. This is a debatable choice, but using both styles seems weird when that does not seem required (overwrite_file is already known when `__init__` is called, right?)

larray/io/session.py
```diff
@@ -173,6 +173,7 @@ def _dump(self, key, value, *args, **kwargs):
 
     def close(self):
         self.handle.close()
+        self._delete_temporary_file()
```
gdementen: Unsure doing this in close is a good thing. These are two separate operations. Wouldn't it be better to call the method from dump_arrays instead?

larray/io/session.py
```diff
@@ -67,6 +69,15 @@ def close(self):
         """
         raise NotImplementedError()
 
+    def _create_temporay_file(self):
```
gdementen: rename to _get_temporary_fname and call it from `__init__` instead? Also I think it would be better to modify self.fname directly and store the original (unmodified) fname in some other attribue (eg self.final_fname), so that each backend do not have to worry about tmp file in their open_for_write method.

larray/io/session.py
```diff
@@ -67,6 +69,15 @@ def close(self):
...
+        if self.overwrite_file and os.path.isfile(self.fname):
+            self.tmp_file = '{}~{}'.format(*os.path.splitext(self.fname))
+
+    def _delete_temporary_file(self):
```
gdementen: rename to _rename_temporary_file (it is a rename with overwrite -- so more a rename than a delete) and call it from dump_arrays instead?

larray/io/session.py
```diff
@@ -67,6 +68,18 @@ def close(self):
...
         raise NotImplementedError()
 
+    def _get_original_file_name(self):
+        original_file_name = None
+        if self.overwrite_file and os.path.isfile(self.fname):
+            original_file_name = self.fname
+            self.fname = '{}~{}'.format(*os.path.splitext(self.fname))
+        return original_file_name
+
+    def _update_original_file(self, original_file_name):
```
gdementen: My original comment about passing state both via arguments and via attributes still stand. I know passing the state via attributes is debatable/is a matter of style in this case, but mixing both styles without a good reason (at least that I can see) feels odd.

larray/tests/test_array.py
```diff
@@ -2877,11 +2877,15 @@ def test_to_excel_xlwings(self):
 
     @pytest.mark.skipif(xw is None, reason="xlwings is not available")
     def test_open_excel(self):
+        fpath = abspath('new_excel_file.xlsx')
+        if os.path.exists(fpath):
+            os.remove(fpath)
```
gdementen: new_excel_file was never meant to exist in the "create new file" test, so you should also add this at the end of the test, and/or use different fpath for the new test and/or rename new_excel_file to this_file_does_not_exist.xlsx in the first test, to make it obvious what we mean.

PR #452, opened 2017-09-28
fix #443 : strip border lines/columns with invisible/blank cells (Excel)

### LINKED ISSUE #443: reading an Excel sheet with invisible/blank cells should strip them by default
see attached file
[bad.xlsx](https://github.com/liam2/larray/files/1320676/bad.xlsx)

```python
>>> read_excel('bad.xlsx', 'bad')
 a\b  2003  2004  None  None  None
  a0   0.0   1.0  None  None  None
  a1   2.0   3.0  None  None  None
None   nan   nan  None  None  None
```
* It should only strip bottom-right blanks
* Unsure the nans should be stripped. We need to investigate why those cells are different from the None cells.
* Unsure having the option to avoid stripping is worth it (let's say no until somebody requests this).

alixdamman:
Not sure new tests are enough. I may add a "bad2" sheet where the top-left corner of the array to read is not located at cell 'A1'.


larray/io/excel.py
```diff
@@ -321,10 +321,9 @@ def __setitem__(self, key, value):
         @property
         def shape(self):
             # include top-left empty rows/columns
-            # XXX: is there an exposed xlwings API for this? expand maybe?
-            used = self.xw_sheet.api.UsedRange
-            return (used.Row + used.Rows.Count - 1,
-                    used.Column + used.Columns.Count - 1)
+            # XXX: requires xlwings 0.9+
+            used = Range(self, 'A1').xw_range.current_region
```
gdementen: I wanted to avoid this because it will not take all data if you have a blank column/row in your sheet. That's why I used "usedRange" (which is equivalent to Ctrl-End in Excel). See the "bad2" sheet in the new file I uploaded on the issue. BTW: "bad2" is loaded "correctly" (as I think we should) by the xlrd engine. On the other hand having the behavior your just implemented is often what user want, so if we can provide an option for that, that would be nice. I would still include everything (except trailing blanks) by default, because it is easier/more intuitive to take a subset when you received too much rather than just miss some data (possibly without noticing).

larray/io/excel.py
```diff
@@ -321,10 +321,20 @@ def __setitem__(self, key, value):
         @property
         def shape(self):
             # include top-left empty rows/columns
-            # XXX: is there an exposed xlwings API for this? expand maybe?
+            # see https://www.excelcampus.com/vba/find-last-row-column-cell/#endcode
+            from xlwings.constants import Direction as xldir
             used = self.xw_sheet.api.UsedRange
-            return (used.Row + used.Rows.Count - 1,
-                    used.Column + used.Columns.Count - 1)
+            last_row = max(used.Cells(used.Rows.Count + 1, col).End(xldir.xlUp).Row
```
gdementen: My fear is that this would be too slow for large sheets. Please test this. If it turns out reasonably fast, this is ok. Otherwise, I would use something like the code below (untested!!!) which should be potentially much faster:

```python
last_cell = None
row = used.Row + used.Rows.Count - 1
# probably needs to add used.Column (so that it works if sheet has blank row/columns before data
col_count = used.Columns.Count
while last_cell is None:
    max_row_cell = used.Cells(row, col_count + 1).End(xldir.xlToLeft)
    max_row_col = max_row_cell.Column - 1
    # we need to check the first column is not filled
    if max_col > 0 or max_row_cell.Value != '':
        last_cell = (row, max_col)
    row -= 1
```

gdementen: How about something like this (untested!)?
```python
col_after_last = used.Column + used.Columns.Count + 1
for last_used_row in range(used.Row + used.Rows.Count, 0, -1):
    max_row_cell = used.Cells(last_used_row, col_after_last).End(xldir.xlToLeft)
    if max_row_cell.Column > 1 or max_row_cell.Value != '':
        break
row_after_last = last_used_row + 1
for last_used_col in range(used.Column + used.Columns.Count, 0, -1):
    max_col_cell = used.Cells(row_after_last, last_used_col).End(xldir.xlUp)
    if max_col_cell.Row > 1 or max_col_cell.Value != '':
        break
return (last_used_row, last_used_col)
```

larray/io/excel.py
```diff
@@ -324,11 +324,17 @@ def shape(self):
...
-            last_row = max(used.Cells(used.Rows.Count + 1, col).End(xldir.xlUp).Row
-                           for col in range(1, used.Columns.Count+1))
-            last_col = max(used.Cells(row, used.Columns.Count + 1).End(xldir.xlToLeft).Column
-                           for row in range(1, used.Rows.Count+1))
-            return (last_row, last_col)
+            last_row = used.Row + used.Rows.Count
+            last_col = used.Column + used.Columns.Count
+            for last_used_row in range(last_row, 0, -1):
+                left_cell = used.Cells(last_used_row, last_col + 1).End(xldir.xlToLeft)
+                if left_cell.Column > 1 or (left_cell.Value is not None and left_cell.Value != ''):
```
gdementen: Are both cases possible? (ie can a cell be None or == '')?

gdementen:
nevermind the changelog :) Just squashing :)

larray/io/excel.py
alixdamman: testing if Value is None seems to be enough.

larray/io/excel.py
```diff
@@ -321,10 +321,26 @@ def __setitem__(self, key, value):
         @property
         def shape(self):
             # include top-left empty rows/columns
-            # XXX: is there an exposed xlwings API for this? expand maybe?
+            # see https://www.excelcampus.com/vba/find-last-row-column-cell/#endcode
```
gdementen: unsure the link is still relevant?

larray/io/excel.py
```diff
@@ -321,10 +321,26 @@ def __setitem__(self, key, value):
...
+                left_cell = used.Cells(last_used_row, last_col + 1).End(xldir.xlToLeft)
+                if left_cell.Column > 1 or left_cell.Value is not None:
+                    break
+            for last_used_col in range(last_col, 0, -1):
+                up_cell = used.Cells(last_row + 1, last_used_col).End(xldir.xlUp)
+                if up_cell.Row > 1 or up_cell.Value is not None:
+                    break
+            return (last_used_row, last_used_col)
+
+        # @property
```
gdementen: remove this commented code

doc/source/changes/version_0_26.rst.inc
```diff
@@ -255,4 +255,7 @@ Fixes
 * fixed array creation with axis(es) given as string containing only one label (axis name and label were inversed).
 
 * fixed original file being deleted when trying to overwrite a file via `Session.save` or `open_excel` failed
-  (closes :issue:`441`)
\ No newline at end of file
+  (closes :issue:`441`)
+
+* fixed loading arrays from Excel sheets containing blank or invisible cells on right column or bottom line
```
gdementen: now that I think of it, "invisible cells" seems weird. The cells themselves are visible, their content might not. But I guess simply using "blank cells" should be clear enough, no?

larray/tests/test_array.py
```diff
@@ -2532,6 +2532,18 @@ def test_read_excel_xlwings(self):
                                                 "the xlwings backend"):
             read_excel(abspath('test.xlsx'), engine='xlwings', dtype=float)
 
+        # Excel sheet with invisible/blank cells on right/bottom border of the array to read
+        fpath = abspath('test_invisible_blank_cells.xlsx')
```
gdementen: rename to test_trailing_blank_cells?

doc/source/changes/version_0_26.rst.inc
gdementen: => "containing blank cells below or right of the array to read"?

larray/io/excel.py
gdementen: Since we do not implement the exact same logic, I would rather have the link in the issue for posterity than in the code.

larray/tests/test_array.py
alixdamman: test_blank_cells.xlsx?

PR #453, opened 2017-09-29
fix #450 : implemented deprecated decorator

### LINKED ISSUE #450: implement @deprecated decorator

alixdamman:
No need for a changelog IMO.

### DIFF
--- a/larray/core/axis.py
+++ b/larray/core/axis.py
@@ -479,7 +479,7 @@ class Axis(ABCAxis):
         rx = re.compile(pattern)
         return LGroup([v for v in self.labels if rx.match(v)], axis=self)
 
-    matches = renamed_to('matches', matching)
+    matches = renamed_to(matching, 'matches')
 
     def startingwith(self, prefix):
         """
@@ -505,7 +505,7 @@ class Axis(ABCAxis):
             prefix = prefix.eval()
         return LGroup([v for v in self.labels if v.startswith(prefix)], axis=self)
 
-    startswith = renamed_to('startswith', startingwith)
+    startswith = renamed_to(startingwith, 'startswith')
 
     def endingwith(self, suffix):
         """
@@ -531,7 +531,7 @@ class Axis(ABCAxis):
             suffix = suffix.eval()
         return LGroup([v for v in self.labels if v.endswith(suffix)], axis=self)
 
-    endswith = renamed_to('endswith', endingwith)
+    endswith = renamed_to(endingwith, 'endswith')
 
     def containing(self, substring):
         """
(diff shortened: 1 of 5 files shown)


larray/tests/test_array.py
```diff
@@ -3534,6 +3534,13 @@ def test_0darray_convert(self):
         expected_np12 = "only integer scalar arrays can be converted to a scalar index"
         assert msg in {expected_np11, expected_np12}
 
+    def test_deprecated_methods(self):
+        with pytest.warns(FutureWarning):
```
gdementen: should check the actual message and that the warning "points" to the correct file with something like:
```python
        with pytest.warns(RuntimeWarning) as caught_warnings:
            ...
        assert len(caught_warnings) == 1
        warn_msg = "..."
        assert caught_warnings[0].message.args[0] == warn_msg
        assert caught_warnings[0].filename == __file__
```

larray/util/misc.py
```diff
@@ -592,3 +593,18 @@ def index_by_id(seq, value):
         if item is value:
             return i
     raise ValueError("%s is not in list" % value)
+
+
+# deprecated decorator
+def deprecated(newfunc, alt_message=None, stacklevel=2):
```
gdementen: * s/alt_message/msg/ ?
* s/deprecated/renamed/ ? Simple renamings are a special case of deprecation.

larray/core/session.py
```diff
@@ -490,9 +490,9 @@ def to_csv(self, fname, names=None, display=False, **kwargs):
         """
         self.save(fname, names, ext_default_engine['csv'], display=display, **kwargs)
 
+    @deprecated(to_csv)
```
gdementen: I don't really like having those old method being defined with no body. I would rather write the "decorator" such that this works:
 
```python
dump_excel = deprecated(to_excel)
```

possibly renaming deprecated to something like renamed_to, old_name_for or deprecated_name_for.

OR

```python
@deprecated('dump_excel')
def to_excel(self, fname, names=None, *args, **kwargs):
    self.save(fname, names, ext_default_engine['xlsx'], overwrite, display, **kwargs)
```
possibly renaming deprecated to something like renamed_from, deprecated_name_was, ...

larray/core/axis.py
```diff
@@ -482,10 +479,7 @@ def matching(self, pattern):
         rx = re.compile(pattern)
         return LGroup([v for v in self.labels if rx.match(v)], axis=self)
 
-    def startswith(self, prefix):
-        warnings.warn("Axis.startswith was renamed to Axis.startingwith, please use that instead", FutureWarning,
-                      stacklevel=2)
-        return self.startingwith(prefix)
+    matches = renamed_to('matches', matching)
```
gdementen: Wouldn't the following read better?

```python
matches = renamed_to(matching, oldname='matches')
```

larray/util/misc.py
```diff
@@ -592,3 +593,11 @@ def index_by_id(seq, value):
         if item is value:
             return i
     raise ValueError("%s is not in list" % value)
+
+
+def renamed_to(old_name, newfunc, stacklevel=2):
+    def wrapper(*args, **kwargs):
```
gdementen: You do not use functools.wrap... I don't know whether it applies in this case, but if it does not, please remove the extra import.

PR #456, opened 2017-10-02
implemented nan_equal

gdementen:
unsure it should be mentioned in the changelog. What do you think?

### DIFF
--- a/doc/source/api.rst
+++ b/doc/source/api.rst
@@ -502,14 +502,14 @@ Miscellaneous
    aslarray
    labels_array
    larray_equal
+   larray_nan_equal
+   nan_equal
    union
    stack
    identity
    diag
    eye
    ipfp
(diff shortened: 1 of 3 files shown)


alixdamman:
A changelog would be nice. You must update api.rst also.

gdementen:
done

doc/source/api.rst
```diff
@@ -520,6 +520,9 @@ Session
    :toctree: _generated/
 
    Session
+   local_arrays
+   global_arrays
+   load_example_data
```
alixdamman: why here?

gdementen: because those functions return sessions/are a way to create a session. Do you see a better place?

alixdamman: No, LGTM.

PR #462, opened 2017-10-05
Formalizing unstructured format (solve 460 + 427)

alixdamman:
Not sure I didn't miss something. It was hard enough to not break any tests for I/O functions + df_aslarray.
Waiting we agree on implementation before to add a changelog.

### DIFF
--- a/doc/source/api.rst
+++ b/doc/source/api.rst
@@ -500,6 +500,7 @@ Miscellaneous
    :toctree: _generated/
 
    aslarray
+   from_frame
    labels_array
    larray_equal
    larray_nan_equal
--- a/doc/source/changes/version_0_26.rst.inc
+++ b/doc/source/changes/version_0_26.rst.inc
@@ -63,6 +63,23 @@ New features
     a    a0    a1    a2
        True  True  True
 
+* implemented from_frame() to convert a Pandas DataFrame to an array:
+
+    >>> df = ndtest((2, 2, 2)).to_frame()
+    >>> df
+    c      c0  c1
+    a  b
+    a0 b0   0   1
+       b1   2   3
+    a1 b0   4   5
+       b1   6   7
+    >>> from_frame(df)
+     a  b\\c  c0  c1
+    a0   b0   0   1
+    a0   b1   2   3
+    a1   b0   4   5
+    a1   b1   6   7
+
 
 Miscellaneous improvements
 --------------------------
@@ -289,6 +306,15 @@ Miscellaneous improvements
      b [2]: 'b0' 'b1'
     dtype: float64
 
+* To create a 1D array using from_string() and the default separator ' ', a tabulation character `\t`
+  (instead of - previously) must be added in front of the data line:
+
+    >>> from_string('''sex  M  F
+    ...                \t   0  1''')
+    sex  M  F
+         0  1
+
+
 Fixes
 -----
 
@@ -310,6 +336,27 @@ Fixes
           M    12    10
 
   Closes :issue:`372`.
+
+* fixed converting a 1xN Pandas DataFrame to an array using `aslarray` (closes :issue:`427`):
+
+    >>> df = pd.DataFrame([[1, 2, 3]], index=['a0'], columns=['b0', 'b1', 'b2'])
+    >>> df
+        b0  b1  b2
+    a0   1   2   3
+    >>> aslarray(df)
+    {0}\{1}  b0  b1  b2
+         a0   1   2   3
+
+    >>> # setting name to index and columns
+    >>> df.index.name = 'a'
+    >>> df.columns.name = 'b'
+    >>> df
+    b   b0  b1  b2
+    a
+    a0   1   2   3
+    >>> aslarray(df)
+    a\b  b0  b1  b2
+     a0   1   2   3
 
 * fixed original file being deleted when trying to overwrite a file via `Session.save` or `open_excel` failed
   (closes :issue:`441`)
--- a/larray/core/array.py
+++ b/larray/core/array.py
@@ -3,9 +3,9 @@ from __future__ import absolute_import, division, print_function
 
 __all__ = [
     'LArray', 'zeros', 'zeros_like', 'ones', 'ones_like', 'empty', 'empty_like', 'full', 'full_like', 'sequence',
-    'create_sequential', 'ndrange', 'labels_array', 'ndtest', 'aslarray', 'identity', 'diag', 'eye', 'larray_equal',
-    'larray_nan_equal', 'all', 'any', 'sum', 'prod', 'cumsum', 'cumprod', 'min', 'max', 'mean', 'ptp', 'var', 'std',
-    'median', 'percentile', 'stack', 'nan', 'nan_equal'
+    'create_sequential', 'ndrange', 'labels_array', 'ndtest', 'aslarray', 'identity', 'diag', 'eye',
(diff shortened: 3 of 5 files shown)


larray/io/array.py
```diff
@@ -377,7 +453,7 @@ def read_sas(filepath, nb_index=None, index_col=None, fill_value=np.nan, na=np.n
         index_col = [index_col]
 
     df = pd.read_sas(filepath, index=index_col, **kwargs)
-    return df_aslarray(df, sort_rows=sort_rows, sort_columns=sort_columns, fill_value=fill_value)
+    return df_aslarray(df, sort_rows=sort_rows, sort_columns=sort_columns, fill_value=fill_value, from_io_func=True)
```
gdementen: I wonder if read_sas should use from_io_func. I guess not (no \, no 1D).

doc/source/api.rst
```diff
@@ -500,6 +500,7 @@ Miscellaneous
    :toctree: _generated/
 
    aslarray
+   from_frame
```
gdementen: needs to add from_series too

larray/core/array.py
```diff
@@ -1048,9 +1048,9 @@ def to_frame(self, fold_last_axis_name=False, dropna=None):
 
         Parameters
         ----------
-        fold_last_axis_name : bool, optional.
-            False by default.
-        dropna : {'any', 'all', None}, optional.
+        fold_last_axis_name : bool, optional
```
gdementen: I wonder if we should not remove this option from the public API and have an internal _to_io_frame function which would do this if necessary before or after calling to_frame. In that case we could remove it from from_frame too. The question is if it can be useful to read an LArray-formatted file using pandas and then converting it to LArray via from_frame. My impression is that even there are cases where users could want to do that **but** those would include some data manipulation that would make that strange axis name useless. Thoughts?

larray/core/array.py
```diff
@@ -6549,6 +6545,173 @@ def split_axes(self, axes=None, sep='_', names=None, regex=None):
...
+        new_index = pd.MultiIndex.from_tuples(list(product(*labels)))
+    columns = sorted(df.columns) if sort_columns else list(df.columns)
+    # the prodlen test is meant to avoid the more expensive array_equal test
+    prodlen = np.prod([len(axis_labels) for axis_labels in labels])
+    if prodlen == len(df) and columns == list(df.columns) and np.array_equal(df.index.values, new_index.values):
+        return df, labels
+    return df.reindex(new_index, columns, **kwargs), labels
+
+
+def from_series(s, sort_rows=False):
```
gdementen: Do we really need to have from_series, from_frame et al in here? They belong more to the io package.

alixdamman: IMO, "I/O" means read and write from files but such definition is maybe too restrictive...

larray/core/array.py
alixdamman: by removing, you mean in both `to_frame` and `from_frame`?

gdementen:
i/o means... input and output. IMO any conversion from/to an external format (even living in memory) falls under i/o. Otherwise, you get some weird inconsitencies: to_excel('file.xlsx') is i/o under your definition but to_excel() (live excel instance) is not. read_csv('some_url') is not i/o, etc.

larray/core/array.py
gdementen: i/o means... input and output. IMO any conversion from/to an external format (even living in memory) falls under i/o. Otherwise, you get some weird inconsitencies: to_excel('file.xlsx') is i/o under your definition but to_excel() (live excel instance) is not. read_csv('some_url') is not i/o, etc.

alixdamman: make more sense indeed.

larray/core/array.py
gdementen: Yes, **IF** we go through with this. To be clear, the functionality would still need to live somewhere, but would not be exposed as arguments of public API method/functions.

doc/source/api.rst
alixdamman: The current implementation is partially. I prefer not.

larray/io/array.py
```diff
@@ -484,8 +560,8 @@ def from_string(s, nb_index=None, index_col=None, sep=' ', **kwargs):
 
     Examples
     --------
-    >>> # if one dimension array and default separator ' ', a - must be added in front of the data line
-    >>> from_string("sex  M  F\\n-  0  1")
+    >>> # if one dimension array and default separator ' ', a tabulation must be added in front of the data line
```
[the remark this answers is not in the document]
alixdamman: where?

alixdamman:
rebased and updated.

larray/io/array.py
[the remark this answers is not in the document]
gdementen: In the changelog. You speak about this in the changelog for version 0.23. So this needs to be mentioned in this version's changelog too.

doc/source/api.rst
gdementen: from_frame needs to appear in the changelog.

doc/source/changes/version_0_26.rst.inc
```diff
@@ -280,6 +297,22 @@ Miscellaneous improvements
...
+
+    >>> from_string('''sex  M  F
+    ...                \t   0  1''')
+    sex  M  F
+         0  1
+    >>> from_string('''sex  M  F
+    ...                nan  0  1''')
+    sex  M  F
+         0  1
+    >>> from_string('''sex  M  F
```
gdementen: I am already a bit sceptical about showing the nan example, but having both NaN and nan examples seems really overkill/ redundant with the fact you already tell you can use either, especially for a feature (from_string) that I think none of our users ever used. Mentioning read_csv here will confuse our users more than anything I think. As far as I understand, this only applies to read_csv when using ' ' as the separator, which nobody ever does.

larray/io/array.py
```diff
@@ -611,10 +611,18 @@ def from_string(s, nb_index=None, index_col=None, sep=' ', **kwargs):
 
     Examples
     --------
-    >>> # if one dimension array and default separator ' ', a tabulation must be added in front of the data line
+    >>> # to create a 1D array using the default separator ' ', a nan, NaN or a tabulation character \t 
+    >>> # must be added in front of the data line
     >>> from_string("sex  M  F\\n\\t  0  1")
     sex  M  F
          0  1
+    >>> from_string("sex  M  F\\nnan  0  1")
```
gdementen: Same as above. IMO, it is enough to only mention that you can use nan or NaN instead of \t

doc/source/changes/version_0_26.rst.inc
```diff
@@ -306,21 +306,14 @@ Miscellaneous improvements
      b [2]: 'b0' 'b1'
     dtype: float64
 
-* To create a 1D array using from_string() or read_csv() and the default separator ' ', a `nan`, `NaN` or a tabulation
-  character `\t` must be used in front of the data line:
+* To create a 1D array using from_string() and the default separator ' ', a tabulation character `\t` must be added
+  in front of the data line:
```
gdementen: s/a tabulation character `\t`/a tabulation character `\t` (instead of `-` previously)/

alixdamman:
I don't know why my last commit is not taken into account.

gdementen:
I don't know either :). Try force pushing your branch?

alixdamman:
I already force pushed my branch. 
I can see my last commit on my fork project https://github.com/alixdamman/larray/commits/formalizing_unstructured_format_460 but for some reasons it does not appear in the current PR?

gdementen:
Strange indeed. Maybe try creating another PR

gdementen:
sure!

PR #470, opened 2017-10-09
include "current" release changes in the changelog

gdementen:
(backported from LIAM2 corresponding script)

### DIFF
--- a/next_release.py
+++ b/next_release.py
@@ -1,18 +1,44 @@
 #!/usr/bin/python
-# coding=utf-8
+# encoding: utf-8
 # script to start a new release cycle
 # Licence: GPLv3
 from os.path import join
-from make_release import relname2fname
+from make_release import relname2fname, short
+from shutil import copy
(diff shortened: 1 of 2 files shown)


next_release.py
```diff
@@ -1,18 +1,44 @@
...
 # script to start a new release cycle
 # Licence: GPLv3
 from os.path import join
-from make_release import relname2fname
+from make_release import relname2fname, short
+from shutil import copy
 
 
 def add_release(release_name):
-    # create "empty" change file for that release
```
alixdamman: `release_name` could not be read directly from meta.yml or by importing larray.__version__? 
Or you want to keep a "generic" method which can be use in several projects?

gdementen: I find it cleaner to have the function take that as an argument. If the caller takes this value from larray.version that's fine (would be better), but I do not want the logic to get to the current version to be duplicated all over the place.

PR #474, opened 2017-10-11
misc

gdementen:


doc/source/changes/version_0_26.rst.inc
```diff
@@ -370,8 +370,11 @@ Fixes
 * fixed loading arrays from Excel sheets containing blank cells below or right of the array to read
   (closes :issue:`443`)
 
-* fixed unary and binary operations between sessions failing entirely when the operation failed/was invalid on any of
-  the array. Now the result will be nan for that array but the operation will carry on for other arrays.
+* fixed unary and binary operations between sessions failing entirely when the operation failed/was invalid on any
+  array. Now the result will be nan for that array but the operation will carry on for other arrays.
+
+* fixed stacking sessions failing entirely when the stacking failed on any array.  Now the result will be nan for that
+  array but the operation will carry on for other arrays.
```
alixdamman: Isn't that a bit dangerous? Do you display any warning?

gdementen: No, but that is probably a good idea to add that. No time for 0.26. If we add it, we should add it for unary/binary ops too. I have opened #475 for this.

Issue #177, opened 2017-03-29
update the release script


gdementen:
It is becoming hard to remember to do everything when doing a release, so we should automate the process as much as possible. Here are a few steps:

- [x]  add new methods/functions in api.rst if not done yet (marked as done because I see no way to automate this)
- [x] update version (`meta.yaml`, `__init__.py` and `setup.py`)
- [x] update changelog date
- [x] git tag release
- [x] git push
- [x] build documentation (we can probably use hooks to get that done automatically)
- [x]  publish on pypi
- [x] update conda-forge feedstock to trigger new packages
- [x] update larrayenv
- [x] announce the release (once appveyor finally built the package for Windows :)) on larray-announce and possibly pyannounce or pydata or whatever.

alixdamman:
Why did you add a bullet `build documentation (we can probably use hooks to get that done automatically)` ? The documentation is build automatically on readthedocs, no?

gdementen:
Yes, but it was not at the time I created the issue!

alixdamman:
assume #411 is done.

alixdamman:
instead of rmtree, use `rd /s /q <directory>`.

gdementen:
@alixdamman shouldn't this issue be closed?

alixdamman:
Considering #411, we open a new issue or we edit this one?

gdementen:
given #411 is closed, I guess we can close this issue, right?

gdementen:
The release script needs to be updated to use the releaser lib. Do you prefer to keep this issue open for that or to close this issue and open a new one?

alixdamman:
I opened a new issue : #537

PR #487, opened 2017-10-17
Update release script (issue 177)

alixdamman:
I would like to proceed step by step. 
I started with `update conda-forge feedstock to trigger new packages` (see #177).

### DIFF
--- a/make_release.py
+++ b/make_release.py
@@ -1,6 +1,9 @@
 #!/usr/bin/python
 # coding=utf-8
+# Release script for LArray
 # Licence: GPLv3
+# Requires:
+# * git
 from __future__ import print_function, unicode_literals
 
 import errno
@@ -11,20 +14,31 @@ import stat
 import subprocess
 import sys
 import zipfile
+import hashlib
+import urllib.request as request
 
 from datetime import date
 from os import chdir, makedirs
-from os.path import exists, getsize, abspath, dirname
+from os.path import exists, abspath, dirname
 from shutil import copytree, copy2, rmtree as _rmtree
 from subprocess import check_output, STDOUT, CalledProcessError
 
+
+PY2 = sys.version_info[0] < 3
+TMP_PATH = r"c:\tmp\larray_new_release"
+TMP_PATH_CONDA = r"c:\tmp\larray_conda_new_release"
+LARRAY_REP = 'https://github.com/liam2/larray'
+LARRAY_READTHEDOCS = "http://larray.readthedocs.io/en/stable/"
+CONDA_LARRAY_FEEDSTOCK_REP = 'https://github.com/larray-project/larray-feedstock.git'
+LARRAY_ANNOUNCE_MAILING_LIST = "larray-announce@googlegroups.com"
+LARRAY_USERS_GROUP = "larray-users@googlegroups.com"
 
 try:
     input = raw_input
 except NameError:
     pass
 
-if sys.version < '3':
+if PY2:
     import io
     # add support for encoding. Slow on Python2, but that is not a problem given what we do with it.
     open = io.open
@@ -42,13 +56,13 @@ def size2str(value):
         if value > 1024.0:
             value /= 1024.0
             unit = "Mb"
-        return "%.2f %s" % (value, unit)
+        return "{:.2f} {}".format(value, unit)
     else:
-        return "%d %s" % (value, unit)
+        return "{:d} {}".format(value, unit)
 
 
 def generate(fname, **kwargs):
-    with open('%s.tmpl' % fname) as in_f, open(fname, 'w') as out_f:
+    with open('{}.tmpl'.format(fname)) as in_f, open(fname, 'w') as out_f:
         out_f.write(in_f.read().format(**kwargs))
 
 
@@ -59,7 +73,7 @@ def _remove_readonly(function, path, excinfo):
         # retry removing
         function(path)
     else:
-        raise
+        raise Exception("Cannot remove {}".format(path))
 
 
 def rmtree(path):
@@ -69,7 +83,7 @@ def rmtree(path):
 def call(*args, **kwargs):
     try:
         res = check_output(*args, stderr=STDOUT, **kwargs)
-        if sys.version >= '3' and 'universal_newlines' not in kwargs:
+        if not PY2 and 'universal_newlines' not in kwargs:
             res = res.decode('utf8')
         return res
     except CalledProcessError as e:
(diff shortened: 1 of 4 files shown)


alixdamman:
Is there an equivalent for `wget` in windows? 
I can download the tar file of the last release with urllib.request.urlopen and then calculate the sha256 but it is different from the one calculated with the tar file downloaded by hand from github ??? Although the size and content seems to be the same.

gdementen:
also I realize now that the make_release.py script is based on a very old version of the script from LIAM2 (https://github.com/liam2/liam2/commit/a3bc9271884e6ad1c5e987492218d1b19ba919d0#diff-e94194a6680f2a567bea5103af5fe9cf). The relevant parts should be updated using the relevant changes until the latest version:

https://github.com/liam2/liam2/compare/a3bc9271884e6ad1c5e987492218d1b19ba919d0...master#diff-e94194a6680f2a567bea5103af5fe9cf

Of note is update_changelog, but also some generic functions and a few bits here and there which were made more robust.

alixdamman:
updated

make_release.py
```diff
@@ -383,6 +383,55 @@ def run_tests():
...
+        return
+
+    chdir(context['build_dir'])
+    version = short(context['release_name'])
+
+    # meta.yaml
+    meta_file = r'condarecipe\larray\meta.yaml'
+    with open(meta_file) as f:
+        lines = f.readlines()
+        lines[2] = "  version: {}\n".format(version)
```
gdementen: this seems fragile, see below.

make_release.py
```diff
@@ -383,6 +383,55 @@ def run_tests():
...
+        lines[5] = "  git_tag: {}\n".format(version)
```
gdementen: too fragile to my taste. See below.

make_release.py
```diff
@@ -383,6 +383,55 @@ def run_tests():
...
+    with open(meta_file, 'w') as f:
+        f.writelines(lines)
+
+    # __init__.py
+    init_file = r'larray\__init__.py'
+    with open(init_file) as f:
+        lines = f.readlines()
+        for i, line in enumerate(lines[:]):
+            if '__version__' in line:
```
gdementen: I would make a generic function for replacing lines in a file and use it for meta, __init__.py and setup.py
```python
def replace_lines(fpath, [(substring_to_find, new_line), (sub2, newline2)]):
    pass
```

make_release.py
```diff
@@ -142,149 +159,185 @@ def short(rel_name):
...
+        f.writelines(lines)
 
-    # include version changelog in changes.rst
-    fpath = r'doc\source\changes.rst'
-    changelog_template = """{title}
-{underline}
 
-Released on {date}.
+def create_source_archive(release_name, rev):
+    call(r'git archive --format zip --output ..\larray-{}-src.zip {}'.format(release_name, rev))
```
alixdamman: using sdist instead.

gdementen: whyN

make_release.py
```diff
@@ -142,149 +163,185 @@ def short(rel_name):
...
 
-def build_doc():
-    chdir('doc')
-    call('buildall.bat')
-    chdir('..')
-
-
-def update_changelog(release_name):
-    fname = relname2fname(release_name)
+def replace_lines(fpath, changes, end="\n"):
```
gdementen: Why have an end argument? Is there a case where it is needed?

make_release.py
```diff
@@ -142,149 +163,185 @@ def short(rel_name):
...
+    Parameters
+    ----------
+    changes : list of pairs
+        List of pairs (substring_to_find, new_line).
+    """
+    with open(fpath) as f:
+        lines = f.readlines()
+        for i, line in enumerate(lines[:]):
+            for substring_to_find, new_line in changes:
+                if substring_to_find in line and '#' not in line:
```
gdementen: I guess the `'#' not in line` is meant to avoid commented lines. In that case I would use something like `line.strip().startswith('#')` to avoid skipping a "used" line with a comment at the end of the line.

make_release.py
```diff
@@ -294,69 +351,317 @@ def make_release(release_name=None, branch='master'):
...
+    (set_context_for_conda_forge, 'Setting context in order to update packages on conda-forge'),
+    (create_tmp_directory, ''),
+    (clone_repository, ''),
+    (update_version_conda_forge_package, ''),
+    (push, ''),
+    (cleanup, 'Cleaning up'),
+    ############################################
+    # UPDATE LARRAY METAPACKAGE ON CONDA-FORGE #
+    ############################################
+    # assume larray package on cond-forge has been updated
```
gdementen: conda-forge

next_release.py
```diff
@@ -29,16 +29,33 @@ def add_release(release_name):
...
+    if no('Does the full changelog look right?'):
+        exit(1)
+    call('git add {}'.format(fpath))
+    call('git commit -m "added release {} as in development in changes.rst"'.format(short(release_name)))
+
+
+def add_release(release_name, branch='master'):
+    update_changelog(release_name)
+    context = {'branch': branch,
+               'release_name': 'dev',
```
gdementen: I would rather have something more descriptive here. Something like `release_name+'dev'`

make_release.py
```diff
@@ -606,6 +621,7 @@ def announce_new_release(context):
     (pull, ''),
     # >>> need internet from here
     (push, ''),
+    (push_on_pypi, 'Pushing on Pypi'),
```
gdementen: I would be nice to have a "push on pypi test server" step before, and potentially split the "push_on_pypi" step into "create_pypi_packages" and "push_on_pypi", so that we can check packages before they are uploaded. This could come later though.

gdementen:
FWIW, here is what I used for the liam2 announce:
https://github.com/liam2/website/blob/master/make_release.py#L220
https://github.com/liam2/website/blob/master/make_release.py#L446

gdementen:
The comments from my review from 2 days ago still stand.

make_release.py
alixdamman: Not now but it might.

next_release.py
```diff
@@ -29,16 +29,33 @@ def add_release(release_name):
...
+
+
+def add_release(release_name, branch='master'):
+    update_changelog(release_name)
+    context = {'branch': branch,
+               'release_name': release_name+'-dev',
+               'repository': abspath(dirname(__file__)),
+               'build_dir': abspath(dirname(__file__)),
+               'public_release': True}
+    update_version(context)
```
gdementen: would it be hard to create a single commit for update_changelog & update_version? If you can do that in less than 10min, then that's probably worth it (cleaner history), otherwise nevermind.

alixdamman:
Squash and merge I guess. I prefer to not use fix 177 in the commit message since it has not been tested in real life yet and current version of `make_release.py` does not include `larray-editor`, `larray-eurostat` and `larrayenv`.

gdementen:
yes, squash & merge indeed :)

gdementen:
nice work!

PR #493, opened 2017-10-23
added reverse arg in methods indicesofsorted and labelsofsorted (issue 490)

alixdamman:


larray/core/array.py
```diff
@@ -3108,18 +3110,24 @@ def labelsofsorted(self, axis=None, kind='quicksort'):
              BE  M  F
              FR  F  M
              IT  M  F
+        >>> arr.labelsofsorted(X.sex, reverse=True)
+        nat\\sex  1  0
```
gdementen: I don't think this is the result we want. At least that was not what I needed when I wanted reverse: I expected:
```python
         nat\\sex  0  1
              BE  F  M
              FR  M  F
              IT  F  M
```
The reasoning is this: I want to sort from highest to lowest, and given that sort, I want to know which is the first element (ie indice 0), which is second(1), etc. Not just display them in flipped order.

larray/core/array.py
```diff
@@ -3151,6 +3161,11 @@ def indicesofsorted(self, axis=None, kind='quicksort'):
               0  2  1
               1  0  2
               2  1  0
+        >>> arr.indicesofsorted(X.nat, reverse=True)
+        nat\\sex  M  F
+              2  1  0
```
gdementen: same comment as above

larray/core/array.py
```diff
@@ -3076,7 +3076,7 @@ def indexofmax(self, axis=None):
 
     posargmax = renamed_to(indexofmax, 'posargmax')
 
-    def labelsofsorted(self, axis=None, kind='quicksort'):
+    def labelsofsorted(self, axis=None, kind='quicksort', reverse=False):
```
gdementen: I wonder if reverse shouldn't be the second argument, as it is much more likely to be used than kind.

larray/core/array.py
```diff
@@ -3151,13 +3160,18 @@ def indicesofsorted(self, axis=None, kind='quicksort'):
...
+              1  0  2
+              2  2  1
         """
         if axis is None:
             if self.ndim > 1:
                 raise ValueError("array has ndim > 1 and no axis specified for posargsort")
             axis = self.axes[0]
         axis, axis_idx = self.axes[axis], self.axes.index(axis)
-        data = self.data.argsort(axis_idx, kind=kind)
+        data = np.argsort(-1 * self.data if reverse else self.data, axis_idx, kind=kind)
```
gdementen: Why not use something like this:
```python
data = self.data.argsort(axis_idx, kind=kind)
if reverse:
    reverser = tuple(slice(None, None, -1) if i == axis_idx else slice(None)
                     for i in range(self.ndim))
    data = data[reverser]
```
this wouldn't make a copy of the data.

larray/core/array.py
```diff
@@ -3168,10 +3168,14 @@ def indicesofsorted(self, axis=None, kind='quicksort', reverse=False):
         """
         if axis is None:
             if self.ndim > 1:
-                raise ValueError("array has ndim > 1 and no axis specified for posargsort")
+                raise ValueError("array has ndim > 1 and no axis specified for indicesofsorted")
```
gdementen: Nice catch! The corresponding message in labelsofsorted needs to be fixed too, though.

larray/core/array.py
```diff
@@ -3129,9 +3136,11 @@ def indicesofsorted(self, axis=None, kind='quicksort'):
         ----------
         axis : int or str or Axis, optional
             Axis along which to sort. This can be omitted if array has only one axis.
+        reverse : bool, optional
+            Sort values in descending order. Defaults to False (ascending order).
         kind : {'quicksort', 'mergesort', 'heapsort'}, optional
             Sorting algorithm. Defaults to 'quicksort'.
-
+
```
gdementen: kill this extra space

alixdamman:
Wainting for 0.26.1 release before to merge

gdementen:
can be merged now

PR #496, opened 2017-10-24
fix #485: fixed LArray.split_axes with sep != '_' or when split labels are ambiguous

### LINKED ISSUE #485: split_axis fails in the presence of ambiguous labels

gdementen:


doc/source/changes/version_0_26_1.rst.inc
```diff
@@ -8,3 +8,6 @@ Fixes
 -----
 
 * fixed reading from and writing to Excel sheets with 16384 columns or 1048576 rows (Excel's maximum).
+
+* fixed LArray.split_axes using a custom separator and not using sort=True or when the split labels are
+  ambiguous with labels from other axes (closes :issue:`485`).
```
alixdamman: If you have an example for one or both situations, that would be nice. Otherwise, forget it.

gdementen: I have examples for both (see added tests) but I am sceptical adding an example in the changelog for a bug which was not present in the release before adds any value.

alixdamman:
You can push then.

Issue #54, opened 2017-01-23
implement insert on LArray and Axis


gdementen:
and possibly Group

PR #512, opened 2017-11-07
#54: implemented LArray.insert and Axis.insert

gdementen:
a few notes:

* concat is purposefully not documented yet, because I would like to at least try to merge it with stack
* append and prepend now support values which already contain the axis being worked on
  - this change is not documented (yet) because I am not 100% sure we should support this (I am leaning towards supporting it)
  - if we decide to support this, it will make .extend obsolete (.append is effectively a superset of .extend)

### DIFF
--- a/larray/core/array.py
+++ b/larray/core/array.py
@@ -5395,8 +5395,8 @@ class LArray(ABCLArray):
         Parameters
         ----------
         axis : axis reference
-            Axis along with to append input array (`value`).
-        value : LArray
+            Axis along which to append input array (`value`).
+        value : scalar or LArray
             Array with compatible axes.
         label : str, optional
             Label for the new item in axis
--- a/larray/core/array.py
+++ b/larray/core/array.py
@@ -15,7 +15,8 @@ Matrix class
 # ? implement multi group in one axis getitem: lipro['P01,P02;P05'] <=> (lipro['P01,P02'], lipro['P05'])
 
 # * we need an API to get to the "next" label. Sometimes, we want to use label+1, but that is problematic when labels
-#   are not numeric, or have not a step of 1. x.agegroup[x.agegroup.after(25):]
+#   are not numeric, or have not a step of 1. X.agegroup[X.agegroup.after(25):]
+#                                             X.agegroup[X.agegroup[25].next():]
 
 # * implement keepaxes=True for _group_aggregate instead of/in addition to group tuples
 
@@ -66,7 +67,9 @@ from larray.core.group import (Group, IGroup, LGroup, remove_nested_groups, _to_
                                _range_to_slice, _translate_sheet_name, _translate_key_hdf)
 from larray.core.axis import Axis, AxisReference, AxisCollection, X, _make_axis
 from larray.util.misc import (table2str, size2str, basestring, izip, rproduct, ReprString, duplicates,
-                              float_error_handler_factory, _isnoneslice, light_product, unique_list, renamed_to)
+                              float_error_handler_factory, _isnoneslice, light_product, unique_list, renamed_to,
+                              common_type)
+
 
 nan = np.nan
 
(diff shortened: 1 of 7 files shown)


larray/core/array.py
```diff
@@ -5572,6 +5542,201 @@ def extend(self, axis, other):
...
+        #  a0   0    42    43   1    42    43   2
+        #  a1   3    44    45   4    44    45   5
+
+        # The later looks less useful and could be emulated easily via:
+        # arr1.insert([arr3, arr3], before='b1,b2')
+        # while the above is a bit harder to achieve manually:
+        # arr1.insert([arr3[[b]] for b in arr3.b], before=['b1', 'b2'])
+        # but the later is *probably* more intuitive (and wouldn't suffer from the inefficiency we currently have).
+
+        # XXX: when we have several lists, we implicitly match them by position, which we should avoid for the usual
```
alixdamman: several values ? I'm not sure what you are talking about with "lists".

larray/core/array.py
```diff
@@ -5572,6 +5542,201 @@ def extend(self, axis, other):
...
+        #  a1   3   4   5
+        # >>> arr2 = full('b=b0.5', 8)
+        # >>> arr2
+        # b  b0.5
+        #       8
+        # >>> arr3 = full('b=b1.5', 9)
+        # >>> arr3
+        # b  b1.5
+        #       9
+        # >>> arr1.insert(before={'b1': arr2, 'b2': arr3})
```
alixdamman: sounds fair

larray/core/array.py
```diff
@@ -5572,6 +5542,201 @@ def extend(self, axis, other):
...
+        # a\\b  b0  b0.5  b1  b1.5  b2
+        #  a0   0     8   1     9   2
+        #  a1   3     8   4     9   5
+
+        # When the inserted arrays/values have no label, this does not really convince me and it prevents using after
```
alixdamman: or we could ask to pass 3 lists: 1 for value, 1 for before or after and 1 for label (they must have the same length).

larray/core/array.py
gdementen: when we pass a list (instead of a scalar) for several of "value", "before/after/pos" and "label".

larray/core/array.py
gdementen: that is precisely what is already implemented and what I think we should avoid, because it matches items by position within each list, which for longer lists is dangerous (for the same reason stack(list_of_values, axis) is not recommended): it will let silly inversion errors pass silently and those will be very hard to find.

larray/core/array.py
```diff
@@ -5572,6 +5542,201 @@ def extend(self, axis, other):
...
+        #  a1   3     8   4     9   5
+
+        # This is shorter but not readable enough/even more arbitrary than the previous option.
+
+        # >>> arr1.insert([(8, 'b1', 'b0.5'), (9, 'b2', 'b1.5')])
+        # a\b  b0  b0.5  b1  b1.5  b2
+        #  a0   0     8   1     9   2
+        #  a1   3     8   4     9   5
+
+        # This is readable but odd and not much gained (except efficiency) compared with multiple insert calls
```
alixdamman: Or a list of dictionary {'value': XX, 'before': YY, 'label': ZZ}. Why not let users to use insert in a loop?

gdementen: they can of course, but it is awfully slow compared to one insert call (because it reallocates & copies the array for each iteration of the loop) while it is possible to do it in one call.

larray/core/array.py
```diff
@@ -5572,6 +5542,201 @@ def extend(self, axis, other):
...
+         a1   3    44    45   4   5
+        """
+
+        # XXX: unsure we should have arr1.insert(arr3, before='b1,b2') result in (see unit tests):
+
+        # a\\b  b0  b0.1  b1  b0.2  b2
+        #  a0   0    42   1    43   2
+        #  a1   3    44   4    45   5
+
+        # we might to implement the following instead:
```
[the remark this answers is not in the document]
gdementen: unsure it is worth having an extra argument. At least let us wait until somebody actually needs the functionality

larray/core/array.py
gdementen: The list of dict seems like a good idea. Might just be the compromise syntax I was looking for. I have to see what it looks like in a example before I can decide on this, though.

gdementen:
Rebased & added fix for a warning. I will merge if Travis is happy with this.

Issue #478, opened 2017-10-13
implement array.sort_values() for 1D arrays


gdementen:
ie no argument necessary.

see also #225.

current workaround is to expand, which is very silly:

```python
arr = ndtest(10)
arr['a5'] = -1
arr['a3'] = 10
arr.expand(Axis(['_'], '_')).sort_values('_')['_']
```

PR #524, opened 2017-11-14
fix #225 and #478 : update LArray.sort_values

### LINKED ISSUE #225: implement larray.sort_values(axis=)
yes, the labels in that case are meaningless (should be 0..N), but it is still useful. Eg, from dynapor: take M best salary out of last N.

alixdamman:


doc/source/changes/version_0_27.rst.inc
```diff
@@ -61,6 +61,42 @@ Miscellaneous improvements
 
 * added support for coloring numeric values for object arrays (e.g. arrays containing both strings and numbers).
 
+* allowed to sort values of an array along an axis (closes :issue:`225`):
```
gdementen: move this up in the list of misc improvements. This is probably one of the most interesting improvements

doc/source/changes/version_0_27.rst.inc
```diff
@@ -61,6 +61,42 @@ Miscellaneous improvements
...
+    >>> a = LArray([[10, 2, 4], [3, 7, 1]], "sex=M,F; nat=EU,FO,BE")
+    >>> a
+    sex\\nat  EU  FO  BE
```
gdementen: Extra `\` between axes names or missing one in the following examples. I don't remember if we need the second \ in the changelog but it should at least be consistent.

larray/core/array.py
```diff
@@ -1603,55 +1609,110 @@ def sort_values(self, key, reverse=False):
...
+        a1   b1   9   2   8
+                   
+        Sort along an axis
+        
+        >>> arr_2D
+        a\\b  b0  b1  b2
+         a0  10   2   4
+         a1   3   7   1
+        >>> # sort values along axis 'a'
+        >>> # equivalent to sort values of each column of array arr_2D
```
gdementen: equivalent to sorting the values of each column of the array

larray/core/array.py
```diff
@@ -1603,55 +1609,110 @@ def sort_values(self, key, reverse=False):
...
+        >>> arr_2D.sort_values(axis='a')
+        a*\\b  b0  b1  b2
+           0   3   2   1
+           1  10   7   4
+        >>> arr_3D
```
gdementen: I do not think having an example of sorting a 3D array along an axis in the docstring is worth its space because this is really the same thing as for the 2D array. The 3D array is fine for sorting along a key (because it is different from the 2D case -- it needs a tuple instead of a scalar). So I would move this to a unit test and **possibly** replace it by arr_2D.sort_values(axis='b')

PR #526, opened 2017-11-17
fix #409 : added documentation for Workbook class

### LINKED ISSUE #409: Add Workbook methods in documentation
open_excel returns a Workbook but Pycharm does not know that and does not provide completion.

alixdamman:
For release 0.28

output here --> http://larray-test.readthedocs.io/en/add_workbook_doc_to_api_ref_409/

### DIFF
--- a/larray/inout/excel.py
+++ b/larray/inout/excel.py
@@ -573,6 +573,9 @@ else:
         def save(self, path=None):
             raise Exception()
 
+        def close(self):
+            raise Exception()
+
     def open_excel(filepath=None, overwrite_file=False, visible=None, silent=None, app=None):
         raise Exception("open_excel() is not available because xlwings is not installed")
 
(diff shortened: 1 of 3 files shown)


larray/inout/excel.py
```diff
@@ -555,13 +559,73 @@ def load(self, header=True, convert_float=True, nb_index=None, index_col=None, f
...
         return Workbook(filepath, overwrite_file, visible, silent, app)
 else:
+    class Workbook(object):
+        def __init__(self, filepath=None, overwrite_file=False, visible=None, silent=None, app=None):
+            raise Exception("Workbook class cannot be instanciated because xlwings is not installed")
+
+        def sheet_names(self):
+            raise Exception()
+
+        def save(self, path=None):
```
gdementen: close method should also be documented/be part of the api

PR #539, opened 2017-11-29
fix issue 531 : Add support for groups in reindex 

alixdamman:
Needs changelog.

### DIFF
--- a/larray/core/array.py
+++ b/larray/core/array.py
@@ -1342,52 +1342,48 @@ class LArray(ABCLArray):
         a\\b  b0  b1
          a0   0   1
          a1   2   3
+        >>> arr2 = ndrange('a=a1,a2;c=c0;b=b2..b0')
+        >>> arr2
+         a  c\\b  b2  b1  b0
+        a1   c0   0   1   2
+        a2   c0   3   4   5
 
         Reindex one axis
 
-        >>> arr.reindex(X.b, ['b1', 'b2', 'b0'], fill_value=-1)
-        a\\b  b1  b2  b0
-         a0   1  -1   0
-         a1   3  -1   2
-        >>> arr.reindex(X.b, 'b0..b2', fill_value=-1)
+        >>> arr.reindex('b', ['b1', 'b2', 'b0'])
+        a\\b   b1   b2   b0
+         a0  1.0  nan  0.0
+         a1  3.0  nan  2.0
+        >>> arr.reindex('b', 'b0..b2', fill_value=-1)
         a\\b  b0  b1  b2
          a0   0   1  -1
          a1   2   3  -1
+        >>> arr.reindex('b', arr2.b, fill_value=-1)
+        a\\b  b2  b1  b0
+         a0  -1   1   0
+         a1  -1   3   2
 
         Reindex several axes
 
-        >>> a = Axis(['a1', 'a2', 'a0'], 'a')
-        >>> b = Axis(['b2', 'b1', 'b0'], 'b')
-        >>> arr.reindex({'a': a, 'b': b}, fill_value=-1)
+        >>> arr.reindex({'a': arr2.a, 'b': arr2.b}, fill_value=-1)
         a\\b  b2  b1  b0
          a1  -1   3   2
          a2  -1  -1  -1
-         a0  -1   1   0
-        >>> arr.reindex({X.a: a, X.b: b})
-        a\\b   b2   b1   b0
-         a1  nan  3.0  2.0
-         a2  nan  nan  nan
-         a0  nan  1.0  0.0
 
-        Reindex using axes from another array
+        Reindex by passing a collection of axes
 
-        >>> arr2 = ndrange('a=a0,a1;c=c0..c0;b=b0..b2')
-        >>> arr2
-         a  c\\b  b0  b1  b2
-        a0   c0   0   1   2
-        a1   c0   3   4   5
-        >>> arr.reindex(arr2.axes)
-         a  b\\c   c0
-        a0   b0  0.0
-        a0   b1  1.0
-        a0   b2  nan
-        a1   b0  2.0
-        a1   b1  3.0
-        a1   b2  nan
-        >>> arr2.reindex(arr.axes)
-         a  c\\b   b0   b1
-        a0   c0  0.0  1.0
-        a1   c0  3.0  4.0
+        >>> arr.reindex(arr2.axes, fill_value=-1)
+         a  b\\c  c0
+        a1   b2  -1
+        a1   b1   3
+        a1   b0   2
+        a2   b2  -1
+        a2   b1  -1
+        a2   b0  -1
+        >>> arr2.reindex(arr.axes, fill_value=-1)
+         a  c\\b  b0  b1
+        a0   c0  -1  -1
+        a1   c0   2   1
         """
         # XXX: can't we move this to AxisCollection.replace?
         if isinstance(new_axis, (int, basestring, list, tuple, np.ndarray)):
--- a/larray/core/array.py
+++ b/larray/core/array.py
@@ -1400,7 +1400,7 @@ class LArray(ABCLArray):
         a1   c0   2   1
         """
         # XXX: can't we move this to AxisCollection.replace?
-        if isinstance(new_axis, (int, basestring, list, tuple, np.ndarray)):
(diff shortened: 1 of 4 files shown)


larray/core/array.py
```diff
@@ -1342,52 +1342,55 @@ def reindex(self, axes_to_reindex=None, new_axis=None, fill_value=np.nan, inplac
...
-        >>> arr.reindex(X.b, 'b0..b2', fill_value=-1)
+        >>> arr.reindex('b', ['b1', 'b2', 'b0'])
+        a\\b   b1   b2   b0
+         a0  1.0  nan  0.0
+         a1  3.0  nan  2.0
+        >>> arr.reindex('b', 'b0..b2', fill_value=-1)
         a\\b  b0  b1  b2
          a0   0   1  -1
          a1   2   3  -1
+        >>> arr.reindex('b', arr2.b, fill_value=-1)
```
gdementen: add a title/comment "Reindex using axes from another array"?

larray/core/array.py
```diff
@@ -1342,52 +1342,55 @@ def reindex(self, axes_to_reindex=None, new_axis=None, fill_value=np.nan, inplac
...
-        a1   b2  nan
-        >>> arr2.reindex(arr.axes)
-         a  c\\b   b0   b1
-        a0   c0  0.0  1.0
-        a1   c0  3.0  4.0
+        >>> arr.reindex({'a': arr2.a['a1'], 'b': arr2.b['b1':]}, fill_value=-1)
+        a\\b  b1  b0
+         a1   3   2
+
+        Reindex by passing a collection of axes
```
gdementen: the documentation for axes_to_reindex needs to be updated/fixed, 

* no mention of dict in the description (only in the type)
* the sentence: 
   > In that case, the number of new axes must match the number of the old ones.

   is wrong (as shown by the examples below).

larray/core/array.py
```diff
@@ -1342,52 +1342,55 @@ def reindex(self, axes_to_reindex=None, new_axis=None, fill_value=np.nan, inplac
...
-        a0   b1  1.0
-        a0   b2  nan
-        a1   b0  2.0
-        a1   b1  3.0
-        a1   b2  nan
-        >>> arr2.reindex(arr.axes)
-         a  c\\b   b0   b1
-        a0   c0  0.0  1.0
-        a1   c0  3.0  4.0
+        >>> arr.reindex({'a': arr2.a['a1'], 'b': arr2.b['b1':]}, fill_value=-1)
```
gdementen: unsure showing reindexing to a single label is worth it. It is likely to confuse people more than it helps. I would scrap this example from the docstring. It is an interesting unit test for several reasons though.

larray/core/array.py
```diff
@@ -1342,52 +1342,55 @@ def reindex(self, axes_to_reindex=None, new_axis=None, fill_value=np.nan, inplac
...
+         a1  3.0  nan  2.0
+        >>> arr.reindex('b', 'b0..b2', fill_value=-1)
         a\\b  b0  b1  b2
          a0   0   1  -1
          a1   2   3  -1
+        >>> arr.reindex('b', arr2.b, fill_value=-1)
+        a\\b  b2  b1  b0
+         a0  -1   1   0
+         a1  -1   3   2
+       >>> arr.reindex('b', arr2.b['b1':], fill_value=-1)
```
gdementen: add title/comment: "Reindex using a subset of an axis"?

larray/core/axis.py
```diff
@@ -1411,7 +1411,7 @@ def slice_bound(bound):
             for k, v in zip(key, value):
                 self[k] = v
         else:
-            if isinstance(value, (int, basestring, list, tuple)):
+            if isinstance(value, (int, basestring, list, tuple, Group)):
```
gdementen: if wonder if we shouldn't rather write: `if not isinstance(value, Axis): ...`

larray/core/array.py
```diff
@@ -1342,52 +1342,55 @@ def reindex(self, axes_to_reindex=None, new_axis=None, fill_value=np.nan, inplac
...
+        a1   b2  -1
+        a1   b1   3
+        a1   b0   2
+        a2   b2  -1
+        a2   b1  -1
+        a2   b0  -1
+        >>> arr2.reindex(arr.axes, fill_value=-1)
+         a  c\\b  b0  b1
+        a0   c0  -1  -1
+        a1   c0   2   1
```
gdementen: replace the `if isinstance(new_axis, (int, basestring, list, tuple, np.ndarray)):` test below by: `if new_axis is not None and not isinstance(new_axis, Axis):`?

larray/core/array.py
alixdamman: That comment also applies for `AxisCollection.replace`

alixdamman: the sentence: 
   > In that case, the number of new axes must match the number of the old ones.
is true when passing a list of Axis (because this how `AxisCollection.replace` works) but not when passing an AxisCollection. 
We may also want to change the behavior of `AxisCollection.replace`. Do I open a new issue for that?

gdementen: I don't think AxisCollection.replace needs to be updated because I don't think it is more dangerous than useful in general to replace the entire collection with more axes that we had initially.
Now the question is whether we want to make arr.reindex(list(arr2.axes)) work. That could make sense, though that is very low priority. We would need one more special case (ie do not pass the list case through AxisCollection.replace).
In any case, the description of axes_to_reindex needs to be updated.

larray/core/array.py
```diff
@@ -1312,9 +1312,9 @@ def reindex(self, axes_to_reindex=None, new_axis=None, fill_value=np.nan, inplac
         axes_to_reindex : axis ref or dict {axis ref: axis} or list of tuple (axis ref, axis) \
                           or list of Axis or AxisCollection
             Axes to reindex. If a single axis reference is given, the `new_axis` argument must be provided.
-            If a list of Axis or an AxisCollection is given, all axes will be reindexed by the new ones.
-            In that case, the number of new axes must match the number of the old ones.
-        new_axis : int, str, list/tuple/array of str or Axis, optional
+            If a list of Axis or an AxisCollection is given, already present axes are reindexed while missing
```
gdementen: s/already present/existing/?
s/missing/missing ones/ or "missing axes"

larray/core/array.py
```diff
@@ -1342,55 +1342,65 @@ def reindex(self, axes_to_reindex=None, new_axis=None, fill_value=np.nan, inplac
...
+        >>> arr2.reindex(arr.axes, fill_value=-1)
+         a  c\\b  b0  b1
+        a0   c0  -1  -1
+        a1   c0   2   1
         """
         # XXX: can't we move this to AxisCollection.replace?
-        if isinstance(new_axis, (int, basestring, list, tuple, np.ndarray)):
+        if new_axis is not None and not isinstance(new_axis, Axis):
             new_axis = Axis(new_axis, self.axes[axes_to_reindex].name)
         if isinstance(new_axis, Axis):
```
gdementen: change to elif

larray/core/axis.py
```diff
@@ -1411,7 +1411,7 @@ def slice_bound(bound):
             for k, v in zip(key, value):
                 self[k] = v
         else:
-            if isinstance(value, (int, basestring, list, tuple)):
+            if not isinstance(value, Axis):
                 value = Axis(value)
             assert isinstance(value, Axis)
```
gdementen: the assert is a bit overkill now

larray/core/array.py
```diff
@@ -1312,8 +1312,7 @@ def reindex(self, axes_to_reindex=None, new_axis=None, fill_value=np.nan, inplac
         axes_to_reindex : axis ref or dict {axis ref: axis} or list of tuple (axis ref, axis) \
                           or list of Axis or AxisCollection
             Axes to reindex. If a single axis reference is given, the `new_axis` argument must be provided.
-            If a list of Axis or an AxisCollection is given, already present axes are reindexed while missing
-            are added.
+            If a list of Axis or an AxisCollection is given, existing axes are reindexed while missing ones are added.
```
gdementen: I wonder if this is actually true. From looking at the code I suspect this is true for AxisCollection but not for a list of Axis?!

alixdamman: See new commit

PR #541, opened 2017-11-29
fix issue 532 : Add support for groups in growth rate

alixdamman:
Needs changelog

### DIFF
--- a/larray/core/array.py
+++ b/larray/core/array.py
@@ -6507,7 +6507,7 @@ class LArray(ABCLArray):
         """
         if isinstance(axis, Group):
             array = self[axis]
-            axis = Axis(axis)
+            axis = array.axes[axis.axis]
         else:
             array = self
         for _ in range(n):
@@ -6573,11 +6573,12 @@ class LArray(ABCLArray):
         """
         if isinstance(axis, Group):
             array = self[axis]
-            axis = Axis(axis)
+            axis = array.axes[axis.axis]
         else:
             array = self
+            axis = array.axes[axis]
         diff = array.diff(axis=axis, d=d, label=label)
-        return diff / array[array.axes[axis].i[:-d]].drop_labels(axis)
+        return diff / array[axis.i[:-d]].drop_labels(axis)
 
     def compact(self):
         """Detects and removes "useless" axes (ie axes for which values are constant over the whole axis)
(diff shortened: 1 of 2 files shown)


alixdamman:
@gdementen changelog added. If OK, I squash and merge.

larray/core/array.py
```diff
@@ -6539,26 +6545,39 @@ def growth_rate(self, axis=-1, d=1, label='upper'):
...
+               F  -0.5   1.0   0.5
         """
-        diff = self.diff(axis=axis, d=d, label=label)
-        axis_obj = self.axes[axis]
-        return diff / self[axis_obj.i[:-d]].drop_labels(axis)
+        if isinstance(axis, Group):
+            array = self[axis]
+            axis = Axis(axis)
+        else:
+            array = self
```
gdementen: I would rather have something like this:
```python
array = self[axis] if isinstance(axis, Group) else self
axis = array.axes[axis]
diff = array.diff(axis=axis, d=d, label=label)
return diff / array[axis.i[:-d]].drop_labels(axis)
```
so that a new axis is not created needlessly.

larray/core/array.py
```diff
@@ -6497,11 +6497,19 @@ def diff(self, axis=-1, d=1, n=1, label='upper'):
...
+        sex\\type  type3
+               M      2
+               F      5
         """
-        array = self
+        if isinstance(axis, Group):
+            array = self[axis]
+            axis = Axis(axis)
+        else:
+            array = self
```
gdementen: same remark as below

larray/core/array.py
alixdamman: line 
```python
axis = array.axes[axis]
```
fails when passing a group:
 >  File "C:\Users\ald\Projects\larray\larray\core\array.py", line 6509, in diff
 >   axis = array.axes[axis]
 >
 > File "C:\Users\ald\Projects\larray\larray\core\axis.py", line 1333, in __getitem__
 >  assert isinstance(key, basestring), type(key)

alixdamman: for groups, it should be 
```python
axis = array.axes[axis.axis]
```
right?

gdementen: ok, but it should still be possible to avoid create a **new** Axis. Maybe something like:
```python
array = self[axis] if isinstance(axis, Group) else self
axis = axis.axis if isinstance(axis, Group) else axis
axis = array.axes[axis]
```

gdementen: you beat me to it :)

Issue #518, opened 2017-11-09
implement LArray.equals


gdementen:
instead of larray_equal

alixdamman:
OK

PR #552, opened 2017-12-04
Implement Session.equals (517) and LArray.equals (518) + make Session.__(n)eq__ use _binop(516)

alixdamman:


doc/source/changes/version_0_28.rst.inc
```diff
@@ -1,7 +1,63 @@
...
+﻿Backward incompatible changes
+-----------------------------
+
+* changed behavior of operators `session1 == session2` and `session1 != session2`: returns a boolean
+  (closes:issue:`516`):
+
+    >>> s1 = Session([('arr1', ndtest(2)), ('arr2', ndtest((2, 2)))])
+    >>> s2 = Session([('arr1', ndtest(2)), ('arr2', ndtest((2, 2)))])
+    >>> # equivalent to all(s1.equals(s2))
+    >>> s1 == s2
```
gdementen: this whole example is wrong/would not work.
after your change, s1 == s2 will return something like:
```python
>>> # should be almost equivalent to Session([(k, s1[k] == s2[k]) for k in s1.keys()])
>>> res = s1 == s2
>>> res.arr1
a    a0    a1
   True  True
>>> res.arr2
a\b    b0    b1
 a0  True  True
 a1  True  True
```

doc/source/changes/version_0_28.rst.inc
```diff
@@ -1,7 +1,63 @@
...
+    True
+    >>> s2.arr1['a1'] = 0
+    >>> s1 == s2
+    False
+    >>> # equivalent to any(~s1.equals(s2))
+    >>> s1 != s2
+    True
+    >>> if s1 == s2:
```
gdementen: unsure the if block adds any value

larray/tests/test_session.py
```diff
@@ -275,44 +275,40 @@ def test_to_globals(self):
...
+        assert res.axes.names == ['name']
+        assert np.array_equal(res.axes.labels[0], ['e', 'g', 'f'])
+        assert list(res) == [False, False, True]
 
-    def test_ne(self):
+    def test_eq(self):
         sess = self.session.filter(kind=LArray)
         expected = Session([('e', self.e), ('f', self.f), ('g', self.g)])
-        self.assertFalse(any(sess != expected))
+        assert sess == expected
```
gdementen: the test is wrong and should be expanded to actually test that the result is correct (as explained in the changelog)

larray/tests/test_session.py
```diff
@@ -275,44 +275,40 @@ def test_to_globals(self):
...
 
+    def test_ne(self):
+        sess = self.session.filter(kind=LArray)
         other = Session({'e': self.e, 'f': self.f})
-        res = sess != other
-        self.assertEqual(res.axes.names, ['name'])
-        self.assertTrue(np.array_equal(res.axes.labels[0], ['e', 'g', 'f']))
-        self.assertEqual(list(res), [False, True, False])
+        assert sess != other
```
gdementen: idem

doc/source/changes/version_0_28.rst.inc
```diff
@@ -1,7 +1,63 @@
...
+
+New features
 ------------
 
-* added a feature (see the :ref:`miscellaneous section <misc>` for details).
+* added method `equals` to `Session` object to compare arrays from two sessions (closes:issue:`517`):
+
+    >>> s1 = Session([('arr1', ndtest(2)), ('arr2', ndtest((2, 2)))])
+    >>> s2 = Session([('arr1', ndtest(2)), ('arr2', ndtest((2, 2)))])
+    >>> s1.equals(s2)
```
gdementen: I **_wonder_**, if we should not rather return a single value (True) like Axis.equals. I know this is not what I wrote in the issue (my bad!), but the general idea in making those changes is to make a Session more like an LArray with one special "array" axis, and given #518, it would make more sense to return a single value. Now to get the old s1 == s2 behaviour, one would need to write: `(s1 == s2).all_by('array')`, but this is not implemented yet and possibly a bit cumbersome to type, so we might want to make Session.equals(other) return a bool and create another method (e.g. Session.array_equals or something like that) which would have the old == behavior. Thoughts?

alixdamman:
reminder: add #518 in this PR

PR #556, opened 2017-12-08
issue 513 (exlude arrays starting with '_') + issue 514 (binary operations on Sessions with non LArray operand) + 515 (rXXX binary ops on Sessions)

alixdamman:


doc/source/changes/version_0_28.rst.inc
```diff
@@ -10,6 +10,67 @@
...
+    ...     # use flag 'include_private' to include arrays starting with '_'
+    ...     s = arrays(include_private=True)
+    ...     print(s.names)
+    >>> foo()
+    ['global_arr1', 'local_arr1']
+    ['_global_arr2', '_local_arr2', 'global_arr1', 'local_arr1']
+
+* implemented sessions binary operations with non sessions objects (closes :issue:`514` and :issue:`515`):
+
+    >>> s = Session({'arr1': ndtest((2, 2)), 'arr2': ndtest((3, 3))})
```
gdementen: use kwargs instead of dict init?

larray/core/session.py
```diff
@@ -908,43 +915,55 @@ def summary(self, template=None):
         return '\n'.join(template.format(**kwargs) for kwargs in templ_kwargs)
 
 
-def local_arrays(depth=0):
+def _exclude_private_array(vars_dict):
```
gdementen: s/_exclude_private_array/_exclude_private_vars/ ? (this is not specific to arrays)

doc/source/changes/version_0_28.rst.inc
```diff
@@ -10,6 +10,67 @@
...
+    a\b  b0  b1
+     a0   3   4
+     a1   5   6
+    >>> s2.arr2
+    a\b  b0  b1  b2
+     a0   3   4   5
+     a1   6   7   8
+     a2   9  10  11
+
+    Define what to add (array or scalar) using a dictionary (or a session)
```
gdementen: this example is nifty but likely to confuse users. I would scrap it. If you want another example, show something that is more likely to be used: eg. percentage of relative difference (relative diff * 100), or revalorisation of amounts by time (this could be very useful for some of our users).

alixdamman: "revalorisation of amounts by time" --> no idea what you mean... sorry

alixdamman:
@gdementen You can skip the first 3 commits (already reviewed).

gdementen:
Say you have a session with various arrays with monetary amounts with
different axes but all have the same time (or year) axis. Next to that you
have one array with inflation coefficients (e.g. 2010 = 1, 2011=1.02,
2012=1.03).

you can pass all your arrays from prices with inflation to prices without
inflation or vice versa (there is a better term for this but I don't
remember it now -- constant prices?) with a single line:

session / array
or
session * array

PR #558, opened 2017-12-11
(issue 489): Axis.union/intersection/difference fails when single string is passed

alixdamman:
Also updated all doctests for those methods in both Axis and Group classes.

### DIFF
--- a/doc/source/changes/version_0_28.rst.inc
+++ b/doc/source/changes/version_0_28.rst.inc
@@ -151,3 +151,18 @@ Fixes
     >>> axis = Axis("a=a0,a1")
     >>> axis['a0'][0]
     'a'
+
+* fixed `Axis.union`, `Axis.intersection` and `Axis.difference` when passed value is a single string
+  (closes :issue:`489`):
+
+    >>> a = Axis('a=a0..a2')
+    >>> a.union('a1')
+    Axis(['a0', 'a1', 'a2'], 'a')
+    >>> a.union('a3')
+    Axis(['a0', 'a1', 'a2', 'a3'], 'a')
+    >>> a.union('a1..a3')
+    Axis(['a0', 'a1', 'a2', 'a3'], 'a')
+    >>> a.intersection('a1..a3')
+    Axis(['a1', 'a2'], 'a')
+    >>> a.difference('a1..a3')
+    Axis(['a0'], 'a')
--- a/larray/core/axis.py
+++ b/larray/core/axis.py
@@ -1067,12 +1067,21 @@ class Axis(ABCAxis):
 
         Examples
         --------
-        >>> letters = Axis('letters=a,b')
-        >>> letters.union(Axis('letters=b,c'))
-        Axis(['a', 'b', 'c'], 'letters')
-        >>> letters.union(['b', 'c'])
-        Axis(['a', 'b', 'c'], 'letters')
+        >>> a = Axis('a=a0..a2')
+        >>> a.union('a1')
+        Axis(['a0', 'a1', 'a2'], 'a')
+        >>> a.union('a3')
+        Axis(['a0', 'a1', 'a2', 'a3'], 'a')
+        >>> a.union(Axis('a=a1..a3'))
+        Axis(['a0', 'a1', 'a2', 'a3'], 'a')
+        >>> a.union('a1..a3')
+        Axis(['a0', 'a1', 'a2', 'a3'], 'a')
+        >>> a.union(['a1', 'a2', 'a3'])
+        Axis(['a0', 'a1', 'a2', 'a3'], 'a')
         """
+        if isinstance(other, basestring):
+            # TODO : remove [other] if ... when FuturWarning raised in Axis.init will be removed
+            other = _to_ticks(other, parse_single_int=True) if '..' in other or ',' in other else [other]
         if isinstance(other, Axis):
             other = other.labels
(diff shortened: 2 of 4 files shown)


larray/core/axis.py
```diff
@@ -1067,12 +1067,17 @@ def union(self, other):
...
+        >>> a.union(Axis('a=a1..a3'))
+        Axis(['a0', 'a1', 'a2', 'a3'], 'a')
+        >>> a.union('a1..a3')
+        Axis(['a0', 'a1', 'a2', 'a3'], 'a')
+        >>> a.union(['a1', 'a2', 'a3'])
+        Axis(['a0', 'a1', 'a2', 'a3'], 'a')
         """
+        if isinstance(other, basestring):
+            # TODO : remove [other] if ... when FuturWarning raised in Axis.init will be removed
+            other = [other] if '..' not in other and ',' not in other else Axis(other)
```
gdementen: use `_to_ticks(other, parse_single_int=True) if '..' in other or ',' in other else [other]` instead to avoid creating an Axis just to throw it away two lines below?

larray/core/axis.py
```diff
@@ -1098,12 +1103,17 @@ def intersection(self, other):
...
         --------
-        >>> letters = Axis('letters=a,b')
-        >>> letters.intersection(Axis('letters=b,c'))
-        Axis(['b'], 'letters')
-        >>> letters.intersection(['b', 'c'])
-        Axis(['b'], 'letters')
+        >>> a = Axis('a=a0..a2')
+        >>> a.intersection(Axis('a=a1..a3'))
+        Axis(['a1', 'a2'], 'a')
+        >>> a.intersection('a1..a3')
```
gdementen: also add a test with a single label

larray/core/axis.py
```diff
@@ -1067,12 +1067,17 @@ def union(self, other):
...
         --------
-        >>> letters = Axis('letters=a,b')
-        >>> letters.union(Axis('letters=b,c'))
-        Axis(['a', 'b', 'c'], 'letters')
-        >>> letters.union(['b', 'c'])
-        Axis(['a', 'b', 'c'], 'letters')
+        >>> a = Axis('a=a0..a2')
+        >>> a.union(Axis('a=a1..a3'))
+        Axis(['a0', 'a1', 'a2', 'a3'], 'a')
+        >>> a.union('a1..a3')
```
gdementen: also add a test with a single label, e.g. `a.union('a1')` or `a.union('a3')`

larray/core/axis.py
```diff
@@ -1126,12 +1136,17 @@ def difference(self, other):
...
-        >>> letters = Axis('letters=a,b')
-        >>> letters.difference(Axis('letters=b,c'))
-        Axis(['a'], 'letters')
-        >>> letters.difference(['b', 'c'])
-        Axis(['a'], 'letters')
-        """
+        >>> a = Axis('a=a0..a2')
+        >>> a.difference(Axis('a=a1..a3'))
+        Axis(['a0'], 'a')
+        >>> a.difference('a1..a3')
```
gdementen: also add a test with a single label

larray/core/axis.py
```diff
@@ -1098,12 +1103,17 @@ def intersection(self, other):
...
+        >>> a.intersection(Axis('a=a1..a3'))
+        Axis(['a1', 'a2'], 'a')
+        >>> a.intersection('a1..a3')
+        Axis(['a1', 'a2'], 'a')
+        >>> a.intersection(['a1', 'a2', 'a3'])
+        Axis(['a1', 'a2'], 'a')
         """
+        if isinstance(other, basestring):
+            # TODO : remove [other] if ... when FuturWarning raised in Axis.init will be removed
+            other = [other] if '..' not in other and ',' not in other else Axis(other)
```
gdementen: same as above

larray/core/axis.py
```diff
@@ -1126,12 +1136,17 @@ def difference(self, other):
...
+        >>> a = Axis('a=a0..a2')
+        >>> a.difference(Axis('a=a1..a3'))
+        Axis(['a0'], 'a')
+        >>> a.difference('a1..a3')
+        Axis(['a0'], 'a')
+        >>> a.difference(['a1', 'a2', 'a3'])
+        Axis(['a0'], 'a')
+        """
+        if isinstance(other, basestring):
+            # TODO : remove [other] if ... when FuturWarning raised in Axis.init will be removed
```
gdementen: same as above

doc/source/changes/version_0_28.rst.inc
```diff
@@ -44,4 +44,13 @@ Miscellaneous improvements
 Fixes
 -----
 
-* fixed something (closes :issue:`1`).
\ No newline at end of file
+* fixed `Axis.union`, `Axis.intersection` and `Axis.difference` when passed value is a single string
+  (closes :issue:`489`):
+
+    >>> a = Axis('a=a0..a2')
+    >>> a.union('a1..a3')
```
gdementen: The examples here are a bit misleading. One could think the problem was only when using .., when it was the case for "normal" strings too. Does not matter much though as nobody is reading the release notes :(

alixdamman: RTFCL (Read The F** Change logs)

gdementen: 😁

PR #559, opened 2017-12-11
fix issue 557 (clean input & ouptut files in unittests)

alixdamman:
* moved input files required by unittests to /test/data

* skip doctests of read_csv and to_csv

* all output files created during unittests are automatically deleted after finishing to run tests


larray/tests/test_array.py
```diff
@@ -88,6 +88,13 @@ def setUp(self):
         self.small_data = np.arange(30).reshape(2, 15)
         self.small = LArray(self.small_data, axes=(self.sex, self.lipro), title=self.small_title)
 
+    @pytest.fixture(autouse=True)
+    def setup(self, tmpdir):
+        self.tmpdir = tmpdir.strpath
+
+    def get_path(self, fname):
```
gdementen: rename get_path to tmp_path or get_tmp_path? Currently, when I see get_path and inputpath, it is not obvious from their name to know which does what, which needs to be used.

PR #561, opened 2017-12-18
fix #437 : allowed to subscript a string LGroup key

### LINKED ISSUE #437: cannot subscript a string LGroup key
Ironically, it works for PGroup:

```python
>>> a = Axis('a=a0,a1')
>>> a.i[0][1:]
'0'
>>> a['a0'][1:]
TypeError: cannot take a subset of a0 because it has a '<class 'str'>' key
>>> # one can workaround this by using str() on the group, but that should not be necessary
>>> str(a['a0'])[1:]
'0'
```

alixdamman:


larray/tests/test_group.py
```diff
@@ -99,6 +99,11 @@ def test_eq(self):
         self.assertEqual(self.list, ['P01', 'P03', 'P04'])
         self.assertEqual(self.list_named, ['P01', 'P03', 'P04'])
 
+    def test_getitem(self):
+        axis = Axis("a=a0,a1")
+        assert axis['a0'][0] == 'a'
+        assert axis['a0'][1] == '0'
```
gdementen: Adding many more tests to test all/most possible combinations would be a good idea I think.

A few examples:
* axis['a0':'a1'][1]
* axis[:][1]
* axis[:][:]
* axis[:][[1, 0]]
* axis[['a0', 'a1', 'a0']][2]
* axis[('a0', 'a1', 'a0')][2]
* axis[LArray(...)][2]

And also add all the same tests for IGroup.

alixdamman:
Could you check only the last commit ?

alixdamman:
thanks

PR #570, opened 2018-02-01
Translated tutorial from  reST to Jupyter notebook + use mybinder (issue 73)

alixdamman:
I finally have something that automatically save the notebook also in yaml format and works on both readthedocs and mybinder:

- `readthedocs` : http://larray-test.readthedocs.io/en/use_mybinder_for_tutorial_73/tutorial.html

- `mybinder` : https://mybinder.org/v2/gh/alixdamman/larray/use_mybinder_for_tutorial_73?filepath=doc%2Fsource%2Ftutorial%2Ftutorial.ipynb

**Some warnings however:**

- the tool `ipyaml` works fine and is really fast but I had difficulties with the part telling to add

    c.NotebookApp.contents_manager_class = 'ipyaml.api.YAMLContentsManager'

  to the Jupyter configuration file `jupyter_notebook_config.py`. Instead I choose to add something similar to [this](https://gist.github.com/masnick/d89c03ba3f0ad48416a5) to my config file `jupyter_notebook_config.py` (see commit "added copy of config file for Jupyter" below). Each time I save the ipynb notebook, its ipyml counterpart is automatically updated.

- Theoretically, it is possible to use the same `environment.yml` file for both `readthedocs` and `mybinder` **but**:
  - `readthedocs` requires to add `python=3.x` as dependency otherwise it crashes
  -  `mybinder` fails if `python=3.x` is set as dependency
  - `mybinder` requires to add `larray` as dependency


alixdamman:
There is absolutely no need to translate tutorial to rst file. Readthedocs and MyBinder only need notebooks. 
By the way, translating a notebook to a rst file take some time while translating it to yaml is immediate. It is a really bad idea to translate notebooks to rst files at each save and it is not required anyway. 

Concerning the merge, you can do it on the "ipyml" notebook and then run `ipyaml <notebook>.ipyml <notebook>.ipynb` (this is very fast!). 

Running Jupyter, I didn't succeeded to load notebooks with .ipyaml extension normally (I see the notebook as a raw yaml file) after updating the `jupyter_notebook_config.py` with:

> c.NotebookApp.contents_manager_class = 'ipyaml.api.YAMLContentsManager'

I guess what you wanted was to work with notebooks stored in yaml format only but I'm not sure mybinder will like it (to do so, mybinder should modify its `jupyter_notebook_config.py` config file before to launch the notebook).   

We can also work with ipynb notebooks and remove all outputs before to push on github as suggested in:
- https://nbsphinx.readthedocs.io/en/0.3.1/usage.html#Using-Notebooks-with-Git
- http://tillahoffmann.github.io/2017/04/17/versioning-jupyter-notebooks-with-git.html
`nbsphinx` automatically run all cells when documentation is built. Clearing all outputs before to push on github would another "trade-off" solution.

alixdamman:
@gdementen For some reason, I had to restart my computer so that the trick `c.NotebookApp.contents_manager_class = 'ipyaml.api.YAMLContentsManager'` works but it works. Sorry.

So we have point 1, 2 and 3. 

I still think we should not include the outputs when pushing to github (outputs produce a lot of noise even in the yaml format).

I will make some tests but I also think `mybinder` cannot work with notebooks in yaml format. If that so, we can use the trick of implementing a 'post-save hook' to automatically save in .ipynb format.

gdementen:
> There is absolutely no need to translate the tutorial notebook to a rst file. Readthedocs and MyBinder only need notebooks.

Oh, ok. I indeed thought it was necessary for readthedocs. Why did you keep tutorial_rest.rst then?

> I guess what you wanted was to work with notebooks stored in yaml format only but I'm not sure mybinder will like it (to do so, mybinder should modify its jupyter_notebook_config.py config file before to launch the notebook).

Indeed, and that is why I wanted to use a post_save_hook to create the .ipynb file

alixdamman:
> Oh, ok. I indeed thought it was necessary for readthedocs. Why did you keep tutorial_rest.rst then?

To compare the previous readthedocs output with the new one. But I should have removed that file earlier.

> Indeed, and that is why I wanted to use a post_save_hook to create the .ipynb file

This how I changed my Jupyter config file and it's working perfectly.

gdementen:
> @gdementen For some reason, I had to restart my computer so that the trick c.NotebookApp.contents_manager_class = 'ipyaml.api.YAMLContentsManager' works but it works. Sorry.
> So we have point 1, 2 and 3.

🎉 

> I still think we should not include the outputs when pushing to github (outputs produce a lot of noise even in the yaml format).

Hmmm. unsure about this. When reviewing, seeing the output is important to tell if the result is "simple" enough to understand.

> I will make some tests but I also think mybinder cannot work with notebooks in yaml format. If that so, we can use the trick of implementing a 'post-save hook' to automatically save in .ipynb format.

Ok. Whatever works. The less manual manipulations we have, the better.

gdementen:
> Concerning the merge, you can do it on the "ipyml" notebook and then run ipyaml <notebook>.ipyml <notebook>.ipynb (this is very fast!).

Yes, I somehow did not think about this option, but if we can have it without the manual step like it seems to be possible from your latest messages, that's even better.

gdementen:
A few more things:
* adding a binder badge in the readme next to |build-status| and |docs| would be nice
* the default view for mybinder notebooks is kinda ugly and probably confusing for users not used to notebooks. Making an interactive slideshow out of the tutorial would make it a lot nicer... but that might not be worth out time, I don't know. https://hub.mybinder.org/user/binder-examples-jupyter-rise-uoou6u7l/notebooks/index.ipynb#/slide-0-0
If that would take too much time, simply hiding the cell toolbar (view -> cell toolbar -> None) by default would already improve the situation, but I don't know if that is possible.

alixdamman:
All the tutorial is in one big notebook. I would like to split it into several ones. I suggest this because I think some sections like `Load/Dump from files`, `Plotting` and `Session` will grow in the future.

gdementen:
Yes, that would be nice. It will be a bit more work (because ideally each one should be readable independently of the others) but I think it would be worth it.

alixdamman:
In this PR or in another? Honestly, I prefer to do it in another one.

gdementen:
Of course, another PR is better. These are all just suggestions anyway. Do what you want with them.

alixdamman:
That makes me think I should use RISE directly last Tuesday instead of Reveal...

alixdamman:
Ready for a first complete review

binder/environment.yml
```diff
@@ -0,0 +1,13 @@
+name: larray-binder
+channels:
+  - damianavila82
```
gdementen: please add a comment why we need this channel. RISE?

doc/source/tutorial/tutorial.ipyml
```diff
@@ -0,0 +1,2918 @@
+cells:
+
+- code: |
+    # @tutorial writers: You can remove cells from the HTML/LaTeX output by adding this to the cell metadata:
```
gdementen: I would be nice if we could have this cell hidden in binder (but ideally silently executed), like in the doc.

PR #573, opened 2018-02-05
Renamed nb_index arg of read_XXX as nb_axes (issue 548)

alixdamman:


larray/util/misc.py
```diff
@@ -631,8 +631,8 @@ def wrapper(*args, **kwargs):
     return wrapper
 
 # deprecate_kwarg is derived from pandas.util._decorators (0.21)
-def deprecate_kwarg(old_arg_name, new_arg_name, mapping=None, stacklevel=2):
-    if not isinstance(mapping, dict):
+def deprecate_kwarg(old_arg_name, new_arg_name, mapping=None, fun=None, stacklevel=2):
```
gdementen: I prefer using func instead of fun. I admit it is less fun 😀 ... but it's less ambiguous and more standard.

doc/source/changes/version_0_28.rst.inc
```diff
@@ -212,6 +212,10 @@ Miscellaneous improvements
      a1   1   1
 
 
+* renamed argument `nb_index` of `read_csv`, `read_excel`, `read_sas`, `from_lists` and `from_string` functions
+  as `nb_axes` (closes :issue:`548`).
```
gdementen: this description does not make it clear enough that you have to change the argument name **and change its value**.

larray/util/misc.py
alixdamman: I started with func but if you look some lines below, you will see : 
```python
     def _deprecate_kwarg(func):
         @wraps(func)
```

gdementen: Didn't see that, but then a more explicit name would both help readability outside the decorator and inside it (help differentiate the two func). Names that come to mind: converter, arg_converter, arg_func, converter_func, ...? I think my preferred is arg_converter.

larray/inout/array.py
```diff
@@ -482,43 +491,45 @@ def read_excel(filepath, sheetname=0, nb_index=None, index_col=None, fill_value=
...
     initialize array from a list of lists (lines)
 
     Parameters
     ----------
     data : sequence (tuple, list, ...)
         Input data. All data is supposed to already have the correct type (e.g. strings are not parsed).
-    nb_index : int, optional
-        Number of leading index columns (ex. 4). Defaults to None, in which case it guesses the number of index columns
+    nb_axes : int, optional
+        Number of axes of the output array. Defaults to None, in which case it guesses the number of axes
```
gdementen: Use the same description than for other functions?

Issue #371, opened 2017-08-25
rename transpose argument of either to_csv or to_excel because they do not mean the same thing


gdementen:
implementing the to_csv version for to_excel would be nice too

PR #576, opened 2018-02-06
replace arg 'transpose' by 'wide' in to_csv and to_excel (issues 371 + 549 + 575)

alixdamman:


larray/core/array.py
```diff
@@ -975,12 +975,14 @@ def to_frame(self, fold_last_axis_name=False, dropna=None):
...
     df = property(to_frame)
 
-    def to_series(self, dropna=False):
+    def to_series(self, name=None, dropna=False):
         """
         Converts LArray into Pandas Series.
 
         Parameters
         ----------
+        name : str, optional
```
gdementen: maybe add a changelog note about this?

gdementen:
this is related to #371 too

larray/core/array.py
```diff
@@ -5816,23 +5849,28 @@ def clip(self, a_min, a_max, out=None):
...
-            transpose = False => no transpose.
-        dialect : 'default' | 'classic'
-            Whether or not to write the last axis name (using '\' )
+        sep : str, optional
+            separator for the csv file. Defaults to `,`.
+        na_rep : str, optional
+            replace NA values with na_rep. Defaults to ''.
+        wide : boolean, optional
+            Whether or not writing arrays in "wide" format. If True, arrays are exported with the last axis
+            represented horizontally. If False, arrays are exported in "narrow" format: all axes and also the values
```
gdementen: "all axes and also the values are written as columns". This formulation is a bit weird => use something like "one column per axis plus one value column" instead?

doc/source/changes/version_0_28.rst.inc
```diff
@@ -234,6 +233,50 @@ Miscellaneous improvements
 
   Closes :issue:`548`:
 
+* replaced argument `transpose` by `wide` in `to_csv` and `to_excel` methods. The new argument `wide` has
```
gdementen: maybe split this in two separate sections: 
* to_csv(transpose -> wide), which does not need any example/explanation because it is a simple rename.
* to_excel(transpose -> wide) which does need an explanation since the behaviour is slightly different

gdementen:
Doh. I didn't say you needed to revert the change. I just wanted to discuss this some more. I am unsure it is worth keeping both arguments as it is likely to confuse users.

As far as I can tell, for 1D arrays, wide=False and transpose=True should give almost the same result (I think the only difference will be the name of the column), for 2D arrays, a.to_excel(transpose=True) should be equivalent to a.T.to_excel() and for >2D arrays, I don't know what transpose=True does :)

alixdamman:
The following file contains examples using flags `wide` and `transpose` for 1D, 2D and 3D arrays: 
[test_excel.xlsx](https://github.com/liam2/larray/files/1726825/test_excel.xlsx)

Some users may find the `transpose` option useful for >=3D arrays. But in that case I personally find the 'pandas' output prettier (using `arr3D.to_frame().transpose().to_excel(...)`) than the 'LArray' output (`arr3D.to_excel(..., transpose=True)`)

larray/core/array.py
```diff
@@ -5923,8 +5969,14 @@ def to_excel(self, filepath=None, sheet_name=None, position='A1', overwrite_file
         header : bool, optional
             Whether or not to write a header (axes names and labels). Defaults to True.
         transpose : bool, optional
-            Whether or not to transpose the resulting array. This can be used, for example, for writing one dimensional
-            arrays vertically. Defaults to False.
+            Whether or not to transpose the resulting array. Defaults to False.
```
gdementen: This probably needs more explanation and/or an example. Maybe also saying that for 2D, it is equivalent to arr.transpose().to_excel()

gdementen:
Thanks for this nice test file. Something like this should be in the "excel" part of the tutorial and MAYBE an abridged version should be in the changelog or at least something explaining the difference between the two arguments (not necessarily with examples)

Issue #237, opened 2017-05-03
implement larray_equiv and larray_nan_equiv


gdementen:
larray_equal is very picky (axes in same order and no broadcasting)

gdementen:
Maybe more arguments to LArray.equals? See #488

alixdamman:
@gdementen In which context that would be useful?

gdementen:
In most contexts. Users usually do not care that axes are ordered differently when comparing two arrays. So we might want to make this the default. Also, currently, you cannot compare an array with a scalar, or to another array with less axes:

```python
>>> arr1 = ndtest(2)
>>> arr2 = full('a=a0,a1;b=b0..b2', arr1)
>>> arr2['a0'].equals(0)
False
>>> arr2.equals(arr1)
False
>>> arr2.equals(arr2.T)
False

but...

>>> arr2 == arr1
a\b    b0    b1    b2
 a0  True  True  True
 a1  True  True  True
>>> (arr2 == arr1).all()
True
>>> arr2 == arr2.T
a\b    b0    b1    b2
 a0  True  True  True
 a1  True  True  True
```

PR #582, opened 2018-02-15
fix #488 : added rtol and atol arguments to LArray.equals

### LINKED ISSUE #488: implement approx_equal or allclose
numpy uses: allclose(a, b, rtol=1e-05, atol=1e-08, equal_nan=False)

alixdamman:
in order to compare float arrays within a relative and/or absolute tolerance.

### DIFF
--- a/doc/source/changes/version_0_28.rst.inc
+++ b/doc/source/changes/version_0_28.rst.inc
@@ -231,7 +231,29 @@ Miscellaneous improvements
 
     >>> arr = read_csv('arr.csv', nb_axes=3)
 
-  Closes :issue:`548`:
+  Closes :issue:`548`.
+
+* added the relative tolerance `rtol` and the absolute tolerance `atol` arguments to the `LArray.equals` method.
+  These two arguments can be used to test the equality between two arrays within a given relative or
+  absolute tolerance:
+
+      >>> arr1 = LArray([6., 8.], "a=a0,a1")
+      >>> arr1
+      a   a0   a1
+         6.0  8.0
+      >>> arr2 = LArray([5.999, 8.001], "a=a0,a1")
+      >>> arr2
+      a     a0     a1
+         5.999  8.001
+      >>> arr1.equals(arr2)
+      False
+      >>> # equals returns True if abs(array1 - array2) <= (atol + rtol * abs(array2))
+      >>> arr1.equals(arr2, atol=0.01)
+      True
+      >>> arr1.equals(arr2, rtol=0.01)
+      True
+
+  Closes :issue:`488`.
 
 * renamed argument `transpose` by `wide` in `to_csv` method.
 
--- a/larray/core/array.py
+++ b/larray/core/array.py
@@ -5225,7 +5225,7 @@ class LArray(ABCLArray):
     def __float__(self):
         return self.data.__float__()
 
-    def equals(self, other, nan_equals=False):
+    def equals(self, other, rtol=0, atol=0, nan_equals=False):
         """
         Compares self with another array and returns True if they have the same axes and elements, False otherwise.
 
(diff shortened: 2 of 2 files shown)


doc/source/changes/version_0_28.rst.inc
```diff
@@ -232,7 +232,28 @@ Miscellaneous improvements
...
+    >>> arr1
+    a      a0      a1
+       0.9999  1.0001
+    >>> arr2 = ones("a=a0,a1")
+    >>> arr2
+    a   a0   a1
+       1.0  1.0
+    >>> arr1.equals(arr2, atol=5e-5)
+    False
+    >>> arr1.equals(arr2, rtol=5e-5)
```
gdementen: the fact that array2 is 1 makes rtol not very interesting.

I would go for something like:
```python
arr1 = LArray([10, 10], "a=a0,a1")
arr2 = LArray([9.99, 10.01], "a=a0,a1")
```

larray/core/array.py
```diff
@@ -5228,6 +5241,24 @@ def equals(self, other, nan_equals=False):
...
+        Return True if absolute(array1 - array2) <= (atol + rtol * absolute(arr2)).
+
+        >>> arr1 = LArray([0.9999, 1.0001], "a=a0,a1")
+        >>> arr1
+        a      a0      a1
+           0.9999  1.0001
+        >>> arr2 = ones("a=a0,a1")
+        >>> arr2
+        a   a0   a1
+           1.0  1.0
```
gdementen: same comments as in the release notes

larray/core/array.py
```diff
@@ -5192,14 +5192,18 @@ def __int__(self):
...
-    def equals(self, other, nan_equals=False):
+    def equals(self, other, rtol=0, atol=0, nan_equals=False):
         """
         Compares self with another array and returns True if they have the same axes and elements, False otherwise.
 
         Parameters
         ----------
         other: LArray-like
             Input array. aslarray() is used on a non-LArray input.
+        rtol : float, optional
```
gdementen: int or float

larray/core/array.py
```diff
@@ -5211,6 +5215,15 @@ def equals(self, other, nan_equals=False):
         bool
             Returns True if self is equal to other.
 
+        Notes
+        -----
+        For finite values, equals uses the following equation to test whether two floating point values are equivalent.
```
gdementen: this is not only valid for floating point values (and it makes sense to use atol or rtol on int arrays too).

larray/core/array.py
```diff
@@ -5228,6 +5241,24 @@ def equals(self, other, nan_equals=False):
         >>> arr1.equals(arr3)
         False
 
+        Test equality of two floating arrays within a given tolerance range.
```
gdementen: not float only

larray/core/array.py
gdementen: s/equivalent/equal/ Otherwise the wording might be ambiguous when we introduce .equiv (#237)

larray/core/array.py
alixdamman: int? why int? 
When users work with arrays containing big integers?

gdementen: Yes. The unit of your arrays could be kg when you only care at the ton level, or whatever.

doc/source/changes/version_0_28.rst.inc
```diff
@@ -232,7 +232,29 @@ Miscellaneous improvements
 
     >>> arr = read_csv('arr.csv', nb_axes=3)
 
-  Closes :issue:`548`:
+  Closes :issue:`548`.
+
+* added the relative tolerance `rtol` and the absolute tolerance `atol` arguments to the `LArray.equals` method.
+  These two arguments can used to test the equality between two float or int arrays within a given relative and
```
gdementen: s/can used/can be used/

Issue #580, opened 2018-02-15
rename to_excel sheet_name to sheet


gdementen:
since sheet_name can also be a sheet index, sheet seems more appropriate than sheet_name.

Avasse:
Hi, i'd be glad to work on this issue, is anybody working on it ?

alixdamman:
@gdementen any objections?

Avasse:
I already read the "How to contribute" http://larray.readthedocs.io/en/stable/contribute.html#sending-your-changes" 

And did some changes to fit that issue. (See : https://github.com/liam2/larray/pull/583) As it is my first PR, telle me if i did anything wrong please. Thanks you !

gdementen:
@alixdamman no, but that assumes you agree with the change...

alixdamman:
@gdementen @Avasse I agree with the change.

gdementen:
closed by de449aa

PR #583, opened 2018-02-16
Fix(#580): Renamed sheet_name to sheet in to_excel

Avasse:
I renamed every "sheet_name" to "sheet" in to_excel def.

### DIFF
--- a/doc/source/changes/version_0_28.rst.inc
+++ b/doc/source/changes/version_0_28.rst.inc
@@ -212,6 +212,29 @@ Miscellaneous improvements
      a1   1   1
 
 
+* renamed argument `nb_index` of `read_csv`, `read_excel`, `read_sas`, `from_lists` and `from_string` functions
+  as `nb_axes`. The relation between `nb_index` and `nb_axes` is given by `nb_axes = nb_index + 1`:
+
+  For a given file 'arr.csv' with content ::
+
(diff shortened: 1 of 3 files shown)


alixdamman:
@Avasse Some unittests failed. Looks like you tried a simple Find/Replace and changed the name of an internal function you shouldn't have ;)

alixdamman:
@Avasse This change also need a changelog (see doc/source/changes/version_0_28.rst.inc), section "Miscellaneous improvements"

gdementen:
A few comments: 
* your proposed change does not work yet, as demonstrated by the failure of Travis CI (the red cross above). You could click on "details" to see why it failed. No worries, but you need to fix this before we can accept the change.
* we cannot simply rename an argument without providing some kind of compatibility layer for our existing users. Please use the deprecate_kwarg decorator. Search in the code for "deprecate_kwarg" for examples on how to use it.
* you need to provide a changelog entry in doc/source/changes/version_0_28.rst.inc
* you need to state somewhere (in the commit message or here) that you accept your contribution to be released under the terms of the GPLv3 license. If you could also include in your message "and all future contributions" that would make things easier if you contribute more in the future.
@alixdamman : we need to update our how to contribute page...

Avasse:
Alright thanks for your replies, i'm on it !

Issue #574, opened 2018-02-05
add 'wide' argument to read_csv/excel (defaults to True)


alixdamman:

alixdamman:
See #549 for previous discussion

gdementen:
The goal is to read data produced using to_csv/excel(wide=False)

alixdamman:
Question: what to do with `Session` objects?

Issue #26, opened 2017-01-04
cleanup unit tests


gdementen:
we should rewrite most LArray unit tests using small-ish arrays created using ndtest() instead of the current demo-related examples.

alixdamman:
cleanup + use a more `pytest` like syntax --> https://docs.pytest.org/en/latest/assert.html

gdementen:
we have 2 separate issues here: the "content" of the tests (ie use smaller test arrays mostly created using ndtest) and the "format" of the tests (ie pytest syntax instead of unittest syntax). We can of course tackle both at the same time, but we do not have to.

alixdamman:
issue #237 should be fixed in the same time.

alixdamman:
to convert nosetests to pytest : https://github.com/pytest-dev/nose2pytest

PR #588, opened 2018-02-21
Wide arg read csv excel (issue #574)

alixdamman:
Documentation for argument `wide` not yet written. Will write documentation when implementation is OK.


alixdamman:
@gdementen add argument `wide` to `Session.save` and `load` (assuming all arrays are stored in wide/narrow format)?

gdementen:
> @gdementen add argument wide to Session.save and load (assuming all arrays are stored in wide/narrow format)?

Would make sense, but that's lower priority IMO => open an issue for this?

gdementen:
btw: using ndtest to create the test files is a good idea and was long overdue. This partly solves #26.

larray/tests/data/test2d.csv
```diff
@@ -1,4 +1,4 @@
-a\b,b0,b1,b2
-a0,0,1,2
-a1,3,4,5
-a2,6,7,8
+a\b,b0,b1
+1,0,1
```
gdementen: My comment about int-like labels was meant for the "horizontal" dimension, which is not handled by pandas but manually parsed by us. Having a test for int-like in a column is interesting too (to make sure Pandas does not change its handling of them) but less critical. In any case, we should have a test explicitly telling in a comment (or test name?) that this is what we test, so that we do not break it accidentally.

larray/tests/data/test1d.csv
```diff
@@ -1,2 +1,2 @@
-time,2007,2010,2013
+a,a0,a1,a2
```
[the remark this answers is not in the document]
gdementen: the from_list test does not count as the year labels already are of the correct type (it uses 1991; not '1991'), and there is no test in from_string doctest with int-like labels in that last axis, and the tests do not check the type of the labels.

larray/tests/data/test2d.csv
alixdamman: If I add a new test array with 'int' labels for all axes, is that OK?

gdementen: sure!

PR #590, opened 2018-02-21
fix #579 : fixed to_excel(transpose=True) on >= 2D arrays

### LINKED ISSUE #579: to_excel(transpose=True) for >=2D must swap the last axes names
This is correct:
```python
>>> ndtest((2, 3)).T.to_excel(None, '2d_to_excel_transpose_true', transpose=True)
b\a  a0  a1
 b0   0   3
 b1   1   4
 b2   2   5
```
This is **wrong** (it should be b\a instead of a\b)
```python
>>> ndtest((2, 3)).to_excel(-1, '2d_transpose_to_excel', transpose=True)
a\b  a0  a1
 b0   0   3
 b1   1   4
 b2   2   5
```

alixdamman:


larray/core/array.py
```diff
@@ -6031,7 +6031,12 @@ def to_excel(self, filepath=None, sheet_name=None, position='A1', overwrite_file
         sheet_name = _translate_sheet_name(sheet_name)
 
         if wide:
-            pd_obj = self.to_frame(fold_last_axis_name=True)
+            if transpose and self.ndim >= 2:
+                axes = self.axes[:-2] + Axis(self.axes[-2].labels, self.axes[-1].name) + \
+                       Axis(self.axes[-1].labels, self.axes[-2].name)
+                pd_obj = self.set_axes(axes).to_frame(fold_last_axis_name=True)
```
gdementen: Wouldn't the implementation be easier if you did it on the pandas object? Something like:

```python
pd_obj = self.to_frame(fold_last_axis_name=True)
if transpose and self.ndim >=2:
    names = df.index.names
    df.index.names = names[:-2] + ['\\'.join(reversed(names[-1].split('\\')))]
```

doc/source/changes/version_0_28.rst.inc
```diff
@@ -351,6 +351,8 @@ Fixes
     >>> arr = ndtest((2, 3))
     >>> arr.to_excel('my_file.xlsx', transpose=True)
 
-    a\b  b0  b1  b2
-    a0    0   1   2
-    a1    3   4   5
+    b\a	 a0	 a1
```
gdementen: there shouldn't be tabs in there.

PR #594, opened 2018-02-26
Element wise equals 593

alixdamman:


larray/core/array.py
```diff
@@ -547,6 +547,92 @@ def _doc_agg_method(func, by=False, long_name='', action_verb='perform', extra_a
...
+    nan_equals: boolean, optional
+        Whether or not to consider nan values at the same positions in the two arrays as equal.
+        By default, an array containing nan values is never equal to another array, even if that other array
+        also contains nan values at the same positions. The reason is that a nan value is different from
+        *anything*, including itself. Defaults to False.
+
+    Returns
+    -------
+    LArray
+        Returns True if the arrays are equal (even in the presence of NaN).
```
gdementen: either remove "(even in the presence of NaN)" or replace it with something vaguely like "(the definition of which depending on the given arguments)". I don't really like this particular wording, but I suppose you get what I mean.

larray/core/array.py
```diff
@@ -547,6 +547,92 @@ def _doc_agg_method(func, by=False, long_name='', action_verb='perform', extra_a
 
 obj_isnan = np.vectorize(lambda x: x != x, otypes=[bool])
 
+
+def equal(a1, a2, rtol=0, atol=0, nan_equals=False):
```
gdementen: I think we should either use equals here or change the equals method to equal. We should have a rule for this (applying s or not) and be consistent. Same thing for nan_equals. We should also have a rule for them too. Not necessarily the same than for function/method names.

larray/core/array.py
```diff
@@ -547,6 +547,92 @@ def _doc_agg_method(func, by=False, long_name='', action_verb='perform', extra_a
...
+        Returns True if the arrays are equal (even in the presence of NaN).
+
+    Notes
+    -----
+    For finite values, nan_equal uses the following equation to test whether two values are equal:
+
+        absolute(array1 - array2) <= (atol + rtol * absolute(array2))
+
+    The above equation is not symmetric in array1 and array2, so that nan_equal(array1, array2) might be different
+    from nan_equal(array2, array1) in some rare cases.
```
gdementen: s/nan_equal/equal/ in both cases. Copy paste is bad 😃

doc/source/changes/version_0_28.rst.inc
```diff
@@ -255,8 +255,20 @@ Miscellaneous improvements
...
+* deprecated `nan_equal` function in favor of `equals` function. The `equals` function has the same optional
+  arguments as the `LArray.equals` method but compares two arrays element-wise and returns array of booleans:
+
+      >>> equals(arr1, arr2)
+      a     a0     a1
+         False  False
+      >>> equals(arr1, arr2, atol=0.01)
+      a    a0    a1
+         True  True
+      >>> equals(arr1, arr2, rtol=0.01)
```
gdementen: given you present the function as a replacement for nan_equal, you should have at least one example with nans and explain or show how to replicate nan_equal behavior

doc/source/changes/version_0_28.rst.inc
```diff
@@ -255,8 +255,20 @@ Miscellaneous improvements
 
   Closes :issue:`488`.
 
-* deprecated `nan_equal` function in favor of `equal` function which has the same optional arguments
-  as the `LArray.equals` method (closes :issue:`593`).
+* deprecated `nan_equal` function in favor of `equals` function. The `equals` function has the same optional
```
gdementen: The current situation with both an equals **function** and an equals **method** which do different things seems very confusing to me. One way to solve this would be to make the equals function private, rename LArray.equals into something like LArray.allequal and add a new method LArray.equals which does what the equals function does. There are other options to fix this though. We might need to rename the corresponding session functions.

doc/source/changes/version_0_28.rst.inc
alixdamman: OK but seriously, has `nan_equal` ever been used at least once by one of our current users? 
Seems like this function was only there to be used in larray-editor project (when comparing arrays/sessions).

doc/source/changes/version_0_28.rst.inc
alixdamman: Once again, I am pretty sure this function will only be used in larray-project. 
IMO, like pandas.DataFrame.equals, LArray.equals should return a boolean value.
I can rename the function 'equals' as `element_equals`?

doc/source/changes/version_0_28.rst.inc
gdementen: Not that I know of, indeed. But then you can not show any example at all. I was just pointing out that your way to presented things was inconsistent.

doc/source/changes/version_0_28.rst.inc
gdementen: yes, that's a good option too.

larray/core/array.py
```diff
@@ -547,48 +547,97 @@ def _doc_agg_method(func, by=False, long_name='', action_verb='perform', extra_a
...
     >>> arr1
     a   a0   a1   a2
-       0.0  nan  2.0
+       6.0  nan  8.0
+
+    Default behavior (same as == operator)
+
     >>> arr2 = arr1.copy()
-    >>> arr1 == arr2
+    >>> element_equals(arr1, arr2)
```
gdementen: use (arr1, arr1) directly?

PR #595, opened 2018-02-28
issue 533 : allow to use 0 padding in axis labels

alixdamman:


larray/util/misc.py
```diff
@@ -515,8 +516,28 @@ def error_handler(error, flag):
...
+    >>> _isintstring('12')
+    True
+    >>> _isintstring('-12')
+    True
+    >>> _isintstring('a1')
+    False
+    >>> _isintstring('01')
+    False
+    """
+    return (s.isdigit() or (len(s) > 1 and s[0] == '-' and s[1:].isdigit())) and \
```
gdementen: This is not very readable IMO and will not work for negative zero padded ints. What do you think of the following?
```python
def isposint(s):
    # avoid zero padded strings
    return s.isdigit() and not (len(s) > 1 and s[0] == '0')
return isposint(s) or (len(s) > 1 and s[0] == '-' and isposint(s[1:]))
```

doc/source/changes/version_0_28.rst.inc
```diff
@@ -320,6 +320,12 @@ Miscellaneous improvements
 
 * Renamed sheet_name of LArray.to_excel to sheet since it can also be an index (closes :issue:580).
 
+* allowed to create an axis with int-like labels starting with one or several 0 (closes :issue:`533`):
```
gdementen: int-like string labels?

larray/util/misc.py
```diff
@@ -515,8 +516,28 @@ def error_handler(error, flag):
     return error_handler
 
 
+_zero_padding_pattern = re.compile('0+[1-9]+')
+
+
 def _isintstring(s):
-    return s.isdigit() or (len(s) > 1 and s[0] == '-' and s[1:].isdigit())
+    """
```
gdementen: explicitly state that zero padded strings are not considered to be int strings?

larray/util/misc.py
alixdamman: Much better indeed. Thanks.

larray/core/group.py
```diff
@@ -394,6 +402,10 @@ def _to_ticks(s, parse_single_int=False):
     ['U']
     >>> list(_to_ticks('..3'))
     [0, 1, 2, 3]
+    >>> _to_ticks('01..12')
+    ['01', '02', '03', '04', '05', '06', '07', '08', '09', '10', '11', '12']
+    >>> _to_ticks('01,02,03,10,11,12')
+    ['01', '02', '03', 10, 11, 12]
```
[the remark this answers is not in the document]
alixdamman: We may make `_to_ticks` return a Numpy array if there is no "Group" object in the list of ticks?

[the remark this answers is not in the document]
gdementen: IIRC the "if" condition is not necessary as in recent versions of larray the **result** of _to_ticks **never** contains Group objects.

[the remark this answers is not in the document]
alixdamman: If I understand correctly, the following piece of code:
```python
            ticks = _to_ticks(labels, parse_single_int=True)
            if _contain_group_ticks(ticks):
                # avoid getting a 2d array if all LGroup have the same length
                labels = np.empty(len(ticks), dtype=object)
                # this does not work if some values have a length (with a valid __len__) and others not
                # labels[:] = ticks
                for i, tick in enumerate(ticks):
                    labels[i] = tick
            else:
                labels = np.asarray(ticks)
```
in `Axis.labels` setter is now useless?

larray/core/group.py
```diff
@@ -382,49 +388,60 @@ def _to_ticks(s, parse_single_int=False):
 
     Notes
     -----
-    This function is only used in Axis.__init__ and union().
+    1) This function is only used in Axis.__init__ and union().
+    2) In Axis.labels setter, the line labels = np.asarray(ticks) returns an Numpy array
```
gdementen: outdated comment

doc/source/changes/version_0_28.rst.inc
```diff
@@ -367,3 +373,13 @@ Fixes
     b0    0   3
     b1    1   4
     b2    2   5
+
+* fixed aggregation on arrays containing zero padded string labels (closes :issue:`522`):
+
+    >>> arr = ndtest(Axis('zero_padding=01,02,03,10,11,12'))
```
gdementen: My comment still stands. Why not, the simpler:

```python
ndtest('zero_padding=01,02,03,10,11,12')
```

larray/core/axis.py
```diff
@@ -197,7 +197,7 @@ def labels(self, labels):
                 for i, tick in enumerate(ticks):
```
gdementen: Have you tried removing the _contain_group_ticks(ticks) code?

larray/util/misc.py
```diff
@@ -7,6 +7,7 @@
 import math
 import itertools
 import sys
+import re
```
gdementen: unused import?

larray/core/axis.py
alixdamman: All tests are green but I am unsure this case (i.e. ticks contain one or several groups) is actually tested somewhere.

gdementen: It was, but the tests have either been removed or commented since this is no longer supported AFAIK

PR #598, opened 2018-03-06
Changelog 0.28 : editor issues

alixdamman:


doc/source/changes/version_0_28.rst.inc
```diff
@@ -98,12 +98,55 @@ New features
...
+
+* allowed to provide an absolute or relative tolerance value when comparing arrays through the `compare` function
+  (closes :editor_issue:`131`).
+
+* made the editor able to detect and display plot objects stored in tuple, list or arrays.
+  For example, arrays of plot objects are returned when using `subplots=True` option in calls of `plot` method:
+
+    >>> a = ndtest('sex=M,F;year=2000..2017')
+    >>> p = a.plot(subplots=True)
+    >>> type(p)
```
gdementen: remove the next 4 lines? our users do not know/care what this is about.

alixdamman: Honestly, my idea was to use the changelog to also teach some little Python tricks to our users but this is maybe pointless and not the right place.

gdementen: I indeed don't think this is the right place for that, but well leave that in if you want, I don't really care 😉

doc/source/changes/version_0_28.rst.inc
```diff
@@ -132,12 +132,11 @@ New features
...
-    >>> p = a.plot(subplots=True)
-    >>> type(p)
-    numpy.ndarray
-    >>> type(p[0])
-    matplotlib.axes._subplots.AxesSubplot
+    >>> a = ndtest('sex=M,F; nat=BE,FO; year=2000..2017')
+    >>> # display 4 plots vertically placed (one plot for each pair (sex, nationality))
+    >>> a.plot(subplots=True)
+    >>> # display 4 plots ordered in a 2 x 2 grid
+    >>> a.plot(subplots=True, layout=(2, 2))
```
gdementen: this is much better IMO, as it is both an interesting feature and related to what was changed

PR #599, opened 2018-03-08
myBinder (0.28)

alixdamman:
The last commit will be dropped before to squash and merge. It is just for testing the mybinder badge.

http://larray-test.readthedocs.io/en/documentation/index.html

### DIFF
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -74,7 +74,7 @@ nbsphinx_allow_errors = True
 
 # This is processed by Jinja2 and inserted before each notebook
 nbsphinx_prolog = r"""
-{% set docname = env.doc2path(env.docname, base='doc') %}
+{% set docname = env.doc2path(env.docname, base='doc/source/') %}
 
 .. only:: html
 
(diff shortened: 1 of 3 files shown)


doc/source/conf.py
```diff
@@ -83,11 +83,9 @@
         
     .. nbinfo::
         Interactive online version:
-        :raw-html:`<a href="https://mybinder.org/v2/gh/liam2/larray/{{ env.config.release }}?
+        :raw-html:`<a href="https://mybinder.org/v2/gh/alixdamman/larray/{{ env.config.release }}?
```
gdementen: ?

doc/source/conf.py
```diff
@@ -115,7 +113,7 @@
 # The short X.Y version.
 version = larray.__version__
 # The full version, including alpha/beta/rc tags.
-release = larray.__version__
+release = 'documentation' #larray.__version__
```
gdementen: Out of curiosity, why do you need this change?

doc/source/conf.py
alixdamman: read the commit message

doc/source/conf.py
alixdamman: read the commit message

doc/source/conf.py
gdementen: Ok, clear now. 

PS: I usually do not read individual commits, unless you somehow hint at it in the PR notes... which I just noticed you actually **did**. I just somehow missed that comment. Sorry about this.

PR #604, opened 2018-03-16
Binder : force version of larray when building myBinder repo (issue 603)

alixdamman:
Result here : https://mybinder.org/v2/gh/alixdamman/larray/binder_postbuild_603

I tried to use a postBuild script a few months ago but didn't work well. 
No, it seems to work really better. 
It seems like myBinder is under fast development right now...

### DIFF
--- a/binder/environment.yml
+++ b/binder/environment.yml
@@ -10,4 +10,5 @@ dependencies:
   - pytables
   - bokeh
   - rise
-  - larray
+  - pip:
+    - larray
--- a/doc/source/changes.rst
+++ b/doc/source/changes.rst
@@ -1,6 +1,14 @@
 ﻿Change log
 ##########
 
+Version 0.29
+============
+
+In development.
+
+.. include:: ./changes/version_0_29.rst.inc
+
+
 Version 0.28
 ============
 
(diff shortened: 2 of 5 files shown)


binder/environment.yml
```diff
@@ -10,4 +10,5 @@ dependencies:
   - pytables
   - bokeh
   - rise
-  - larray=0.28
+  - pip:
+    - larray == 0.28
```
gdementen: Why did you fix the version? Is this necessary? The pypi release should not suffer from the conda-forge lag... And if it turns out to be necessary, we need to update the release script to update this version number automatically, otherwise we will forget to update it.

alixdamman: No, it is not necessary. I just fix the version as an additional security. I will remove this.

Also, I think tutorial notebooks should be exported to a new dedicated repository in larray-project.

gdementen: what would we gain by moving the tutorial to a separate repository?

alixdamman: 1. readthedocs takes a huge time to build the API Reference section. This is very annoying when we update the tutorial and want to check how it is rendered.
2. build a myBinder repo with only tutorial notebooks.
3. Have a cleaner conf.py in larray.

gdementen: On the downside, there is a slightly higher risk of forgetting to update the tutorial when a new version requests it when it is in a separate repository, but the upside seems to be larger.

PR #605, opened 2018-03-20
Make default excel addins available (issue 154) + do not close excel workbook when filepath=-1 (issue 473)

alixdamman:


larray/inout/array.py
```diff
@@ -505,7 +505,7 @@ def read_excel(filepath, sheet=0, nb_axes=None, index_col=None, fill_value=np.na
     Parameters
     ----------
     filepath : str
-        Path where the Excel file has to be read.
+        Path where the Excel file has to be read or use -1 to refer to the last active workbook.
```
gdementen: s/last/currently/ ? I mean I don't want people to think it reopens closed workbooks.

larray/inout/array.py
```diff
@@ -550,9 +550,12 @@ def read_excel(filepath, sheet=0, nb_axes=None, index_col=None, fill_value=np.na
             raise TypeError("'{}' is an invalid keyword argument for this function when using the xlwings backend"
                             .format(list(kwargs.keys())[0]))
         from larray.inout.excel import open_excel
-        with open_excel(filepath) as wb:
-            return wb[sheet].load(index_col=index_col, fill_value=fill_value, sort_rows=sort_rows,
+        wb = open_excel(filepath)
+        res = wb[sheet].load(index_col=index_col, fill_value=fill_value, sort_rows=sort_rows,
                                       sort_columns=sort_columns, wide=wide)
+        if filepath != -1:
+            wb.close()
```
gdementen: explicit close is not exactly the same as the context manager, which also closes when an exception occurs.

larray/inout/array.py
```diff
@@ -550,9 +550,12 @@ def read_excel(filepath, sheet=0, nb_axes=None, index_col=None, fill_value=np.na
             raise TypeError("'{}' is an invalid keyword argument for this function when using the xlwings backend"
                             .format(list(kwargs.keys())[0]))
         from larray.inout.excel import open_excel
-        with open_excel(filepath) as wb:
-            return wb[sheet].load(index_col=index_col, fill_value=fill_value, sort_rows=sort_rows,
+        wb = open_excel(filepath)
+        res = wb[sheet].load(index_col=index_col, fill_value=fill_value, sort_rows=sort_rows,
                                       sort_columns=sort_columns, wide=wide)
```
gdementen: indent is off.

larray/inout/excel.py
```diff
@@ -139,6 +139,13 @@ def __init__(self, filepath=None, overwrite_file=False, visible=None, silent=Non
             if silent is None:
                 silent = not visible
 
+            # activate XLA(M) addins
```
gdementen: I think we should only do this when creating the App instance, not for "active" nor explicit app= given by user (and to a lesser extent, not when reusing an **existing** global app).

larray/inout/array.py
```diff
@@ -505,7 +505,7 @@ def read_excel(filepath, sheet=0, nb_axes=None, index_col=None, fill_value=np.na
     Parameters
     ----------
     filepath : str
-        Path where the Excel file has to be read or use -1 to refer to the last active workbook.
+        Path where the Excel file has to be read or use -1 to refer to the currently active workbook.
```
gdementen: I just noticed open_excel uses the same wording (and possibly LArray.to_excel too), so if we change this, we probably need to change the other(s) too. Unsure it's worth it though.

Issue #611, opened 2018-03-26
rename LArray class to Array?


gdementen:
Pros:
* We are in the larray module, so the L of LArray is kinda redundant
* It would be one less character to type
* The L of LArray is a bit weird for arrays without labels (arrays with only wildcard axes)
* I like it :)

Cons:
* more deprecated warnings for existing users (though a single string replace could fix the problem)
* LGroup will look slightly odd (but having Group and IGroup would be worse). On the other hand, if we converge both kinds of groups to Group (like I discussed in a few issues) this would no longer be a con but a pro.

alixdamman:
Why not. I always found weird to have a specific object with the same as the library. 
But considering users hate syntax change and love stability I would to move this issue for milestone 0.32 or later.

alixdamman:
In that case, I think we should also use Frame instead of LFrame for issue #81, right?

gdementen:
> In that case, I think we should also use Frame instead of LFrame for issue #81, right?

Sure

Issue #608, opened 2018-03-26
Session.summary should include non-LArray objects


alixdamman:
@gdementen should ``Session.summary`` also include Group objects?

gdementen:
I wonder if it's a good idea to exclude other objects...

gdementen:
> @gdementen should Session.summary also include Group objects?

Yes. I think so (I thought I answered this question earlier today, but it seems like not).

PR #612, opened 2018-03-26
issue 578: include Axis and Group objects to Session.save and Session.load methods

alixdamman:
It is normal if Travis crashes. For this issue, I would like to try the TDD approach.


alixdamman:
@gdementen I have implemented something (not definitive) for the HDF format. Could you please take a look (commit by commit) and tell me if I'm going in the right direction before to go further?

larray/core/axis.py
```diff
@@ -1194,6 +1195,34 @@ def align(self, other, join='outer'):
                 other = Axis(other)
             return other
 
+    def to_hdf(self, filepath, key, **kwargs):
```
gdementen: key should be optional. The question is the default value. Defaulting to self.name would be easiest, I guess.

larray/core/group.py
```diff
@@ -1275,6 +1275,39 @@ def containing(self, substring):
             substring = substring.eval()
         return LGroup([v for v in self.eval() if substring in v], axis=self.axis)
 
+    def to_hdf(self, filepath, key, axis_key, **kwargs):
```
gdementen: key and axis_key should be optional. Default to self.name and self.axis.name if present.

larray/core/axis.py
```diff
@@ -1194,6 +1195,34 @@ def align(self, other, join='outer'):
...
+        --------
+        >>> a = Axis("a=a0..a2")
+        >>> a.to_hdf('test.h5', 'a')       # doctest: +SKIP
+
+        Save axis in a specific HDF group
+
+        >>> a.to_hdf('test.h5', 'axes/a')  # doctest: +SKIP
+        """
+        key = _translate_key_hdf(key)
+        s = pd.Series(data=self.labels, name=self.name)
```
gdementen: Can't we go via LArray instead of via Series?

larray/inout/array.py
```diff
@@ -492,8 +498,20 @@ def read_hdf(filepath_or_buffer, key, fill_value=np.nan, na=np.nan, sort_rows=Fa
 
     key = _translate_key_hdf(key)
     df = pd.read_hdf(filepath_or_buffer, key, **kwargs)
```
gdementen: rename df to pd_obj? because AFAICT, it can also be a Series

larray/inout/array.py
```diff
@@ -480,6 +481,11 @@ def read_hdf(filepath_or_buffer, key, fill_value=np.nan, na=np.nan, sort_rows=Fa
     sort_columns : bool, optional
         Whether or not to sort the columns alphabetically (sorting is more efficient than not sorting).
         Defaults to False.
+    type : { 'LArray' | 'Axis' | 'Group' }, optional
+        Type of object to be read. Defaults to 'LArray'.
```
gdementen: It is nice to have this option, but this shouldn't be required for HDF. We should store the type in the HDF as extra metadata.

larray/core/group.py
```diff
@@ -1275,6 +1275,39 @@ def containing(self, substring):
...
+        >>> a.to_hdf('test.h5', 'a')
+        >>> a01 = a['a0,a1'] >> 'a01'
+        >>> a01.to_hdf('test.h5', 'a01', 'a')   # doctest: +SKIP
+
+        Save group in a specific HDF group
+
+        >>> a.to_hdf('test.h5', 'groups/a', 'a')   # doctest: +SKIP
+        """
+        key = _translate_key_hdf(key)
+        s = pd.Series(data=self.eval(), name='{}@{}'.format(self.name, axis_key))
```
gdementen: Ideally we should test that the axis is in the HDF file, but that will probably going to cause an ordering problem (there is currently no guarantee that the group will be saved after the axis), so testing that the correct axis exists in the session seems like a good proxy. It would still be possible to save a group without its axis, but that would be much harder.

larray/core/session.py
```diff
@@ -17,30 +18,36 @@
...
 
     Parameters
     ----------
     args : str or dict of str, array or iterable of tuples (str, array)
-        Name of file to load or dictionary containing couples (name, array).
-    kwargs : dict of str, array
-        List of arrays to add written as 'name'=array, ...
+        Path to the file containing the session to load or
+        list/tuple/dictionary containing couples (name, object).
+    kwargs : dict of str, object
```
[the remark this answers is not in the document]
alixdamman: [pandas](https://python-sprints.github.io/pandas/guide/pandas_docstring.html#parameter-types) says it's ``dict of {str: object}``.

larray/core/session.py
```diff
@@ -983,7 +983,7 @@ def apply(self, func, *args, **kwargs):
 
     def summary(self, template=None):
         """
-        Returns a summary of the content of the session.
+        Returns a summary of the content of the session (arrays only).
```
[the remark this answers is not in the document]
alixdamman: I hope to solve issue #608 in the same release version as this PR. But, in the meantime, I prefer to specify that ``summary`` only shows arrays.

larray/core/axis.py
alixdamman: What would be the added value of using LArray since LArray use pandas internally?

larray/core/session.py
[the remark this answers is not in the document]
gdementen: ok

larray/core/axis.py
gdementen: 1. less reliance on pandas, if we ever want to provide our own serializers
2. different output format. This is both a good thing (arrays and axes use the same format) and a bad thing (the format is more verbose).

alixdamman: 1. Considering the time to develop our own serializers divided by the maximum number of users we can expect, I'm not sure it will ever worth it. pandas makes some sanity checks that we don't need to implement ourselves.   
2. While I'm developing the code to import/export Axis and Group objects, I keep in mind that some day we will also include ``metadata`` to sessions (i.e. objects of basic types like string, int, float, boolean, list and dictionary). So, I'm not sure it is a good idea to go via LArray.

Why not using ``pytables`` directly instead of pandas for non-LArray objects?

gdementen: 1. In general, it is a good idea to use the higher level interface than the lower level one to do the same thing, because it simplifies maintenance. In some cases, the performance difference justifies calling the lower level API directly. Here I consider pandas to be the lower level API. That being said, in this case the different format is more important than the ease of maintenance so going via pandas is probably the better option.
2. you are right but I don't see how it relates to my previous comment.

> Why not using pytables directly instead of pandas for non-LArray objects?

We could but pandas is so prevalent, that we need to be able to read (and possibly write) the pandas dialect of hdf, which has a very specific and a bit odd way to store objects.

larray/inout/array.py
alixdamman: Reminder for myself : http://pandas-docs.github.io/pandas-docs-travis/cookbook.html#hdfstore

larray/core/axis.py
alixdamman: So I will continue to rely on pandas. For metadata, there is a solution: http://pandas-docs.github.io/pandas-docs-travis/cookbook.html#hdfstore (see also https://stackoverflow.com/questions/29129095/save-additional-attributes-in-pandas-dataframe/29130146#29130146)

larray/inout/array.py
```diff
@@ -496,31 +496,40 @@ def read_hdf(filepath_or_buffer, key, fill_value=np.nan, na=np.nan, sort_rows=Fa
...
-        elif _type == 'Group':
-            if name is None:
-                name = str(pd_obj.name)
-            if name == 'None':
-                name = None
-            if 'axis_key' not in dir(attrs):
-                raise ValueError("Attribute 'axis_key' is missing for group {}".format(key))
-            axis = read_hdf(filepath_or_buffer, attrs['axis_key'])
-            res = LGroup(key=pd_obj.values, name=name, axis=axis)
+    if isinstance(filepath_or_buffer, pd.HDFStore):
```
alixdamman: skip this commit during review

larray/core/group.py
```diff
@@ -1300,13 +1302,28 @@ def to_hdf(self, filepath, key=None, axis_key=None, **kwargs):
...
+        Save group with a specific key
+
+        >>> a01.to_hdf('test.h5', 'a_01')           # doctest: +SKIP
 
         Save group in a specific HDF group
 
-        >>> a.to_hdf('test.h5', 'groups/a', 'a')   # doctest: +SKIP
+        >>> a.to_hdf('test.h5', 'groups/a01')       # doctest: +SKIP
+
+        The associated axis is saved at the time with the group if not present in the HDF file yet
```
gdementen: I don't really understand this sentence. I guess you hesitated between "with the group" and "at the same time than the group" and ended up with a mix of both?

larray/inout/array.py
```diff
@@ -497,21 +495,33 @@ def read_hdf(filepath_or_buffer, key, fill_value=np.nan, na=np.nan, sort_rows=Fa
...
+        if _type == 'LArray':
+            res = df_aslarray(pd_obj, sort_rows=sort_rows, sort_columns=sort_columns, fill_value=fill_value, parse_header=False)
+        elif _type == 'Axis':
+            if name is None:
+                name = str(pd_obj.name)
+            if name == 'None':
+                name = None
+            res = Axis(labels=pd_obj.values, name=name)
+            if 'wildcard' not in dir(attrs):
+                raise ValueError("Attribute 'wildcard' is missing for axis {}".format(key))
```
gdementen: Is there a good reason for this exception? Since this case shouldn't happen in normal circumstances, unless the "raw"/original exception of accessing a missing attribute is awfully not understandable, I would let it pass.

larray/inout/array.py
```diff
@@ -497,21 +495,33 @@ def read_hdf(filepath_or_buffer, key, fill_value=np.nan, na=np.nan, sort_rows=Fa
...
-        return None
+    res = None
+    with pd.HDFStore(filepath_or_buffer, **kwargs) as store:
+        pd_obj = store.get(key)
+        attrs = store.get_storer(key).attrs
+        # for backward compatibility but any object read from an hdf file should have an attribute 'type'
+        _type = attrs.type if 'type' in dir(attrs) else 'LArray'
+        if _type == 'LArray':
+            res = df_aslarray(pd_obj, sort_rows=sort_rows, sort_columns=sort_columns, fill_value=fill_value, parse_header=False)
+        elif _type == 'Axis':
```
gdementen: It kinda bothers me that this code is in the inout/**array** module. Not urgent to fix this but still... Here are a few potential solutions:
* defer to some function in a inout/axis module when type == 'Axis'
* rename inout/array to inout/core
* have inout/api with read_hdf which defers to functions in inout/array and inout/axis modules depending on type.

larray/inout/array.py
```diff
@@ -497,21 +495,33 @@ def read_hdf(filepath_or_buffer, key, fill_value=np.nan, na=np.nan, sort_rows=Fa
...
+            if name is None:
+                name = str(pd_obj.name)
+            if name == 'None':
+                name = None
+            res = Axis(labels=pd_obj.values, name=name)
+            if 'wildcard' not in dir(attrs):
+                raise ValueError("Attribute 'wildcard' is missing for axis {}".format(key))
+            res._iswildcard = attrs['wildcard']
+        elif _type == 'Group':
```
gdementen: same as above

larray/inout/array.py
```diff
@@ -497,21 +495,33 @@ def read_hdf(filepath_or_buffer, key, fill_value=np.nan, na=np.nan, sort_rows=Fa
...
+            if name is None:
+                name = str(pd_obj.name)
+            if name == 'None':
+                name = None
+            if 'axis_key' not in dir(attrs):
+                raise ValueError("Attribute 'axis_key' is missing for group {}".format(key))
```
gdementen: same as above

larray/core/array.py
```diff
@@ -6019,7 +6019,9 @@ def to_hdf(self, filepath, key, *args, **kwargs):
         >>> a.to_hdf('test.h5', 'a')  # doctest: +SKIP
         """
         key = _translate_key_hdf(key)
-        self.to_frame().to_hdf(filepath, key, *args, **kwargs)
+        with LHDFStore(filepath) as store:
+            store.put(key, self.to_frame())
+            store.get_storer(key).attrs.type = 'LArray'
```
gdementen: We might want to call this 'Array' instead if we decide to implement #611 later.

larray/util/misc.py
```diff
@@ -732,3 +734,23 @@ def common_type(arrays):
         return np.dtype(('U' if need_unicode else 'S', max_size))
     else:
         return object
+
+
+class LHDFStore(object):
```
gdementen: Why do we need our own HDFStore instead of using the one from pandas directly? To avoid closing the store if we were given an open store?

alixdamman: Because ``HDFStore.__init__`` only accept string as path argument. Both [pandas.pytables.to_hdf](https://github.com/pandas-dev/pandas/blob/v0.22.0/pandas/io/pytables.py#L268) and [pandas.pytables.read_hdf](https://github.com/pandas-dev/pandas/blob/v0.22.0/pandas/io/pytables.py#L286) functions contains an if-else to test if argument ``path_or_buf`` is already an HDFStore instance or not. 

I chose to implement my own context manager ( which is actually based on the beginning of the function [pandas.pytables.read_hdf](https://github.com/pandas-dev/pandas/blob/v0.22.0/pandas/io/pytables.py#L327) )

larray/core/group.py
alixdamman: Bad interactive rebase...

larray/inout/array.py
alixdamman: Let's start with option 2 (rename inout/array to inout/core). It will be easy to move from option 2 to option 1 or 3 later if we changed our mind.

larray/core/array.py
```diff
@@ -6019,7 +6019,9 @@ def to_hdf(self, filepath, key, *args, **kwargs):
         >>> a.to_hdf('test.h5', 'a')  # doctest: +SKIP
         """
         key = _translate_key_hdf(key)
-        self.to_frame().to_hdf(filepath, key, *args, **kwargs)
+        with LHDFStore(filepath) as store:
+            store.put(key, self.to_frame())
+            store.get_storer(key).attrs.type = 'Array'
```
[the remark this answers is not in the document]
gdementen: Also, it somewhat bothers me to have to convert to a pandas object first. I know the LHDFStore is internal for now, but when we make it part of the API, this will need to be changed. We should be able to write something like:

```python
arr = ndtest((2, 3))
with HDFStore(filepath) as store:
    store.put('arr', arr)
```

larray/core/group.py
```diff
@@ -1275,6 +1276,74 @@ def containing(self, substring):
...
+        >>> b = Axis("b=b0..b2")
+        >>> b01 = b['b0,b1'] >> 'b01'
+        >>> # save both the group 'b01' and the associated axis 'b'
+        >>> b01.to_hdf('test.h5')                   # doctest: +SKIP
+        """
+        if key is None:
+            if self.name is None:
+                raise ValueError("Argument key must be provided explicitly in case of anonymous group")
+            key = self.name
+        key = _translate_key_hdf(key)
```
gdementen: only translate if key comes from self.name?

larray/core/group.py
```diff
@@ -1275,6 +1276,74 @@ def containing(self, substring):
...
+        if axis_key is None:
+            if self.axis.name is None:
+                raise ValueError("Argument axis_key must be provided explicitly if the associated axis is anonymous")
+            axis_key = self.axis.name
+        s = pd.Series(data=self.eval(), name=self.name)
+        with LHDFStore(filepath) as store:
+            store.put(key, s)
+            store.get_storer(key).attrs.type = 'Group'
+            if axis_key not in store:
+                self.axis.to_hdf(filepath, key=axis_key)
```
gdementen: use store instead of filepath to avoid opening the file twice?

larray/core/group.py
```diff
@@ -1275,6 +1276,74 @@ def containing(self, substring):
             substring = substring.eval()
         return LGroup([v for v in self.eval() if substring in v], axis=self.axis)
 
+    def to_hdf(self, filepath, key=None, axis_key=None, **kwargs):
```
gdementen: AFAICT, kwargs are not used so should be removed here and in docstring

larray/core/axis.py
```diff
@@ -1194,6 +1195,43 @@ def align(self, other, join='outer'):
                 other = Axis(other)
             return other
 
+    def to_hdf(self, filepath, key=None, **kwargs):
```
gdementen: AFAICT, kwargs are not used so should be removed here and in docstring

larray/core/session.py
```diff
@@ -259,18 +280,30 @@ def save(self, fname, names=None, engine='auto', overwrite=True, display=False,
         display : bool, optional
             Whether or not to display which file is being worked on. Defaults to False.
 
+        Notes
+        -----
+        When session's objects are saved in CSV files, all Axis (Group) objects are saved in the same CSV file
```
gdementen: you could start by saying that array objects are saved one per CSV file/excel sheet.

larray/inout/hdf.py
```diff
@@ -0,0 +1,114 @@
+from __future__ import absolute_import, print_function
+
+import numpy as np
+from pandas import HDFStore
+import warnings
+
+from larray.core.axis import Axis
+from larray.core.group import Group
+from larray.core.array import LArray
+from larray.core.group import _translate_key_hdf, LGroup
```
gdementen: group this line with the one 2 lines above

larray/inout/hdf.py
```diff
@@ -0,0 +1,114 @@
+from __future__ import absolute_import, print_function
+
+import numpy as np
+from pandas import HDFStore
+import warnings
```
gdementen: move this above numpy (python builtin should be before third party)

larray/inout/pandas.py
```diff
@@ -0,0 +1,262 @@
+from __future__ import absolute_import, print_function
+
+import numpy as np
+import pandas as pd
+
+from itertools import product
```
gdementen: move above numpy

larray/inout/string.py
```diff
@@ -0,0 +1,197 @@
...
+from larray.inout.utils import _get_index_col
+from larray.inout.pandas import df_aslarray
+from larray.inout.csv import read_csv
+
+
+__all__ = ['from_lists', 'from_string']
+
+
+@deprecate_kwarg('nb_index', 'nb_axes', arg_converter=lambda x: x + 1)
+def from_lists(data, nb_axes=None, index_col=None, fill_value=np.nan, sort_rows=False, sort_columns=False, wide=True):
```
gdementen: does not really belong in a string module. Maybe rename the module to misc?

larray/inout/utils.py
```diff
@@ -0,0 +1,157 @@
...
+    elif isinstance(index_col, int):
+        index_col = [index_col]
+
+    return index_col
+
+
+_allowed_types = (LArray, Axis, Group)
+
+
+class FileHandler(object):
```
gdementen: not really an utility class (which should be usable in a lot of different contexts). This is a base class, so move this to a module named "base", "core", "abstractbases" or something like that.

larray/tests/test_axis.py
```diff
@@ -1,880 +1,439 @@
...
+    age20 = LGroup('20')
+    age20bis = LGroup('20,')
+    age20ter = LGroup(['20'])
+    age20qua = '20,'
+    age247 = age['2,4,7']
+    age247bis = age[['2', '4', '7']]
+    age359 = age[['3', '5', '9']]
+    age468 = age['4,6,8'] >> 'even'
+    assert 5 in age
+    assert not ('5' in age)
```
gdementen: `'5' not in age` is slightly more readable IMO

larray/tests/test_axis.py
```diff
@@ -1,880 +1,439 @@
...
+    assert age2 in age
+    assert not (age2bis in age)
```
gdementen: same as above: `age2bis not in age`

larray/tests/test_axis.py
```diff
@@ -1,880 +1,439 @@
...
+    assert not (age2ter in age)
```
gdementen: idem... (and for all subsequent lines)

larray/core/array.py
```diff
@@ -68,7 +68,7 @@
 from larray.core.axis import Axis, AxisReference, AxisCollection, X, _make_axis
 from larray.util.misc import (table2str, size2str, basestring, izip, rproduct, ReprString, duplicates,
                               float_error_handler_factory, _isnoneslice, light_product, unique_list, common_type,
-                              renamed_to, deprecate_kwarg)
+                              renamed_to, deprecate_kwarg, LHDFStore)
```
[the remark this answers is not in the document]
alixdamman: I will open a new issue for this. In my mind, LHDFStore was never meant to be public. 
I'm not sure that our users will ever use the HDF format to do something else than save and load Session objects.

larray/inout/string.py
alixdamman: Sure

larray/tests/test_axis.py
alixdamman: I used ``unittest2pytest`` tool. It works pretty well but sometimes lead to some ugly output.

larray/core/array.py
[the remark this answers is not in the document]
gdementen: Good point. I will continue the conversation on the issue.

larray/tests/test_axis.py
gdementen: ok

larray/core/group.py
alixdamman: Isn't the opposite? Only translate if key does not come from self.name?

larray/core/array.py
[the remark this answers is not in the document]
alixdamman: For the record: see #614

larray/core/axis.py
```diff
@@ -1194,6 +1195,43 @@ def align(self, other, join='outer'):
...
+        """
+        Writes axis to a HDF file.
+
+        A HDF file can contain multiple axes.
+        The 'key' parameter is a unique identifier for the axis.
+
+        Parameters
+        ----------
+        filepath : str
+            Path where the hdf file has to be written.
```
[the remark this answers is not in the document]
alixdamman: See #614

larray/core/group.py
gdementen: Oh. My bad, sorry. I thought it translated "raw" strings too. In this case, leave this line as-is, it's fine.

That's the problem with reviews on github, you don't have goto definition. Changing the method name to something like _translate_group_key_hdf to avoid such confusion might be a good idea (or not 😉).

alixdamman:
@gdementen If you don't mind, I think I prefer to make a different PR for each file extension (one for HDF, one for CSV, one for EXCEL and one for pickle)?

gdementen:
@alixdamman sure!

PR #616, opened 2018-04-09
issue 578: include Axis and Group objects to Session.save and Session.load methods

alixdamman:
@gdementen Before to proceed further, could you please review only the last commit ( 041fde494920adbf6b11dcdccd68b429f9563a60 ) ?

### DIFF
--- a/larray/inout/common.py
+++ b/larray/inout/common.py
@@ -3,8 +3,6 @@ from __future__ import absolute_import, print_function
 import os
 from collections import OrderedDict
 
-from larray.core.axis import Axis
-from larray.core.group import Group
 from larray.core.array import LArray
 
 
@@ -21,9 +19,6 @@ def _get_index_col(nb_axes=None, index_col=None, wide=True):
         index_col = [index_col]
 
     return index_col
-
-
-_allowed_types = (LArray, Axis, Group)
 
 
 class FileHandler(object):
@@ -51,16 +46,18 @@ class FileHandler(object):
     def _open_for_write(self):
         raise NotImplementedError()
 
-    def list(self):
+    def list_items(self):
         """
-        Returns the list of objects' names.
+        Return list containing pairs (name, type) for all stored objects
         """
         raise NotImplementedError()
 
-    def _read_item(self, key, *args, **kwargs):
+    def _read_item(self, key, type, *args, **kwargs):
+        """Read item"""
         raise NotImplementedError()
 
-    def _dump(self, key, value, *args, **kwargs):
+    def _dump_item(self, key, value, *args, **kwargs):
+        """Dump item. Raises an TypeError if type not taken into account by the FileHandler subclass."""
         raise NotImplementedError()
 
     def save(self):
@@ -111,14 +108,15 @@ class FileHandler(object):
         display = kwargs.pop('display', False)
         ignore_exceptions = kwargs.pop('ignore_exceptions', False)
         self._open_for_read()
+        key_types = self.list_items()
+        if keys is not None:
+            key_types = [(key, type) for key, type in key_types if key in keys]
         res = OrderedDict()
-        if keys is None:
-            keys = self.list()
-        for key in keys:
+        for key, type in key_types:
             if display:
-                print("loading", key, "...", end=' ')
+                print("loading", type, "object", key, "...", end=' ')
             try:
-                key, item = self._read_item(key, *args, **kwargs)
+                key, item = self._read_item(key, type, *args, **kwargs)
                 res[key] = item
             except Exception:
                 if not ignore_exceptions:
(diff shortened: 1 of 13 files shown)


alixdamman:
@gdementen The argument ``names`` of the method ``Session.load`` allows to define the list of objects (previously arrays) to load. For HDF format, choosing which axes or groups to load is easy. For CSV and Excel, it's gonna be really complicated as I currently save and load all axes and groups at once. 

My question is: Do we limit the argument ``names`` to LArray objects only?

gdementen:
You can load them all and then filter those which are not in names. This is
not super efficient if users have massive axes, but if they use massive
axes, they should use HDF anyway.

larray/inout/common.py
```diff
@@ -112,19 +123,40 @@ def read_items(self, keys, *args, **kwargs):
...
+            axes = self._read_axes(axes_keys, *args, **kwargs)
+            res.update(axes)
+            if display:
+                print("done")
+        # read groups
+        if len(groups_keys) > 0:
+            if display:
+                print("loading axes:", ','.join(groups_keys), "...", end=' ')
+            groups = self._read_groups(groups_keys, *args, **kwargs)
+            res.update(groups)
```
gdementen: a check that there is not a group overwriting an axis might be nice (but I guess its more important to do that during dump).

larray/inout/common.py
```diff
@@ -112,19 +123,40 @@ def read_items(self, keys, *args, **kwargs):
...
+            if display:
+                print("done")
+        # read arrays
+        for key in arrays_keys:
             if display:
                 print("loading", key, "...", end=' ')
             try:
-                key, item = self._read_item(key, *args, **kwargs)
+                key, item = self._read_array(key, *args, **kwargs)
                 res[key] = item
```
gdementen: idem (axes or groups not being overwritten by arrays)

larray/inout/common.py
```diff
@@ -142,17 +174,35 @@ def dump_items(self, key_values, *args, **kwargs):
...
-            if isinstance(value, LArray) and value.ndim == 0:
+        # dump axes
+        axes = [(k, v) for k, v in key_values if isinstance(v, Axis)]
+        if len(axes) > 0:
+            if display:
+                print("dumping axes:", ','.join([t[0] for t in axes]), "...", end=' ')
+            self._dump_axes(axes, *args, **kwargs)
+            if display:
+                print("done")
+        #dump groups
```
gdementen: s/#d/# d/

larray/inout/common.py
```diff
@@ -142,17 +174,35 @@ def dump_items(self, key_values, *args, **kwargs):
...
+        groups = [(k, v) for k, v in key_values if isinstance(v, Group)]
+        if len(groups) > 0:
+            if display:
+                print("dumping groups:", ','.join([t[0] for t in axes]), "...", end=' ')
+            self._dump_groups(groups, *args, **kwargs)
+            if display:
+                print("done")
+        #dump arrays
```
gdementen: space

larray/inout/pandas.py
```diff
@@ -261,3 +262,24 @@ def df_aslarray(df, sort_rows=False, sort_columns=False, raw=False, parse_header
...
+def _group_to_series(group, dtype=None):
+    name = group.name if group.name is not None else '{?}'
+    if group.axis.name is None:
+        raise ValueError("Cannot save a group with an anonymous associated axis")
+    name += '@{}'.format(group.axis.name)
+    return pd.Series(data=group.eval(), name=name, dtype=dtype)
+
+
+def _series_to_group(series, axis):
+    name, axis = series.name.split('@')
```
gdementen: the axis arg is not used. This can be a problem I guess.

larray/inout/pandas.py
```diff
@@ -6,6 +6,7 @@
 import pandas as pd
 
 from larray.core.axis import Axis
+from larray.core.group import LGroup, IGroup
```
gdementen: is IGroup used?

larray/inout/csv.py
```diff
@@ -212,9 +218,33 @@ def __init__(self, fname, overwrite_file=False):
...
 
+    def _to_filepath(self, key):
+        if self.directory is not None:
+            return os.path.join(self.directory, '{}.csv'.format(key))
+        else:
+            return key
+
+    def _load_axes_and_groups(self):
+        # load all axes
+        self.axes = OrderedDict()
```
gdementen: set self.axes and self.groups to None in `__init__`

larray/inout/csv.py
```diff
@@ -223,26 +253,58 @@ def _open_for_write(self):
             except OSError:
                 if not os.path.isdir(self.directory):
                     raise ValueError("Path {} must represent a directory".format(self.directory))
+        self._load_axes_and_groups()
```
gdementen: This shouldn't be necessary (only initializing self.axes and self.groups to OrderedDict should be necessary) and is potentially dangerous as this will (AFAICT) keep old "unused" axes and groups around, which is not what you want in 99% of cases.

gdementen:
LGTM, please squash though

PR #618, opened 2018-04-13
fix #617: Added doctests for read_excel and read_hdf

### LINKED ISSUE #617: Add examples to read_excel() and read_hdf()

alixdamman:
and updated doctests of read_csv


larray/inout/csv.py
```diff
@@ -70,58 +70,133 @@ def read_csv(filepath_or_buffer, nb_axes=None, index_col=None, sep=',', headerse
...
     ...     print(f.read().strip())
-    nat\\sex,M,F
-    BE,0,1
-    FO,2,3
+    a,b\c,c0,c1,c2
+    1,b0,0,1,2
+    1,b1,3,4,5
+    2,b1,9,10,11
+    3,b0,12,13,14
+    >>> # by default, cells associated with missing label combinations are fulfilled with NaN.
```
gdementen: s/fulfilled/filled/

larray/inout/csv.py
```diff
@@ -70,58 +70,133 @@ def read_csv(filepath_or_buffer, nb_axes=None, index_col=None, sep=',', headerse
...
+    >>> # In that case, an int array are converted to a float array.
```
gdementen: s/are/is/

larray/inout/csv.py
```diff
@@ -70,58 +70,133 @@ def read_csv(filepath_or_buffer, nb_axes=None, index_col=None, sep=',', headerse
...
+       a1   3   4   5
+       a2   6   7   8
 
-    Specify the number of axes of the output array (useful when the name of the last axis is implicit)
+    Read array saved in "narrow" format (wide=False)
 
-    >>> a.to_csv(fname, dialect='classic')
+    >>> fname = os.path.join(EXAMPLE_FILES_DIR, 'test3d_narrow.csv')
+    >>> # let's take a look inside the CSV file.
+    >>> # Here, data are stored in a 'narrow' format.
```
gdementen: s/are/is/

larray/inout/csv.py
```diff
@@ -70,58 +70,133 @@ def read_csv(filepath_or_buffer, nb_axes=None, index_col=None, sep=',', headerse
...
-
-    Sort columns
+    a  b\c    c0    c1    c2
+    1   b0   0.0   1.0   2.0
+    1   b1   3.0   4.0   5.0
+    2   b0   nan   nan   nan
+    2   b1   9.0  10.0  11.0
+    3   b0  12.0  13.0  14.0
+    3   b1   nan   nan   nan
+    >>> # using argument 'fill_value', you can choose which value to set in cells
```
gdementen: which value to use to fill missing cells?

larray/inout/csv.py
```diff
@@ -70,58 +70,133 @@ def read_csv(filepath_or_buffer, nb_axes=None, index_col=None, sep=',', headerse
...
+    >>> # associated with missing label .
+    >>> read_csv(fname, fill_value=0)
```
gdementen: the last two example are too long/too big data files to be in the doc.

larray/inout/csv.py
```diff
@@ -70,58 +70,133 @@ def read_csv(filepath_or_buffer, nb_axes=None, index_col=None, sep=',', headerse
...
-    >>> read_csv(fname, wide=False)
-    nat\\sex  M  F
-         BE  0  1
-         FO  2  3
+    a,b0,b1,b2
+    a0,0,1,2
+    a1,3,4,5
+    a2,6,7,8
+    >>> # read the array stored in the CSV file as it
+    >>> read_csv(fname)
```
gdementen: as I told you the other day, I think we should break this case. ie make it return:
```
{0}   a  b0  b1  b2
  0  a0   0   1   2
  1  a1   3   4   5
  2  a2   6   7   8
```
because there is no guarantee the first column is an index

larray/inout/csv.py
```diff
@@ -70,58 +70,133 @@ def read_csv(filepath_or_buffer, nb_axes=None, index_col=None, sep=',', headerse
...
+    a\{1}  b0  b1  b2
+       a0   0   1   2
+       a1   3   4   5
+       a2   6   7   8
+    >>> # using argument 'nb_axes', you can force the number of axes of the output array
+    >>> read_csv(fname, nb_axes=2)
```
gdementen: you should show an example which actually needs the argument. Here it does the same thing as if you did not specify the argument.

larray/inout/csv.py
```diff
@@ -70,58 +70,133 @@ def read_csv(filepath_or_buffer, nb_axes=None, index_col=None, sep=',', headerse
...
     >>> with open(fname) as f:
     ...     print(f.read().strip())
-    nat,M,F
-    BE,0,1
-    FO,2,3
-    >>> read_csv(fname, nb_axes=2)
-    nat\\{1}  M  F
-         BE  0  1
-         FO  2  3
+    a,b,c,value
```
gdementen: Arg. This is way too long for a docstring.

larray/inout/csv.py
```diff
@@ -70,58 +70,133 @@ def read_csv(filepath_or_buffer, nb_axes=None, index_col=None, sep=',', headerse
...
+    >>> read_csv(fname, wide=False)
+    a  b\c  c0  c1  c2
+    1   b0   0   1   2
+    1   b1   3   4   5
+    2   b0   6   7   8
+    2   b1   9  10  11
+    3   b0  12  13  14
+    3   b1  15  16  17
+
+    Sort rows and columns
```
gdementen: Since the argument is self-explanatory in this case, I don't think it's worth to add examples for it.

larray/inout/excel.py
```diff
@@ -56,6 +55,138 @@ def read_excel(filepath, sheet=0, nb_axes=None, index_col=None, fill_value=np.na
         Engine to use to read the Excel file. If None (default), it will use 'xlwings' by default if the module is
         installed and relies on Pandas default reader otherwise.
     **kwargs
+
+    Returns
+    -------
+    LArray
+
+    Examples
```
gdementen: same comments than for read_csv

alixdamman:
I can understand that some examples are too long. 
My first wish was to get rid of the line ``tmpdir = getfixture('tmpdir')`` which is totally obscure for any users and so to use existing files in /tests/data. Do we agree on that?

I really don't want to write fake tests using ``# doctest: +SKIP`` because this is the kind of examples that we always forgot to update when we update functions/methods.

I suggest to add shorter test files in tests/data/ if you don't mind?

doc/source/changes/version_0_29.rst.inc
```diff
@@ -43,6 +43,8 @@ Miscellaneous improvements
     >>> s2
     Session(arr1, arr2, a, b, a01)
 
+* added examples for `read_excel` and `read_hdf` function (closes :issue:`617`).
```
gdementen: functions?

larray/inout/csv.py
```diff
@@ -70,58 +70,91 @@ def read_csv(filepath_or_buffer, nb_axes=None, index_col=None, sep=',', headerse
...
     >>> with open(fname) as f:
     ...     print(f.read().strip())
-    nat,M,F
-    BE,0,1
-    FO,2,3
+    a,b0,b1,b2
+    a0,0,1,2
+    a1,3,4,5
+    a2,6,7,8
+    >>> # read the array stored in the CSV file as it
```
gdementen: as is?

larray/inout/excel.py
```diff
@@ -56,6 +55,96 @@ def read_excel(filepath, sheet=0, nb_axes=None, index_col=None, fill_value=np.na
...
+
+    a  b\c  c0  c1  c2
+    1  b0   0   1   2
+    1  b1   3   4   5
+    2  b1   9   10  11
+    3  b0   12  13  14
+
+    >>> # by default, cells associated with missing label combinations are filled with NaN.
+    >>> # In that case, an int array is converted to a float array.
+    >>> read_excel(fname, 'missing_values')
```
gdementen: I **_wonder_** if the combination of this sheet name and the fact the argument ("sheet") is used positionally is not potentially confusing. Maybe using sheet='missing_values', or renaming the sheet to something like 'with_missings' would make it more obvious.

The same question applies to HDF array names.

alixdamman:
The last version of pytables, i.e. 3.4.3, makes Travis crashing with Python 2.7

PR #624, opened 2018-04-25
fix #623 : added app property to Workbook class (shortcut for self.xw_wkb.app)

### LINKED ISSUE #623: Add self.app to Workbook

alixdamman:


alixdamman:
to add a docstring to what? (considering I remove the app property since it is not necessary)

gdementen:
...add a docstring... in the case you keep the property.

larray/inout/xw_excel.py
```diff
@@ -661,6 +664,10 @@ def open_excel(filepath=None, overwrite_file=False, visible=None, silent=None, a
 >>> wb['arr3'] = arr3.dump()                                      # doctest: +SKIP
 >>> wb.save()                                                     # doctest: +SKIP
 >>> wb.close()                                                    # doctest: +SKIP
+"""
+
+    Workbook.app.__doc__ = """
+Return the currently active Excel instance.
```
gdementen: This is wrong/misleading. It should rather be something along the lines of: `Return the Excel instance used by this workbook` or  `Return the Excel instance this workbook is attached to`.

PR #626, opened 2018-04-25
fix #608 : include non-LArray objects in Session.summary

### LINKED ISSUE #608: Session.summary should include non-LArray objects

### LINKED ISSUE #622: Argument label in LArray.append and LArray.prepend is not optional in some cases
This fails: 
```python
   a = ones('nat=BE,FO;sex=M,F')
   a.append(X.sex, a.sum(X.sex))
```

alixdamman:
need additional changelog

### DIFF
--- a/larray/core/session.py
+++ b/larray/core/session.py
@@ -1091,7 +1091,7 @@ class Session(object):
 
         >>> print(s.summary())  # doctest: +NORMALIZE_WHITESPACE
         axis1: a ['a0' 'a1' 'a2'] (3)
-        group1: a01@a ['a0', 'a1'] (2)
+        group1: a['a0', 'a1'] >> a01 (2)
         arr1: a, b (2 x 2) [int64]
             array 1
         arr2: a (4) [int64]
@@ -1107,11 +1107,11 @@ class Session(object):
         ...     return "{} -> {} ({})\\n  title = {}\\n  dtype = {}".format(key, axes_names, shape,
         ...                                                                 array.title, array.dtype)
         >>> template = {Axis:  "{key} -> {name} [{labels}] ({length})",
-        ...             Group: "{key} -> {name}:{axis_name} {labels} ({length})",
+        ...             Group: "{key} -> {name}: {axis_name}{labels} ({length})",
         ...             LArray: print_array}
         >>> print(s.summary(template))
         axis1 -> a ['a0' 'a1' 'a2'] (3)
-        group1 -> a01:a ['a0', 'a1'] (2)
+        group1 -> a01: a['a0', 'a1'] (2)
         arr1 -> a, b (2 x 2)
           title = array 1
           dtype = int64
@@ -1127,7 +1127,7 @@ class Session(object):
         if Axis not in template:
             template[Axis] = "{key}: {name} [{labels}] ({length})"
         if Group not in template:
-            template[Group] = "{key}: {name}@{axis_name} {labels} ({length})"
+            template[Group] = "{key}: {axis_name}{labels} >> {name} ({length})"
         if LArray not in template:
             template[LArray] = "{key}: {axes_names} ({shape}) [{dtype}]\n    {title}"
 
--- a/larray/core/session.py
+++ b/larray/core/session.py
@@ -929,7 +929,7 @@ class Session(object):
         >>> arr2 = ndtest((2, 2))
         >>> sess = Session([('arr1', arr1), ('arr2', arr2)])
         >>> def print_summary(s):
-        ...     print(s.summary({LArray: "{name} -> {axes_names}"}))
+        ...     print(s.summary({LArray: "{key} -> {axes_names}"}))
         >>> print_summary(sess)
         arr1 -> a, b, c
         arr2 -> a, b
@@ -1066,10 +1066,10 @@ class Session(object):
             The string provided for a given type must include specific arguments written inside brackets {}.
             Available arguments are:
 
-                - for LArray: 'name', 'group_name', 'axis', 'labels' and 'length' for groups,
-                - for Axis: 'name', 'axis_name', 'labels' and 'length' for axes and
-                - for Group: 'name', 'axes_names', 'shape', 'dtype' and 'title' for arrays.
-                - for all other types: 'name', 'value'
+                - for LArray: 'key', 'name', 'axis_name', 'labels' and 'length' for groups,
+                - for Axis: 'key', 'name', 'labels' and 'length' for axes and
+                - for Group: 'key', 'axes_names', 'shape', 'dtype' and 'title' for arrays.
+                - for all other types: 'key', 'value'
 
         Returns
         -------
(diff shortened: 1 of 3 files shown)


larray/core/session.py
```diff
@@ -1061,40 +1061,84 @@ def summary(self, template=None):
 
         Parameters
         ----------
-        template: str
+        template: dict {str: str}
             Template describing how items are summarized (see examples).
-            Available arguments are 'name', 'axes_names' and 'title'
+            Accepted keys are 'axis', 'group' and 'array'.
```
gdementen: wouldn't it be better to use directly the type as key instead of string, so that it is future proof (and users can add their own objects in the session and use summary)? eg:
print(s.summary({LArray: "{name} -> {axes_names}"}))

larray/core/session.py
```diff
@@ -1061,40 +1061,84 @@ def summary(self, template=None):
...
+        if 'axis' not in template:
+            template['axis'] = "{name}: {axis_name} [{labels}] ({length})"
+        if 'group' not in template:
+            template["group"] = "{name}: {group_name}@{axis} {labels} ({length})"
+        if 'array' not in template:
+            template["array"] = "{name}: {axes_names} ({shape})\n    {title}\n    {dtype}"
+
+        def kind_kwargs(k, v):
+            if isinstance(v, Axis):
+                return ('axis', {'name': k, 'axis_name': v.name, 'labels': v.labels_summary(), 'length': len(v)})
```
gdementen: I think it would be better if there was a tighter correspondence between the template keyword and the object attribute, so I think we should:
s/'name'/'key'/
s/'axis_name'/'name'/
s/'labels'/'labels_summary'/   <-- maybe not this one?
s/'group_name'/'name'/
s/'axis'/'axis_name'/
In fact we might want to make the correspondence automatic (ie use getattr), so that a user could provide a template (using attributes only) for a type we do not know about.

larray/core/session.py
```diff
@@ -1061,40 +1061,84 @@ def summary(self, template=None):
...
+            elif isinstance(v, Group):
+                return ('group', {'name': k, 'group_name': v.name, 'axis': v.axis.name, 'labels': v.key,
+                                  'length': len(v)})
+            elif isinstance(v, LArray):
+                return ('array', {'name': k, 'axes_names': ', '.join(v.axes.display_names),
+                                  'shape': ' x '.join(str(i) for i in v.shape), 'title': v.title, 'dtype': v.dtype})
+            else:
+                return None
```
gdementen: I think supporting "any kind of object" by using repr(v) (or str(v)) for them would be better.

larray/core/session.py
```diff
@@ -1061,40 +1061,84 @@ def summary(self, template=None):
...
-                         'title': v.title}
-                        for k, v in self.items() if isinstance(v, LArray)]
-        return '\n'.join(template.format(**kwargs) for kwargs in templ_kwargs)
+            template = {}
+        if 'axis' not in template:
+            template['axis'] = "{name}: {axis_name} [{labels}] ({length})"
+        if 'group' not in template:
+            template["group"] = "{name}: {group_name}@{axis} {labels} ({length})"
+        if 'array' not in template:
+            template["array"] = "{name}: {axes_names} ({shape})\n    {title}\n    {dtype}"
```
gdementen: I think the default template for arrays should have the dtype on the first line (a bit like in the editor title)

gdementen:
Also, I think that at some point, users will want to display array (and other objects) arbitrary metadata, so we need some way to let them access that without specifying all possible keywords. On the other hand I am a bit nervous to allow arbitrary code execution in templates (like in f-strings). Maybe access to arbitrary-deep attributes without allowing `__call__` would be an acceptable compromise. I don't think this would be a security concern.

larray/core/session.py
alixdamman: If you want to allow users to display any attribute and especially `LArray.data`, the method shouldn't be called "summary". By the way, I'm not sure it is such a good idea to have a very generic `summary` method.

alixdamman: Another solution would be to let the choice between passing a string template or a user defined function returning a string representation of the object (like LArray.info).

larray/core/session.py
```diff
@@ -1057,16 +1057,19 @@ def apply(self, func, *args, **kwargs):
...
+        template: dict {object type: str}
             Template describing how items are summarized (see examples).
-            Accepted keys are 'axis', 'group' and 'array'.
-            Available arguments are 'name', 'group_name', 'axis', 'labels' and 'length' for groups,
-            'name', 'axis_name', 'labels' and 'length' for axes and
-            'name', 'axes_names', 'shape', 'dtype' and 'title' for arrays.
+            The string provided for a given type must include specific arguments written inside brackets {}.
+            Available arguments are:
+
+                - for LArray: 'key', 'name', 'axis_name', 'labels' and 'length' for groups,
```
gdementen: for LArray ... for groups
for Axis ... for axes
for Group ... for arrays???

larray/core/session.py
```diff
@@ -1117,24 +1117,24 @@ def summary(self, template=None):
...
-                return ('group', {'name': k, 'group_name': v.name, 'axis': v.axis.name, 'labels': v.key,
+                return (Group, {'key': k, 'name': v.name, 'axis_name': v.axis.name, 'labels': v.key,
                                   'length': len(v)})
             elif isinstance(v, LArray):
-                return ('array', {'name': k, 'axes_names': ', '.join(v.axes.display_names),
+                return (LArray, {'key': k, 'axes_names': ', '.join(v.axes.display_names),
                                   'shape': ' x '.join(str(i) for i in v.shape), 'title': v.title, 'dtype': v.dtype})
             else:
-                return None
+                return (type(v),{'key': k, 'value': str(v)})
```
gdementen: add space after comma

gdementen:
Indeed, this is a much better solution that what I suggested. This can come
later though since it will be backward compatible.

gdementen:
Displaying arbitrary attributes in a summary does not bother me. The point
is that we don't know what kind of objects users will put into the session
and which attribute(s) are most important to display in a summary.

But you are right that what I proposed was not a great idea. Specifying a
callable to return a string is much better. I just didn't think about that
solution.

larray/core/session.py
```diff
@@ -1057,44 +1057,88 @@ def apply(self, func, *args, **kwargs):
...
+                return (Group, {'key': k, 'name': v.name, 'axis_name': v.axis.name, 'labels': v.key,
+                                  'length': len(v)})
+            elif isinstance(v, LArray):
+                return (LArray, {'key': k, 'axes_names': ', '.join(v.axes.display_names),
+                                  'shape': ' x '.join(str(i) for i in v.shape), 'title': v.title, 'dtype': v.dtype})
+            else:
+                return (type(v),{'key': k, 'value': str(v)})
+
+        templ_kwargs = [kind_kwargs(k, v) for k, v in self.items()]
+        templ_kwargs = [k for k in templ_kwargs if k is not None]
```
gdementen: this line is now useless I think

larray/core/session.py
```diff
@@ -1124,21 +1131,26 @@ def summary(self, template=None):
...
-            elif isinstance(v, Group):
-                return (Group, {'key': k, 'name': v.name, 'axis_name': v.axis.name, 'labels': v.key,
-                                  'length': len(v)})
-            elif isinstance(v, LArray):
-                return (LArray, {'key': k, 'axes_names': ', '.join(v.axes.display_names),
-                                  'shape': ' x '.join(str(i) for i in v.shape), 'title': v.title, 'dtype': v.dtype})
+        def display(k, v):
+            t = Group if isinstance(v, Group) else type(v)
+            tmpl = template.get(t, None)
+            if not (isinstance(tmpl, basestring) or callable(tmpl)):
```
gdementen: Wouldn't using a default template for unknown types: "{key}: {value}" be more useful?

Issue #154, opened 2017-03-16
make default Excel addins available when using array.to_excel() / Ctrl-E in the viewer


gdementen:
This can at least be solved for a specified Excel addin:
http://stackoverflow.com/questions/22169877/load-excel-add-in-using-win32com-from-python

not sure we can do it for all "default" addins.

If that is not possible, we could have a global config for larray where the user could setup this kind of things (like pd.options).

gdementen:
https://mail.python.org/pipermail/python-win32/2007-April/005757.html
might help

gdementen:
This fails for ctrl-E in the viewer. I have a fix for it though.

gdementen:
closed by larray-project/larray-editor#149

PR #627, opened 2018-04-25
loadaddins

gdementen:


gdementen:
These are related but different issues, I will make a separate PR tomorrow to fix #154

PR #628, opened 2018-04-25
fix #610 : included test of Group and Axis objects in Session.equals

### LINKED ISSUE #610: update Session.equals to also test Axis and Group objects
@gdementen also add ``Session.axis_equals`` and ``Session.group_equals``?

alixdamman:
+ renamed Session.array_equals to Session.element_equals

### DIFF
--- a/larray/core/group.py
+++ b/larray/core/group.py
@@ -1055,6 +1055,69 @@ class Group(object):
     __ne__ = _binop('ne')
     __eq__ = _binop('eq')
 
+    def equals(self, other):
+        """
+        Checks if this group is equal to another group.
+        Two groups are equal if they have the same group and axis names and correspond to the same labels.
+
+        Parameters
+        ----------
+        other : Group
+            Group to compare with.
+
+        Returns
+        -------
+        bool
+            True if the other group is equal to this group, False otherwise.
+
+        Examples
+        --------
+        >>> from larray import Axis
+        >>> a = Axis('a=a0..a3')
+        >>> a02 = a['a0:a2'] >> 'group_a'
+
+        Same group names, axis names and labels
+
+        >>> a02.equals(a02)
+        True
+
+        Different group names (one is None)
+
+        >>> a02.equals(a['a0:a2'])
+        False
+
+        Different axis name
+
+        >>> other_axis = a.rename('other_name')
+        >>> a02.equals(other_axis['a0:a2'] >> 'group_a')
+        False
+
+        Different labels
+
+        >>> a02.equals(a['a1:a3'] >> 'group_a')
+        False
+
+        Mixing slice and list groups
+
+        >>> a['a0:a2'].equals(a['a0,a1,a2'])
+        True
+
+        Mixing LGroup and IGroup
+
+        >>> a['a0:a2'].equals(a.i[0:3])
+        True
+        """
+        if self is other:
+            return True
+        if self.name != other.name:
+            return False
+        self_axis_name = self.axis.name if isinstance(self.axis, ABCAxis) else self.axis
+        other_axis_name = other.axis.name if isinstance(other.axis, ABCAxis) else other.axis
+        if self_axis_name != other_axis_name:
+            return False
+        res = self == other
+        return res if isinstance(res, bool) else all(res)
+
     def set(self):
         """Creates LSet from this group
 
--- a/larray/core/session.py
+++ b/larray/core/session.py
@@ -800,10 +800,11 @@ class Session(object):
     __abs__ = _unaryop('abs')
     __invert__ = _unaryop('invert')
 
-    def array_equals(self, other):
-        """Test if arrays of the current session are equal to those of another session.
+    def element_equals(self, other):
+        """Test if groups, axes and arrays of the current session are equal to those of another session.
 
-        Equivalent to apply :py:meth:`LArray.equals` with flag nan_equals=True to all arrays from two sessions.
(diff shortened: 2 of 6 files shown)


larray/core/session.py
```diff
@@ -820,44 +821,56 @@ def array_equals(self, other):
...
+        name      a   a01   arr1   arr2   arr3
+              False  True  False  False  False
         """
-        self_keys = set(self.keys())
-        all_keys = list(self.keys()) + [n for n in other.keys() if n not in self_keys]
-        def larray_nan_equal(a1, a2):
-            try:
-                a1 = aslarray(a1)
-            except Exception:
+        self_keys =[k for k, v in self.items() if isinstance(v, (Axis, Group, LArray))]
```
gdementen: s/=[/= [/

gdementen: use something like: `supported_objects = (Axis, Group, LArray)`?

larray/core/session.py
```diff
@@ -820,44 +821,56 @@ def array_equals(self, other):
...
+        other_keys = [k for k, v in other.items() if isinstance(v, (Axis, Group, LArray)) and k not in self_keys]
+        all_keys = self_keys + other_keys
+
+        def elem_equal(e1, e2):
+            if not type(e1) is type(e2):
```
gdementen: `x is not y` is more readable IMO than `not x is y`

larray/core/session.py
```diff
@@ -820,44 +821,56 @@ def array_equals(self, other):
...
                 return False
-            return a1.equals(a2, nan_equals=True)
-        res = [larray_nan_equal(self.get(key), other.get(key)) for key in all_keys]
+            if isinstance(e1, Group):
+                return e1 == e2
```
gdementen: This is wrong if the group keys are arrays:

```python
>>> a = Axis('a=a0..a2')
>>> a['a0,a2'] == a['a1,a2']
False
>>> a[np.array(['a0', 'a2'])] == a[np.array(['a1', 'a2'])]
array([False,  True], dtype=bool)
```
and besides, I wonder if we should not go beyond == for groups, because == does not test .name. 

Maybe we should implement Group.equals which would solve both problems at the same time.

alixdamman: Seems a good idea. Do I open a new issue for that?

alixdamman:
@gdementen in `array.py`, we have defined the function `element_equal` without 's' at the end. So, do I rename `element_equal` as `element_equals` in array.py or do I rename `Session.element_equals` as `Session.element_equal`?

larray/core/session.py
gdementen: do as you prefer.

alixdamman: I will add a changelog for Group.equals but no new issue.

gdementen:
> @gdementen in array.py, we have defined the function element_equal without 's' at the end. So, do I rename element_equal as element_equals in array.py or do I rename Session.element_equals as Session.element_equal?

I am unsure we should rename anything yet. Open an issue for this (that we need to resolve before 0.29). There is also the question of the nan_equals argument of element_equal. We should find a rule when to apply s or no s and stick to it everywhere.

Grammatically, **I think** we can justify both:
* element_equal => each element is equal to the corresponding element in the other object
* element_equals => each element equals the corresponding element in the other object

larray/core/group.py
```diff
@@ -1055,6 +1055,74 @@ def opmethod(self, other):
     __ne__ = _binop('ne')
     __eq__ = _binop('eq')
 
+    # TODO : test also group.axis?
```
gdementen: We should probably have **some** explicit check but I don't know how far we should go. There is already some kind of indirect/implicit check because you use == which uses .eval() which ensures the labels produced are the same (even if the axes are different).

I think the goal of .equals is to check whether or not two groups would return the same values if used on an array (both getitem and aggregates). For that to be true in the presence of ambiguous labels, we need to check the axis name is the same in both groups, or, if we have arrays with several axes with the same name (as is common when using matrices) we have to check for that the axis object is the same. But this pickiness might be annoying in other cases, so I don't really know what's best.

larray/core/group.py
```diff
@@ -1055,6 +1055,74 @@ def opmethod(self, other):
...
+    def equals(self, other):
+        """
+        Checks if self is equal to another group.
```
gdementen: s/self/this group/

larray/core/group.py
```diff
@@ -1055,6 +1055,74 @@ def opmethod(self, other):
...
+        Two groups are equal if they have the same name and key.
```
gdementen: Not really true. As you implemented it, two groups can be equal with different keys (as in your example below mixing IGroup & LGroup) and can be different with the same (slice) key (if the axes differ)

larray/core/group.py
alixdamman: we may add an argument `test_axis=True` to Group.equals?

gdementen: I think that by default we should check the name of the axis (if any) but not the object because it would be too confusing if we have:
```python
>>> a = Axis('a=a0..a2')
>>> a2 = Axis('a=a0..a2')
>>> a.equals(a2)
True
>>> # but
>>> a['a0,a2'].equals(a2['a0,a2'])
False
```
because if you don't check the ids of the objects, you don't see that the axes are different.

Unsure having an argument to check that the axis objects are actually the same is worth it. Let's avoid it for now, if that's ok with you.

alixdamman: OK

PR #637, opened 2018-05-31
issue 26 (cleanup unit tests): rewritten test_array.py: unittest -> pytest

alixdamman:


alixdamman:
Yes, that's why the commit message doesn't start with fix #26.

PR #641, opened 2018-06-05
Add metadata to array objects

alixdamman:
See comment https://github.com/larray-project/larray/issues/78#issuecomment-383603416 for specifications

### DIFF
--- a/doc/source/changes/version_0_29.rst.inc
+++ b/doc/source/changes/version_0_29.rst.inc
@@ -29,6 +29,43 @@ New features
     >>> a02.equals(a['a1:a3'] >> 'group_a')
     False
 
+* allowed arrays to have metadata (e.g. title, description, authors, ...).
+  Metadata can be accessed using the syntax ``array.meta.name``.
+  See bellow for use cases:
+
+    >>> # add metadata to/from an array
+    >>> # for Python <= 3.5
+    >>> arr = ndtest((3, 3), meta=[('title', 'array for testing'), ('author', 'John Smith')])
+    >>> # for Python >= 3.6
+    >>> arr = ndtest((3, 3), meta=Metadata(title='array for testing', author='John Smith'))
+    >>> # access an item
+    >>> arr.meta.author
+    'John Smith'
+    >>> # update/add an item (the item is added if not found in existing metadata)
+    >>> arr.meta.city = 'London'
+    >>> # delete an item
+    >>> del arr.meta.city
+
+  Warnings:
+
+  - Currently, only the HDF (.h5) file format supports saving and loading metadata.
+  - Metadata are not kept when actions or methods are applied on a array
+    except for operations modifying the object in-place, such as `pop[age < 10] = 0`,
+    and when the method `copy()` is called. Do not add metadata to an array if you know
+    you will apply actions or methods on it before dumping it.
+
+  Closes :issue:`78` and :issue:`79`.
+
+* deprecated attribute `title` of array objects. A title is now considered as a metadata
+  and must be added as:
+
+    >>> # add title at array creation
+    >>> arr = ndtest((3, 3), meta=[('title', 'array for testing')])
+
+    >>> # or after array creation
+    >>> arr = ndtest((3, 3))
+    >>> arr.meta.title = 'array for testing'
+
 
 .. _misc:
 
--- a/larray/core/array.py
+++ b/larray/core/array.py
@@ -778,7 +778,7 @@ class LArray(ABCLArray):
                             "instead of {}".format(type(meta).__name__))
         self._meta = meta if isinstance(meta, Metadata) else Metadata(meta)
 
-    # XXX: rename to posnonzero and implement a label version of nonzero
+    # TODO: rename to posnonzero and implement a label version of nonzero
     def nonzero(self):
         """
         Returns the indices of the elements that are non-zero.
--- a/larray/core/array.py
+++ b/larray/core/array.py
@@ -542,7 +542,6 @@ def _doc_agg_method(func, by=False, long_name='', action_verb='perform', extra_a
         .format(specific=doc_specific, action_verb=action_verb, long_name=long_name)
     parameters = """Parameters
         ----------{args}{varargs}{kwargs}""".format(args=doc_args, varargs=doc_varargs, kwargs=doc_kwargs)
-
     func.__doc__ = func.__doc__.format(signature=signature, parameters=parameters)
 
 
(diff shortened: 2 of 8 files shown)


larray/core/array.py
```diff
@@ -7168,11 +7221,13 @@ def zeros_like(array, title='', dtype=None, order='K'):
     """
     if not title:
         title = array.title
-    return LArray(np.zeros_like(array, dtype, order), array.axes, title)
+    if meta is None:
```
gdementen: I don't think this is a good idea to copy metadata by default in all the *_like functions. At least it does not match current usage. Usually, our users create new arrays to represent a different "concept" than the original array. They just want the same axes. Having an easy option to copy the metadata might be a good idea (e.g. meta='copy'). Unsure it worth it compared to the user doing it explicitly:
```python
>>> a = ndtest((2, 3))
>>> b = zeros_like(a, meta=a.meta)
```
In either case, it would be nice if you mentioned in the docstring whether or not metadata is copied.

larray/tests/test_array.py
```diff
@@ -176,40 +174,53 @@ def small_array(meta):
...
 
 
-def test_ndrange():
-    arr = ndtest('a=a0..a2')
+def test_concat(meta):
+    res = concat((ndtest((2, 3)), ndtest('b=b4,b5')), 'b', meta=meta)
+    assert res.meta == meta
+
+def test_compact(meta):
+    arr = LArray([[1, 2], [1, 2]], [Axis('sex=M,F'), Axis('nat=BE,FO')], meta=meta)
```
gdementen: LArray([[1, 2], [1, 2]], 'sex=M,F;nat=BE,FO') ?

larray/tests/test_array.py
```diff
@@ -176,40 +174,53 @@ def small_array(meta):
 io_narrow_missing_values[2, 'b1', 'c1'] = np.nan
 
 
-def test_ndrange():
-    arr = ndtest('a=a0..a2')
+def test_concat(meta):
```
gdementen: the name of the function is misleading... it seems to be general while it only tests metadata.

doc/source/changes/version_0_29.rst.inc
```diff
@@ -29,6 +29,32 @@ New features
...
+
+    >>> # update/add an item (the item is added if not found in existing metadata)
+    >>> arr.meta.city = 'London'
+    >>> # or
+    >>> arr.meta['city'] = 'London'
+
+    >>> # delete an item
+    >>> del arr.meta.city
+    >>> # or
+    >>> del arr.meta['city']
```
gdementen: 1. More important than showing the two different syntaxes for modifying/deleting metadata would be to show metadata propagation and explain the rule used to decide whether a method propagates metadata or not. To avoid surprises, it is **extremely** important that we have a simple rule for this and that users understand it. I think the rule should be something like: all operations which only modify the axes (including reordering, subset, etc.) preserve metadata, all operations which modify the values drop metadata. I already see one notable exception: `__setitem__` should not drop metadata since we modify the object in-place.
2. All functions which take a meta argument should have three options: keep previous metadata, take some new (explicit) metadata, or drop metadata. I am still unsure which should be the default. Maybe dropping by default with an easy option to keep it might be saner and an easier rule for users to remember (FWIW, this is what xarray advertise it does, but I don't see the option in their API documentation).

gdementen: I am unsure about those but I would rather keep metadata.

alixdamman: much better

gdementen:
hmm, wait... I see a few remaining problems...

larray/core/array.py
```diff
@@ -7097,20 +7142,23 @@ def wrapper(*args, **kwargs):
...
+def zeros(axes, title=None, dtype=float, order='C', meta=None):
     """Returns an array with the specified axes and filled with zeros.
 
     Parameters
     ----------
     axes : int, tuple of int, Axis or tuple/list/AxisCollection of Axis
         Collection of axes or a shape.
     title : str, optional
-        Title.
+        Deprecated. See 'meta' bellow.
```
gdementen: below

larray/core/array.py
```diff
@@ -7166,26 +7220,30 @@ def zeros_like(array, title='', dtype=None, order='K'):
...
+def ones(axes, title=None, dtype=float, order='C', meta=None):
     """Returns an array with the specified axes and filled with ones.
 
     Parameters
     ----------
     axes : int, tuple of int, Axis or tuple/list/AxisCollection of Axis
         Collection of axes or a shape.
     title : str, optional
-        Title.
+        Deprecated. See 'meta' bellow.
```
gdementen: below

larray/core/array.py
```diff
@@ -7200,25 +7258,31 @@ def ones(axes, title='', dtype=float, order='C'):
...
+def ones_like(array, title=None, dtype=None, order='K', meta=None):
     """Returns an array with the same axes as array and filled with ones.
 
     Parameters
     ----------
     array : LArray
         Input array.
     title : str, optional
-        Title.
+        Deprecated. See 'meta' bellow.
```
gdementen: below

larray/core/array.py
```diff
@@ -7134,25 +7182,31 @@ def zeros(axes, title='', dtype=float, order='C'):
          BE  0.0  0.0
          FO  0.0  0.0
     """
+    if title is not None:
+        import warnings
+        warnings.warn("Argument title is deprecated. Use meta instead", FutureWarning, stacklevel=2)
```
gdementen: when title is provided, and shouldn't we use `meta['title'] = title; title = None` (and initialize meta if None)? Otherwise, you will get twice the same warning: once in zeros (or other array creation functions) then once from `LArray.__init__`. You probably want to make an helper function for this to avoid duplication this code over and over.

larray/core/array.py
```diff
@@ -643,6 +642,17 @@ def nan_equal(a1, a2):
     return element_equal(a1, a2, nan_equals=True)
 
 
+def _handle_deprecated_argument_title(meta, title):
+    if meta is None:
+        meta = Metadata()
+    if title is not None:
+        import warnings
+        warnings.warn("title argument is deprecated. Please use meta argument instead", FutureWarning, stacklevel=2)
```
gdementen: did you test this? shouldn't that be stacklevel=3?

larray/core/array.py
```diff
@@ -7076,14 +7115,20 @@ def aslarray(a):
           c     3.0      1.0
     """
     if isinstance(a, LArray):
-        return a
+        res = a.copy()
```
gdementen: only copy if meta is not None? (np.asarray does not copy if possible)

larray/core/array.py
```diff
@@ -643,6 +642,17 @@ def nan_equal(a1, a2):
...
 
+def _handle_deprecated_argument_title(meta, title):
+    if meta is None:
+        meta = Metadata()
+    if title is not None:
+        import warnings
+        warnings.warn("title argument is deprecated. Please use meta argument instead", FutureWarning, stacklevel=2)
+        meta['title'] = title
+        title = None
+    return meta, title
```
gdementen: AFAICT, you don't have to return title since it is (or at least should) never (be) used

Issue #155, opened 2017-03-16
ensure it is easy to create groups from an Excel or .csv file


gdementen:
A good example is the NUTS codes from eurostat.

http://ec.europa.eu/eurostat/ramon/nomenclatures/index.cfm?TargetUrl=LST_CLS_DLD&StrNom=NUTS_33&StrLanguageCode=EN#

Maybe there is nothing specific to develop. Maybe directing users to use Pandas for that is fine, but we need at least to be able to tell users how to do it the easiest way. Currently the answer is "create the groups by hand", which is not good enough.

alixdamman:
See also #578

alixdamman:
See point ``2`` in #578

PR #647, opened 2018-06-11
Tutorial

alixdamman:
I would to update and reorganize the IO section of the tutorial. 
But first of all, I would to be sure about the "table of content". 

I assume that issue #155 will be done before this PR is merged


doc/source/tutorial/tutorial_IO.ipyml
```diff
@@ -1,17 +1,119 @@
...
+- markdown: |
+    ## Load and Dump Arrays
+    
+    ### Excel
+    
+    #### Loading an Array from an Excel Sheet
+    
+    ##### Basic Options
+    
+    ##### Specifying the Number of Axes
```
gdementen: will need #648

doc/source/tutorial/tutorial_IO.ipyml
```diff
@@ -1,17 +1,119 @@
...
+    
+    ##### NA and Missing Data Handling
+    
+    ##### Sorting Axes
+    
+    ##### Specifying Wide VS Narrow format
+    
+    ##### Specifying Position in Sheet (open_excel only)
+    
+    #### Dumping an Array to an Excel Sheet
```
gdementen: unsure where to place this but we need to show dumping and reading with header or without headers, including reconstructing an array with axes labels in arbitrary positions

doc/source/tutorial/tutorial_IO.ipyml
```diff
@@ -1,17 +1,119 @@
...
+    #### Dumping an Array to an HDF5 File
+    
+    ##### Metadata
+
+
+- markdown: |
+    ## Load and Dump Sessions
+    
+    
+    ### Loading a Session from an Excel Sheet
```
gdementen: s/Sheet/file|Workbook/?

doc/source/tutorial/tutorial_IO.ipyml
```diff
@@ -1,17 +1,119 @@
...
+    
+    #### Basic Options
+    
+    #### Specifying Objets To Be Loaded
+    
+    ### Dumping a Session to an Excel Sheet
```
gdementen: idem

doc/source/tutorial/tutorial_IO.ipyml
```diff
@@ -1,17 +1,119 @@
...
+    
+    ##### Metadata
+    
+    #### Dumping an Array to an HDF5 File
+    
+    ##### Metadata
+
+
+- markdown: |
+    ## Load and Dump Sessions
```
gdementen: what about load and dump sessions from .hdf?

alixdamman:
reminder: add a note about labels like ``102E3`` in Excel which are converted to integers by xlwings.

gdementen:
I always prefer to fix issues than document them.

doc/source/tutorial/tutorial_IO.ipyml
```diff
@@ -1,18 +1,605 @@
...
+  id: 5
+
+- markdown: |
+    #### Dumping Arrays to Excel Sheets with open_excel 
+
+
+- markdown: |
+    ##### Basic Usage
+    
+    ``open_excel`` must be used with the Python keyword ``with``. If the Excel file doesn't exist, the argument ``overwrite_file`` must set to True. The method ``save`` must be called to actually write data in the Excel file.
```
gdementen: s/must be/should be/?
s/``with``/``with``, which ensures that the file is properly closed even if an error occurs/

doc/source/tutorial/tutorial_IO.ipyml
```diff
@@ -1,18 +1,605 @@
...
+    arr2 = zeros((2, 2))
+    arr3 = ones((2, 2))
+    
+    with open_excel('new_excel_file.xlsx') as wb:
+        # add a new sheet 'arrays' and dump the array 'arr' starting at cell 'A1' 
+        wb['arrays'] = arr.dump()
+        # dump the array 'arr2' in sheet 'arrays' starting at cell 'A5' 
+        wb['arrays']['A5'] = arr2.dump()
+        # dump the array 'arr3' in sheet 'arrays' starting at cell 'A9' 
+        wb['arrays']['A9'] = arr3.dump()
```
gdementen: also show that if you have many arrays to store on the same sheet, you can store the sheet in a variable and use that?

```python
sheet = wb['arrays']
sheet['A5'] = arr2.dump()
```

doc/source/tutorial/tutorial_IO.ipyml
```diff
@@ -1,18 +1,605 @@
...
+- code: |
+    arr = ndtest((2, 2))
+    arr2 = zeros((2, 2))
+    arr3 = ones((2, 2))
+    
+    with open_excel('new_excel_file.xlsx') as wb:
+        # add a new sheet 'arrays' and dump the array 'arr' starting at cell 'A1' 
+        wb['arrays'] = arr.dump()
+        # dump the array 'arr2' in sheet 'arrays' starting at cell 'A5' 
+        wb['arrays']['A5'] = arr2.dump()
```
gdementen: please also show somewhere (here?) how to create an new empty/blank sheet (this question has come up several times in the past):
wb['new_sheet'] = ''

doc/source/tutorial/tutorial_IO.ipyml
```diff
@@ -1,18 +1,605 @@
...
+        # dump the array 'arr3' in sheet 'arrays' starting at cell 'A9' 
+        wb['arrays']['A9'] = arr3.dump()
+        wb.save()
+
+  id: 7
+
+- markdown: |
+    ##### Specifying Wide VS Narrow format
+    
+    Like with the ``to_excel`` method, it possible to export the data in a ``narrow`` format by setting the argument ``wide`` to False:
```
gdementen: it **is** possible

doc/source/tutorial/tutorial_IO.ipyml
```diff
@@ -1,18 +1,605 @@
...
+    # a0 | b1 | 1     |
+    # a1 | b0 | 2     |
+    # a1 | b1 | 3     |
+
+  id: 8
+
+- markdown: |
+    ##### Exporting only data
+    
+    To export only data in Excel sheets, you can set argument ``header`` of the method ``dump`` to false or not calling ``dump``:
```
gdementen: s/argument header/header argument
s/method dump/dump method/
s/false/False
s/not calling/avoid calling/

doc/source/tutorial/tutorial_IO.ipyml
```diff
@@ -1,18 +1,605 @@
...
+
+
+- code: |
+    arr = ndtest((2, 2))
+    
+    with open_excel('new_excel_file.xlsx') as wb:
+        # export only data
+        wb['data_only'] = arr.dump(header=False)
```
gdementen: When users export only data, they are usually in one of two situations:
1. they want to fill data in a "predefined/template file".
2. they want to export data and axes in a specific way. For example, two 2d tables on top of each other with a single "time" axis in columns but potentially blank lines between the two tables.

In both cases, it uses 'precise' position.
1. is probably hard to demonstrate but is **very** frequent so should at least be mentioned. If you can come up with a simple example to demonstrate this, it would be nice. Honestly I don't see how we could make this simple enough and still interesting but I think our users would appreciate if you managed that. Maybe first writing the array with headers, then saying this is the template to fill and updating only the values would be enough, or maybe using the specific formatted sheet done for 2. and updating the values would be more telling.
2. could be demonstrated, and if so, needs to include an example of how to output an axis labels vertically instead of horizontally (sheet[idx].options(transpose=True).value = axis.labels)

gdementen:
Sorry for the slow review. As usual, I started my review then got side tracked for long enough that I forgot about the review and started something else... When I don't review something for a day or two, feel free to ping me.

doc/source/tutorial/tutorial_IO.ipyml
alixdamman: Concerning the second situation, I'm thinking of an additional column 'description' next to a vertical axis column. What do you think?

doc/source/tutorial/tutorial_IO.ipyml
```diff
@@ -132,7 +145,8 @@ cells:
 - markdown: |
     ##### Basic Usage
     
-    ``open_excel`` must be used with the Python keyword ``with``. If the Excel file doesn't exist, the argument ``overwrite_file`` must set to True. The method ``save`` must be called to actually write data in the Excel file. 
+    ``open_excel`` should be used with the Python keyword ``with``, which ensures that the file is properly closed even if an error occurs.
+    If the Excel file doesn't exist, the argument ``overwrite_file`` must set to True. The method ``save`` must be called to actually write data in the Excel file.
```
gdementen: s/argument overwrite_file/overwrite_file argument/
s/must set/must be set/
s/method save/save method/

doc/source/tutorial/tutorial_IO.ipyml
```diff
@@ -201,28 +219,56 @@ cells:
...
+    ##### Exporting only data (without headers)
+    
+    To export only data in Excel sheets, you can choose between setting the ``header`` argument of the ``dump`` method to False or avoid calling ``dump``. 
+    
+    This can be usefull as you to export data following a specific format. For example, let's consider you have a 2D array `pop` representing the projection of the population of some selected cities until 2025. Let's assume this array has two axes `city` and `time` respectively but you need to include the associated continent in the output file. What you want is:  
     
-    To export only data in Excel sheets, you can set argument ``header`` of the method ``dump`` to false or not calling ``dump``:
+    |city     |	continent	  | 2015 | 2016 | 2017 | 2018 | 2019 | 2020 | 2021 | 2022 | 2023 | 2024 | 2025 |
+    | ------- | ------------- | ---- | ---- | ---- | ---- | ---- | ---- | ---- | ---- | ---- | ---- | ---- |
+    |Brussel  |	Europe	      | 0    | 1	| 2	   | 3	  | 4    | 5    | 6    | 7    | 8    | 9    | 10   |
```
gdementen: Nice example!! I would have used fewer years though. Beyond 4 or 5, it doesn't add anything to the example and it makes it unnecessarily big.

doc/source/tutorial/tutorial_IO.ipyml
```diff
@@ -201,28 +219,56 @@ cells:
...
+    city = Axis("city=Brussel,Paris,New York,Mexico,Cairo,Tokyo,Moscow")
+    continent = Axis("continent=Europe,Europe,North America,North America,Africa,Asia,Asia")
+    time = Axis("time=2015..2025")
+    pop = ndtest((city, time))
     
-    with open_excel('new_excel_file.xlsx') as wb:
-        # export only data
-        wb['data_only'] = arr.dump(header=False)
+    with open_excel('pop_projection.xlsx', overwrite_file=True) as wb:
+        # create new sheet 'pop_2015_2025'
```
gdementen: s/new/new empty/

doc/source/tutorial/tutorial_IO.ipyml
```diff
@@ -201,28 +219,56 @@ cells:
...
+        wb['pop_2015_2025'] = []
+        sh = wb['pop_2015_2025']
```
gdementen: Add a comment for this line too (like in the previous example)? It doesn't hurt to repeat some stuff :)

doc/source/tutorial/tutorial_IO.ipyml
```diff
@@ -201,28 +219,56 @@ cells:
...
+        # export column names
+        sh['A1'] = ['city', 'continent'] + list(time.labels)
```
gdementen: maybe split this line in two to avoid having to convert to list?
```python
sh['A1'] = ['city', 'continent']
sh['C1'] = time.labels
```

doc/source/tutorial/tutorial_IO.ipyml
```diff
@@ -8,11 +8,25 @@ cells:
...
+    sys.path.insert(0, os.path.abspath('../../../'))
+
+  id: 0
+
 - code: |
     # first of all, import the LArray library
     from larray import *
+    
+    # then, let's define a function to get full path of any example file
+    def get_filepath(fname):
```
gdementen: maybe we should add a function like that to larray itself?

doc/source/tutorial/tutorial_IO.ipyml
```diff
@@ -233,31 +247,33 @@ cells:
...
 - code: |
     city = Axis("city=Brussel,Paris,New York,Mexico,Cairo,Tokyo,Moscow")
     continent = Axis("continent=Europe,Europe,North America,North America,Africa,Asia,Asia")
-    time = Axis("time=2015..2025")
+    time = Axis("time=2015..2020")
     pop = ndtest((city, time))
     
     with open_excel('pop_projection.xlsx', overwrite_file=True) as wb:
-        # create new sheet 'pop_2015_2025'
+        # create new empty sheet 'pop_2015_2025'
```
gdementen: 2020

doc/source/tutorial/tutorial_IO.ipyml
```diff
@@ -233,31 +247,33 @@ cells:
...
         wb['pop_2015_2025'] = []
+        # store sheet 'pop_2015_2025' in a temporary variable sh
```
gdementen: 2020

doc/source/tutorial/tutorial_IO.ipyml
```diff
@@ -233,31 +247,33 @@ cells:
...
     city = Axis("city=Brussel,Paris,New York,Mexico,Cairo,Tokyo,Moscow")
     continent = Axis("continent=Europe,Europe,North America,North America,Africa,Asia,Asia")
-    time = Axis("time=2015..2025")
+    time = Axis("time=2015..2020")
     pop = ndtest((city, time))
     
     with open_excel('pop_projection.xlsx', overwrite_file=True) as wb:
-        # create new sheet 'pop_2015_2025'
+        # create new empty sheet 'pop_2015_2025'
         wb['pop_2015_2025'] = []
```
gdementen: 2020

doc/source/tutorial/tutorial_IO.ipyml
```diff
@@ -233,31 +247,33 @@ cells:
...
+        # store sheet 'pop_2015_2025' in a temporary variable sh
         sh = wb['pop_2015_2025']
```
gdementen: 2020

larray/inout/csv.py
```diff
@@ -118,26 +118,32 @@ def read_csv(filepath_or_buffer, nb_axes=None, index_col=None, sep=',', headerse
...
-    a\{1}  b0  b1  b2
-       a0   0   1   2
-       a1   3   4   5
-       a2   6   7   8
+    a,b,c0,c1
+    a0,b0,0,1
+    a0,b1,2,3
+    a1,b0,4,5
+    a1,b1,6,7
+    >>> # read the array stored in the CSV file as it
```
gdementen: "as is", not "as it" (it was correct before ;-))

larray/inout/csv.py
```diff
@@ -118,26 +118,32 @@ def read_csv(filepath_or_buffer, nb_axes=None, index_col=None, sep=',', headerse
...
+    >>> arr = read_csv(fname)
+    >>> arr.info
+    4 x 3
+     a [4]: 'a0' 'a0' 'a1' 'a1'
+     {1} [3]: 'b' 'c0' 'c1'
+    dtype: object
+    memory used: 96 bytes
```
gdementen: adding a comment somewhere to say that this is not what we want/need/expected would help I think

larray/inout/excel.py
```diff
@@ -117,18 +117,23 @@ def read_excel(filepath, sheet=0, nb_axes=None, index_col=None, fill_value=np.na
 
     Specify the number of axes of the output array (useful when the name of the last axis is implicit)
 
-    >>> # read the array stored in the CSV file as it
-    >>> read_excel(fname, sheet='missing_axis_name')
-    a\{1}  b0  b1  b2
-       a0   0   1   2
-       a1   3   4   5
-       a2   6   7   8
+    >>> # read the array stored in the sheet 'missing_axis_name' as it
```
gdementen: as is

larray/tests/test_excel.py
```diff
@@ -80,11 +80,6 @@ def test_setitem(self):
             assert wb.sheet_names() == ['sheet1', 'sheet2', 'sheet3']
             assert wb['sheet2']['A1'].value == 'sheet1 content'
 
-            with open_excel(visible=False, app="new") as wb2:
```
gdementen: Please keep this test intact. It **must** raise. It should be fixed in master (commit c8aac0b576599b8ba6dabddbb0a760b7fa3d49ac). I guess you haven't rebased yet.

doc/source/tutorial/tutorial_IO.ipyml
alixdamman: I don't really think that users will play with example files outside the tutorial.

gdementen: maybe not, but it would make the tutorial and at least one doctest a bit more readable

alixdamman: OK. It doesn't hurt anyway.

larray/inout/excel.py
```diff
@@ -117,8 +117,18 @@ def read_excel(filepath, sheet=0, nb_axes=None, index_col=None, fill_value=np.na
 
     Specify the number of axes of the output array (useful when the name of the last axis is implicit)
 
-    >>> # read the array stored in the sheet 'missing_axis_name' as it
+    >>> # The content of the sheet 'missing_axis_name' is:
```
gdementen: remove the >>> #

larray/example.py
```diff
@@ -1,12 +1,41 @@
...
+        Filename of an existing example file.
+
+    Returns
+    -------
+    Filepath
+        Absolute filepath to an example file if exists.
+
+    Notes
+    -----
+    A ValueError is raised if provided filename does not represent an existing example file.
```
gdementen: the provided

larray/tests/test_misc.py
```diff
@@ -0,0 +1,19 @@
+from __future__ import absolute_import, division, print_function
+
+import pytest
+from larray import get_example_filepath
+
+
+def test_get_example_filepath():
+    with pytest.raises(ValueError, message="Example file non_existing_example_file.xlsx does not exist. "
```
gdementen: This will be annoying. We will need to change the test each time we add or rename an example file. It would be better to use a pattern to only check the beginning of the error message.

larray/example.py
```diff
@@ -25,14 +25,14 @@ def get_example_filepath(fname):
...
-    A ValueError is raised if provided filename does not represent an existing example file.
+    A ValueError is raised if the provided filename does not represent an existing example file.
 
     Examples
     --------
     >>> fpath = get_example_filepath('examples.xlsx')
     """
     fpath = os.path.abspath(os.path.join(EXAMPLE_FILES_DIR, fname))
-    if not os.path.isfile(fpath):
+    if not (os.path.isfile(fpath) or os.path.isdir(fpath)):
```
gdementen: use os.path.exists instead?

larray/inout/csv.py
```diff
@@ -75,98 +75,106 @@ def read_csv(filepath_or_buffer, nb_axes=None, index_col=None, sep=',', headerse
...
-    >>> fname = get_example_filepath('test2d.csv')
+    >>> csv_dir = get_example_filepath('examples')
+    >>> fname = csv_dir + '/pop.csv'
     >>> read_csv(fname)
-    a\\b  b0  b1
-      1   0   1
-      2   2   3
-      3   4   5
+    city\\time  2010  2011  2012
+      Brussel     0     1     2
```
gdementen: s/Brussel/Brussels/

doc/source/tutorial/tutorial_IO.ipyml
```diff
@@ -579,10 +623,12 @@ cells:
 - code: |
 
 
-  id: 38
+  id: 40
 
 - markdown: |
     ## Load and Dump Sessions
+    
+    One of the main advantage of grouping arrays in session objects is that you can load and save all them in one shot.
```
gdementen: * advantages
* all **of** them

doc/source/tutorial/tutorial_IO.ipyml
```diff
@@ -307,23 +307,23 @@ cells:
...
     
-    | city    | gender\time | 2010 | 2011 | 2012 |
-    | ------- | ----------- | ---- | ---- | ---- |
-    | Brussel | female      | 0    | 1    | 2    |
-    | Brussel | male        | 3    | 5    | 6    |
-    | Paris   | female      | 7    | 8    | 9    |
-    | Paris   | male        | 10   | 11   | 12   |
+    | city     | gender\time | 2010 | 2011 | 2012 |
+    | -------- | ----------- | ---- | ---- | ---- |
+    | Brussels | female      | 0    | 1    | 2    |
```
gdementen: Use eurostat data here too?

larray/inout/csv.py
```diff
@@ -31,12 +31,13 @@ def read_csv(filepath_or_buffer, nb_axes=None, index_col=None, sep=',', headerse
...
     -----
     csv file format:
-    arr,ages,sex,nat\time,1991,1992,1993
-    A1,BI,H,BE,1,0,0
-    A1,BI,H,FO,2,0,0
-    A1,BI,F,BE,0,0,1
-    A1,BI,F,FO,0,0,0
-    A1,A0,H,BE,0,0,0
+    geo,gender\\time,2015,2016,2017
+    BE,M,5524068.0,5569264.0,5589272.0
```
gdementen: wouldn't using longer labels make things more readable? (BE -> Belgium, M -> Male, ...)?

larray/tests/data/examples/fertility.csv
```diff
@@ -0,0 +1,7 @@
+geo,gender\time,2013,2014,2015
+BE,M,64371,64173,62561
```
gdementen: These seem to be birth numbers, not fertility (which I *assume* are usually given as rates).

alixdamman: The dataset comes from ``Database by themes > Population and social conditions > Demography and migration (demo) > Fertility``

doc/source/tutorial/tutorial_IO.ipyml
alixdamman: Yes, it's in the pipe. Give me some time.

alixdamman:
Could you please review only the last commit?

larray/inout/excel.py
```diff
@@ -178,6 +178,17 @@ def read_excel(filepath, sheet=0, nb_axes=None, index_col=None, fill_value=np.na
     geo\\time      2013      2014      2015
      Belgium  11137974  11180840  11237274
       France  65600350  65942267  66456279
+
+    Extract array from a given range (useful when several arrays are stored in the same sheet)
```
gdementen: There are other cases where this could be useful (e.g. if there is some text before, after or next to the data you want to load), so I am unsure the parentheses add anything.

larray/inout/excel.py
```diff
@@ -197,7 +208,9 @@ def read_excel(filepath, sheet=0, nb_axes=None, index_col=None, fill_value=np.na
                             .format(list(kwargs.keys())[0]))
         from larray.inout.xw_excel import open_excel
         with open_excel(filepath) as wb:
-            return wb[sheet].load(index_col=index_col, fill_value=fill_value, sort_rows=sort_rows,
+            if range is None:
+                range = slice(None, None, None)
```
gdementen: * `range = slice(None)` is enough. 
* you could have `range` default to slice(None) directly since a slice object is readonly.

larray/inout/excel.py
```diff
@@ -61,7 +61,7 @@ def read_excel(filepath, sheet=0, nb_axes=None, index_col=None, fill_value=np.na
         Engine to use to read the Excel file. If None (default), it will use 'xlwings' by default if the module is
         installed and relies on Pandas default reader otherwise.
     range : str, optional
-        Range in which array is stored. Used only if engine is 'xlwings'. Defaults to None.
+        Range in which array is stored. Used only if engine is 'xlwings'. Defaults to slice(None).
```
gdementen: should explain what that slice(None) means. Something like "(which means the entire sheet)"

doc/source/tutorial/tutorial_IO.ipyml
```diff
@@ -555,75 +555,169 @@ cells:
...
   id: 34
 
-- code: |
+- markdown: |
+    2) Call the ``load`` method on an existing session and pass the path to the Excel/HDF5 file or to the directory containing CSV files as first argument:
 
 
+- code: |
+    # create a session with two arrays
+    session = Session([('arr1', ndtest((3, 3))), ('arr2', ndtest((2, 2, 2)))])
```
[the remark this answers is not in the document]
alixdamman: The idea is to show that the `load` method add new items to a session and does not delete the existing items.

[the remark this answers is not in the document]
gdementen: I would rather use an empty session (using Session()) and use .load several times on it IF THAT MAKES SENSE given the current files (if you have distinct arrays in different files), because this is closer to what people actually do. If that's not the case, leave the thing as it is.

PR #649, opened 2018-06-12
misc fixes & comments

gdementen:


alixdamman:
LGTM

alixdamman:
in file `setup.cfg`, there is a commented option to ask pytest to also test PEP8 compliance. 
Do you want to unlock this option?

gdementen:
> in file setup.cfg, there is a commented option to ask pytest to also test PEP8 compliance.
Do you want to unlock this option?

It is worth a try... Depends if we can configure some of it (for example 120 char lines -- otherwise we will be swamped in messages) and if there are not too many false positives.

alixdamman:
Yes, you can change the maximum number of characters per line: `pep8maxlinelength = 119` (already in `setup.cfg`). For the other options, I don't know. Examples are given here: https://pypi.org/project/pytest-pep8/

PR #651, opened 2018-06-19
choice(larray)

gdementen:
a small PR on top of #644

### DIFF
--- a/larray/random.py
+++ b/larray/random.py
@@ -25,7 +25,7 @@
 # SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 import numpy as np
 
-from larray.core import AxisCollection, LArray, aslarray
+from larray.core import Axis, AxisCollection, LArray, aslarray
 from larray.core.array import raw_broadcastable
 import larray as la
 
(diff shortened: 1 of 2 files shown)


gdementen:
oops, I forgot to document the extra functionality in choices docstring

gdementen:
I wonder if choice(p=larray), wouldn't be more logical than choice(choices=larray) 
That is getting choices from the probabilities array, instead of taking the probabilities from the choices array.

larray/random.py
```diff
@@ -0,0 +1,456 @@
...
+        If an LArray, a random sample is generated using its axis as choices and values as probabilities.
+        If an ndarray, a random sample is generated from its elements.
+        If an int n, the random sample is generated as if choices was la.sequence(n)
+    axes : int, tuple of int, str, Axis or tuple/list/AxisCollection of Axis, optional
+        Axes (or shape) of the resulting array. If ``axes`` is None (the default), a single value is returned.
+        Otherwise, if the resulting axes have a shape of, e.g., ``(m, n, k)``, then ``m * n * k`` samples are drawn.
+    replace : boolean, optional
+        Whether the sample is with or without replacement.
+    p : 1-D array-like, optional
+        The probabilities associated with each entry in a.
```
alixdamman: s/a/choices/ ?

larray/random.py
```diff
@@ -0,0 +1,456 @@
...
+    --------
+    LArray or scalar
+        The generated random samples with given ``axes`` (or shape).
+
+    Raises
+    -------
+    ValueError
+        If a is an int and less than zero, if a or p are not 1-dimensional, if a is an array-like of size 0,
+        if p is not a vector of probabilities, if a and p have different lengths, or
+        if replace=False and the sample size is greater than the population size.
```
alixdamman: s/a/choices/ in all `Raises` section

larray/random.py
gdementen: good catch. Bad copy-paste from numpy.

larray/random.py
gdementen: idem

gdementen:
@alixdamman : your opinion on this:
> I wonder if choice(p=larray), wouldn't be more logical than choice(choices=larray)
That is getting choices from the probabilities array, instead of taking the probabilities from the choices array.

?

alixdamman:
It makes sense to me.

Issue #648, opened 2018-06-12
excel.Range.load and excel.Sheet.load do not have nb_axes argument


gdementen:
They still use nb_index, which produces a warning.

PR #655, opened 2018-06-21
issue #648 : deprecated argument nb_index in favor of nb_axes in excel.(...).load methods

alixdamman:


alixdamman:
I hesitated indeed. OK to add one.

PR #659, opened 2018-06-29
Add metadata to Session objects

alixdamman:
Needs changelog

### DIFF
--- a/doc/source/changes/version_0_29.rst.inc
+++ b/doc/source/changes/version_0_29.rst.inc
@@ -75,8 +75,8 @@ New features
     False
 
 * allowed arrays to have metadata (e.g. title, description, authors, ...).
-  Metadata can be accessed using the syntax ``array.meta.name``.
-  See bellow for use cases:
+  Array metadata can be accessed using the syntax ``array.meta.name``.
+  See below for use cases:
 
     >>> # add metadata to/from an array
     >>> # for Python <= 3.5
@@ -93,8 +93,8 @@ New features
 
   Warnings:
 
-  - Currently, only the HDF (.h5) file format supports saving and loading metadata.
-  - Metadata are not kept when actions or methods are applied on a array
+  - Currently, only the HDF (.h5) file format supports saving and loading array metadata.
+  - Metadata is not kept when actions or methods are applied on an array
     except for operations modifying the object in-place, such as `pop[age < 10] = 0`,
     and when the method `copy()` is called. Do not add metadata to an array if you know
     you will apply actions or methods on it before dumping it.
@@ -125,6 +125,32 @@ New features
     >>> import larray as la
     >>> la.pi
     3.141592653589793
+
+
+* allowed sessions to have metadata.
+  Like for arrays, session metadata can be accessed using the syntax ``session.meta.name``.
+  See below for use cases:
+
+    >>> # add metadata to a session at creation
+    >>> # Python <= 3.5
+    >>> s = Session([('arr1', ndtest(2)), ('arr2', ndtest(3)], meta=[('title', 'my title'), ('author', 'John Smith')])
+    >>> # Python 3.6+
+    >>> s = Session(arr1=ndtest(2), arr2=ndtest(3), meta=Metadata(title='my title', author='John Smith'))
+    >>> # show/access metadata
+    >>> s.meta
+    title: my title
+    author: John Smith
+
+  Warnings:
+
+  - Contrary to array metadata, saving and loading session metadata is supported for
+    all current session file formats: Excel, CSV and HDF (.h5)
+  - Metadata is not kept when actions or methods are applied on a session
+    except for operations modifying a specific array, such as: `s['arr1'] = 0`.
+    Do not add metadata to a session if you know you will apply actions or methods
+    on it before dumping it.
+
+  Closes :issue:`640`.
 
 
 .. _misc:
--- a/larray/core/session.py
+++ b/larray/core/session.py
@@ -8,6 +8,7 @@ from collections import OrderedDict
 
 import numpy as np
 
+from larray.core.metadata import Metadata
 from larray.core.group import Group
 from larray.core.axis import Axis
 from larray.core.constants import nan
@@ -31,7 +32,18 @@ class Session(object):
         Path to the file containing the session to load or
         list/tuple/dictionary containing couples (name, object).
     **kwargs : dict of {str: object}
-        Objects to add written as name=object, ...
+
+        * Objects to add written as name=object
+        * meta : list of pairs or dict or OrderedDict or Metadata
+            Metadata (title, description, author, creation_date, ...) associated with the array.
+            Keys must be strings. Values must be of type string, int, float, date, time or datetime.
+
+    Warnings
+    --------
+    Metadata is not kept when actions or methods are applied on a session
+    except for operations modifying a specific array, such as: `s['arr1'] = 0`.
+    Do not add metadata to a session if you know you will apply actions or methods
+    on it before dumping it.
 
     Examples
     --------
@@ -57,9 +69,27 @@ class Session(object):
     load Session from file
 
     >>> s = Session('my_session.h5')  # doctest: +SKIP
+
+    create a session with metadata
+
+    >>> # Python <= 3.5
+    >>> s = Session([('arr1', arr1), ('arr2', arr2)], meta=[('title', 'my title'), ('author', 'John Smith')])
+    >>> s.meta
+    title: my title
+    author: John Smith
+    >>> # Python 3.6+
+    >>> s = Session(arr1=arr1, arr2=arr2, meta=Metadata(title='my title', author='John Smith'))  # doctest: +SKIP
+    >>> s.meta
+    title: my title
+    author: John Smith
     """
     def __init__(self, *args, **kwargs):
         object.__setattr__(self, '_objects', OrderedDict())
+
+        meta = kwargs.pop('meta', None)
+        if meta is None:
+            meta = Metadata()
+        self.meta = meta
 
         if len(args) == 1:
             a0 = args[0]
@@ -172,14 +202,24 @@ class Session(object):
     def __delitem__(self, key):
         del self._objects[key]
 
+    # TODO: add a a meta property when Python 2.7 will be dropped
     def __getattr__(self, key):
-        if key in self._objects:
+        if key is 'meta':
+            return self._meta
+        elif key in self._objects:
             return self._objects[key]
         else:
             raise AttributeError("'{}' object has no attribute '{}'".format(self.__class__.__name__, key))
 
+    # TODO: implement meta.setter when Python 2.7 will be dropped
     def __setattr__(self, key, value):
-        self._objects[key] = value
+        if key is 'meta':
+            if not isinstance(value, (list, dict, OrderedDict, Metadata)):
+                raise TypeError("Expected list of pairs or dict or OrderedDict or Metadata object "
+                                "instead of {}".format(type(value).__name__))
+            object.__setattr__(self, '_meta', value if isinstance(value, Metadata) else Metadata(value))
+        else:
+            self._objects[key] = value
 
     def __delattr__(self, key):
         del self._objects[key]
@@ -198,7 +238,7 @@ class Session(object):
 
     def load(self, fname, names=None, engine='auto', display=False, **kwargs):
         """
-        Loads LArray, Axis and Group objects from a file, or several .csv files.
+        Load LArray, Axis and Group objects from a file, or several .csv files.
 
         WARNING: never load a file using the pickle engine (.pkl or .pickle) from an untrusted source, as it can lead
         to arbitrary code execution.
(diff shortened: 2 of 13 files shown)


larray/core/session.py
```diff
@@ -873,8 +946,9 @@ def elem_equal(e1, e2):
 
     array_equals = renamed_to(element_equals, 'array_equals')
 
+    # TODO: include metadata?
```
gdementen: This could be an option. But we need a default value and I am unsure about it. Do you have an idea?

larray/core/session.py
```diff
@@ -1091,7 +1165,7 @@ def summary(self, template=None):
                 - for groups: 'key', 'name', 'axis_name', 'labels' and 'length',
                 - for axes: 'key', 'name', 'labels' and 'length',
                 - for arrays: 'key', 'axes_names', 'shape', 'dtype' and 'title',
-                - for all other types: 'key', 'value'.
+                - for metadata: 'key', 'value'.
```
gdementen: what about other kinds of objects?

larray/core/session.py
```diff
@@ -1150,9 +1231,15 @@ def summary(self, template=None):
             template[Group] = "{key}: {axis_name}{labels} >> {name} ({length})"
         if LArray not in template:
             template[LArray] = "{key}: {axes_names} ({shape}) [{dtype}]\n    {title}"
+        if Metadata not in template:
+            template[Metadata] = "{key} (metadata): {value}"
 
         def display(k, v):
-            t = Group if isinstance(v, Group) else type(v)
+            if not isinstance(v, (Axis, Group, LArray)):
```
gdementen: I would **much** rather have a single template for Metadata objects, so that users can chose to indent it more than the rest or not to display it at all. It would also (re-) allow user-defined objects in the Session.

larray/inout/excel.py
```diff
@@ -249,13 +256,26 @@ def _dump_item(self, key, value, *args, **kwargs):
...
+    def _dump_metadata(self, metadata):
+        if len(metadata) > 0:
+            s = _metadata_to_series(metadata)
+            s.to_excel(self.handle, '__metadata__', engine='xlsxwriter')
+
     def save(self):
         if len(self.axes) > 0:
             df = _axes_to_df(self.axes.values())
-            df.to_excel(self.handle, '__axes__', engine='xlsxwriter')
+            df.to_excel(self.handle, '__axes__', index=False, engine='xlsxwriter')
```
gdementen: is this change really related to session metadata???

larray/inout/pandas.py
```diff
@@ -291,9 +293,27 @@ def _series_to_group(series, axis):
     return LGroup(key=series.values, name=name, axis=axis)
 
 
-# ################################## #
-#    DATAFRAME <--> AXES, GROUPS     #
-# ################################## #
+def _metadata_to_series(metadata):
+    s = pd.Series(data=list(metadata.values()), index=metadata.keys(), name='')
```
gdementen: Why don't you use a 1D object LArray for this?

larray/inout/pandas.py
```diff
@@ -291,9 +293,27 @@ def _series_to_group(series, axis):
...
+    s.index.name = 'METADATA'
+    return s
+
+
+def _series_to_metadata(series):
+
+    def _convert_value(key, value):
```
gdementen: this function does not seem to use the key argument. Remove it.

larray/core/session.py
```diff
@@ -172,13 +202,21 @@ def __delitem__(self, key):
         del self._objects[key]
 
     def __getattr__(self, key):
-        if key in self._objects:
+        if key is 'meta':
```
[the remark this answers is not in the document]
alixdamman: I started by trying to implement ``meta`` as a property but I faced some difficulties to make it work with Python 2.7

larray/tests/test_session.py
```diff
@@ -469,11 +495,17 @@ def test_rdiv(session):
...
-def test_pickle_roundtrip(session):
+def test_pickle_roundtrip(session, meta):
     original = session.filter(kind=LArray)
+    original.meta = meta
     s = pickle.dumps(original)
     res = pickle.loads(s)
     assert res.equals(original)
+    if PY2:
+        from larray.core.metadata import Metadata
+        assert Metadata(res.meta) == meta
```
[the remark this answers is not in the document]
alixdamman: I'm not proud of it. After trying to implement ``meta`` in Session as a property and facing difficulties with Python 2.7 I just gave up for this part. I will change that.

larray/inout/hdf.py
```diff
@@ -165,5 +166,12 @@ def _dump_item(self, key, value, *args, **kwargs):
         else:
             raise TypeError()
 
+    def _read_metadata(self):
+        attrs = self.handle.get_node('')._v_attrs
```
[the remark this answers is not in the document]
alixdamman: I didn't found any other way to add 'global' metadata to the HDF5 file (see https://www.pytables.org/usersguide/tutorials.html#setting-and-getting-user-attributes for more details)

larray/core/session.py
```diff
@@ -262,9 +315,10 @@ def load(self, fname, names=None, engine='auto', display=False, **kwargs):
             handler = handler_cls(fname, kwargs['sep'])
         else:
             handler = handler_cls(fname)
-        objects = handler.read_items(names, display=display, **kwargs)
+        metadata, objects = handler.read_items(names, display=display, **kwargs)
```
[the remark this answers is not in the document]
alixdamman: I load metadata from `read_items()` to not call `self._open_for_read()` and `self.close()` twice.

larray/inout/excel.py
alixdamman: Nope, sorry.

larray/inout/excel.py
```diff
@@ -190,14 +193,14 @@ def _load_axes_and_groups(self):
         # load all axes
         sheet_axes = '__axes__'
         if sheet_axes in self.handle.sheet_names:
-            df = pd.read_excel(self.handle, sheet_axes)
+            df = pd.read_excel(self.handle, sheet_axes, index_col=None)
```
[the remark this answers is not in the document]
alixdamman: Nope, sorry.

larray/core/session.py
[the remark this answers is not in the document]
gdementen: Yes, I understand why you did this and it was the best move without modifying too much the FileHandler API. It's just that I think we can do better if we change it. Now, this is no big deal while that API is purely internal but I think it could be useful to make it public some day and in that case, the current API is not good enough.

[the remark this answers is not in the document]
alixdamman: I agree

larray/core/session.py
alixdamman: IMO metadata should not be tested in equals. There are not properly items of a Session.

gdementen: ok, let's go for this. Should be documented explicitly though.

larray/inout/pandas.py
alixdamman: Because LArray objects are converted to pandas objects before to be dumped or loaded. I think using 1D array here instead of a pandas Series will add an unnecessary layer. Especially since we need to iterate over each element at reading to convert each metadata to the right type.

gdementen: This might be a case of premature optimization. Passing via LArray has negligible cost in my working copy (disclaimer: I modified .series slightly to go faster for 1D and I had already a few speed optimizations all over the place in my working copy, so this might not be the case on your computer).

```python
>>> metadata = {'title': 'yad aez dgrz jgzeigfejzi', 'date': datetime.now()}
>>> %timeit l = LArray(list(metadata.values()), Axis(metadata.keys())); s = l.series; s.name = ''
1.22 ms ± 57.8 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
>>> %timeit s = pd.Series(data=list(metadata.values()), index=metadata.keys(), name='')
1.3 ms ± 65.8 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```

Besides, I think that it might be a case where avoiding pandas structures altogether would be faster. Going via pandas is a net win for reading .csv files because their parser is so fast, but for anything else, we only gain convenience, not performance. This is probably especially true when we go via xlwings, which in the end must transform everything to "simple" Python types. Anyway, you can leave this as-is, but please at least add a comment that this is done for performance reasons.

larray/inout/pandas.py
```diff
@@ -291,9 +293,27 @@ def _series_to_group(series, axis):
...
+        value = str(value)
```
gdementen: Why is this line needed? It seems weird to convert numeric types to string to convert them back on the next line to numeric?

larray/inout/pandas.py
```diff
@@ -291,9 +293,27 @@ def _series_to_group(series, axis):
...
+        value = pd.to_numeric([value], errors='ignore')[0]
```
gdementen: why do you use [value] instead of just value? pd.to_numeric seems to handle scalars just fine (even if that is undocumented)

larray/core/session.py
alixdamman: The docstring starts with "*Test if groups, axes and arrays of the current session are equal to those of another session*". It's quite explicit no?

gdementen: Not enough I think. Users have to think for themselves that "hey metadata is not in that list". I think there should be a note somewhere stating something vaguely like: "other metadata (from .meta) is **NOT** compared" or "is ignored".

larray/inout/pandas.py
alixdamman: What version of pandas do you have? If I try to pass value instead of [value], I get the following error message: TypeError: 'int' object is not subscriptable

gdementen:
```python
>>> pd.to_numeric('10', errors='ignore')
10
>>> pd.to_numeric(10, errors='ignore')
10
>>> pd.__version__
'0.20.3'
```

larray/core/session.py
```diff
@@ -1170,14 +1273,18 @@ def display(k, v):
             else:
                 return tmpl(k, v)
 
-        return '\n'.join(display(k, v) for k, v in self.items())
+        res = ''
+        if len(self.meta) > 0:
+            res = '\n'.join(display_metadata(k, v) for k, v in self.meta.items()) + '\n'
+        res += '\n'.join(display_items(k, v) for k, v in self.items())
```
gdementen: then here, something like:
```python
if self.meta:
    res += display_items('Metadata', v) + '\n'
res += '\n'.join(display_items(k, v) for k, v in self.items())
```

The goal being that users can produce something like the following if they want to (maybe it should be the default):
```python
Metadata: 
    title: my title
    author: John Smith
axis1: a ['a0' 'a1' 'a2'] (3)
group1 -> a01: a['a0', 'a1'] (2)
```
**Now, I am unsure it's worth it** to change what you have but having no way to distinguish a metadata item from a "normal" item clearly bothers me.

alixdamman:
There is a bug with conda for Python 2.7 : https://github.com/conda/conda/issues/7503

doc/source/changes/version_0_29.rst.inc
```diff
@@ -48,7 +48,7 @@ New features
 
   Warnings:
 
-  - Currently, only the HDF (.h5) file format supports saving and loading metadata.
+  - Currently, only the HDF (.h5) file format supports saving and loading array metadata.
   - Metadata are not kept when actions or methods are applied on a array
```
gdementen: s/Metadata are/Metadata is/
s/a/an/

doc/source/changes/version_0_29.rst.inc
```diff
@@ -67,6 +67,31 @@ New features
     >>> arr.meta.title = 'array for testing'
 
 
+* allowed sessions to have metadata.
+  Like for arrays, session metadata can be accessed using the syntax ``session.meta.name``.
+  See bellow for use cases:
```
gdementen: s/bellow/below/

doc/source/changes/version_0_29.rst.inc
```diff
@@ -67,6 +67,31 @@ New features
...
+    >>> # Python 3.6+
+    >>> s = Session(arr1=ndtest(2), arr2=ndtest(3), meta=Metadata(title='my title', author='John Smith'))
+    >>> # show/access metadata
+    >>> s.meta
+    title: my title
+    author: John Smith
+
+  Warnings:
+
+  - Excel, CSV and HDF (.h5) file formats support saving and loading session metadata.
```
gdementen: * These are all the common formats, so it is a bit weird to have this in a Warning section, but maybe putting that in a Notes section by itself would be impractical.
* I would emphasize more the difference with arrays by saying something like "Contrary to array metadata, saving and loading session metadata is supported for all current session file formats: Excel, CSV and HDF (.h5).

doc/source/changes/version_0_29.rst.inc
```diff
@@ -67,6 +67,31 @@ New features
...
+  - Metadata are not kept when actions or methods are applied on a session
```
gdementen: s/are/is/ 😄

larray/core/session.py
```diff
@@ -30,7 +31,18 @@ class Session(object):
...
-        Objects to add written as name=object, ...
+
+        * Objects to add written as name=object
+        * meta : list of pairs or dict or OrderedDict or Metadata
+            Metadata (title, description, author, creation_date, ...) associated with the array.
+            Keys must be strings. Values must be of type string, int, float, date, time or datetime.
+
+    Warnings
+    --------
+    Metadata are not kept when actions or methods are applied on a session
```
gdementen: are...is...

larray/core/session.py
```diff
@@ -1164,20 +1258,25 @@ def display(k, v):
...
+                                       shape=' x '.join(str(i) for i in v.shape), dtype=v.dtype)
                 else:
                     return tmpl.format(key=k, value=str(v))
             else:
                 return tmpl(k, v)
 
-        return '\n'.join(display(k, v) for k, v in self.items())
+        res = ''
+        if len(self.meta) > 0:
+            res = 'Metadata:\n'
```
gdementen: Not what I had in mind, but oh well... I am tired of this...

PR #663, opened 2018-07-12
choice(ndlarray)

gdementen:
added support for choice with an ND LArray

### DIFF
--- a/larray/core/axis.py
+++ b/larray/core/axis.py
@@ -2645,6 +2645,46 @@ class AxisCollection(object):
         axes_changes = list(zip(join_axes, new_axes))
         return self.replace(axes_changes), other.replace(axes_changes)
 
+    # XXX: make this into a public method/property? axes_col.flat_labels[flat_indices]?
+    def _flat_lookup(self, flat_indices):
+        r"""Return labels corresponding to indices into the flattened axes
+
+        Parameters
+        ----------
+        flat_indices : array-like
+            indices to get
+
+        Examples
+        --------
+        >>> from larray import ndtest, LArray
+        >>> arr = ndtest((2, 3))
+        >>> arr
+        a\b  b0  b1  b2
+         a0   0   1   2
+         a1   3   4   5
+        >>> indices = LArray([2, 5, 0], 'draw=d0..d2')
+        >>> indices
+        draw  d0  d1  d2
+               2   5   0
+        >>> arr.axes._flat_lookup(indices)
+        draw\axis   a   b
+               d0  a0  b2
+               d1  a1  b2
+               d2  a0  b0
+        """
+        from larray.core.array import aslarray, LArray, stack
+
+        flat_indices = aslarray(flat_indices)
+        shape = self.shape
+        divisors = np.roll(np.cumprod(shape[::-1])[::-1], -1)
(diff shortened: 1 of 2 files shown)


gdementen:
@alixdamman I wonder if it is worth a mention in the release notes. I developed this because I know I will need this in LIAM2 at some point (not directly though) but in the general case, I don't think any of our current users cares about this.

larray/random.py
```diff
@@ -379,33 +379,34 @@ def choice(choices=None, axes=None, replace=True, p=None, meta=None):
         Otherwise, if the resulting axes have a shape of, e.g., ``(m, n, k)``, then ``m * n * k`` samples are drawn.
     replace : boolean, optional
         Whether the sample is with or without replacement.
-    p : 1-D array-like, optional
+    p : array-like, optional
         The probabilities associated with each entry in choices.
-        If p is a 1-D LArray, choices are taken from its axis.
+        If p is an LArray, choices are taken from its axes. If p is N dimensional, each cell represents the probability
+        that the combination of labels will occur.
         If not given the sample assumes a uniform distribution over all entries in choices.
```
alixdamman: I wonder if it wouldn't be better to say "If p is a 1-D LArray, choices are taken from its axis. If p is N dimensional, each cell represents the probability that the combination of labels will occur."
I don't know.

larray/random.py
```diff
@@ -446,16 +447,77 @@ def choice(choices=None, axes=None, replace=True, p=None, meta=None):
...
+        else:
+            choices = p.axes[0].labels
+            p = p.data
     if choices is None:
         raise ValueError("choices argument must be provided unless p is an LArray")
     return LArray(np.random.choice(choices, axes.shape, replace, p), axes, meta=meta)
+
+
+# XXX: make this into a method or property on AxisCollection? axes.flat_labels[flat_indices]?
+def axes_flat_lookup(axes, flat_indices):
```
alixdamman: at least move this function to core/axis.py or make it a private method of AxisCollection.

alixdamman:
I also think there is no need to mention it in the release notes.

larray/random.py
gdementen: I hesitated with that too. It seemed redundant, but is probably easier to understand, so ok, I'll change that.

larray/random.py
gdementen: ok

larray/core/axis.py
```diff
@@ -2641,6 +2641,48 @@ def align(self, other, join='outer', axes=None):
...
         return self.replace(axes_changes), other.replace(axes_changes)
 
+    # XXX: make this into a public method/property? axes_col.flat_labels[flat_indices]?
+    def _flat_lookup(self, flat_indices):
+        r"""Return labels corresponding to indices into the flattened axes
+
+        Parameters
+        ----------
+        axes : AxisCollection-like
+            axes to look up into.
```
alixdamman: There is no `axes` argument anymore

gdementen: doh! ☹️

gdementen:
@alixdamman want to review once more or do I just merge this as is? (I only removed the docstring lines)

alixdamman:
Merge

PR #665, opened 2018-07-12
stack(array-likes)

gdementen:


alixdamman:
For what purpose? LGTM anyway.

gdementen:
because I tried to stack numpy arrays (which usually are mostly equivalent to LArrays with anonymous&wildcard axes) but it failed

Issue #652, opened 2018-06-19
remove LArray.insert pos and axis arguments


gdementen:
this is a leftover from our general change from position to index

gdementen:
if we do it, it should happen in 0.29, because I am currently implementing (or rather cleaning up one of the myriad features I have implemented partially in the train) .delete which needs the same argument

alixdamman:
Let's do it

gdementen:
@alixdamman This needs to happen in 0.29, or we defer .delete to 0.30

alixdamman:
Changed to 0.29

gdementen:
now I am wondering if this argument (and axis) shouldn't be simply removed (from both insert and drop). To work using indices one can always use an IGroup:
```python
array.insert(value, before=axis.i[1])
```
and if one does not have the axis available, either the X notation or the string notation can help:
```python
array.insert(value, before=X.axisname.i[1])
array.insert(value, before='axisname.i[1]')
```

alixdamman:
OK. Let's remove pos/indices from insert and drop.

PR #670, opened 2018-07-20
drop

gdementen:
see also #652

### DIFF
--- a/doc/source/changes/version_0_29.rst.inc
+++ b/doc/source/changes/version_0_29.rst.inc
@@ -1,6 +1,8 @@
 ﻿Syntax changes
 --------------
 
+* renamed LArray.drop_labels to ignore_labels to avoid confusion with the new LArray.drop (closes :issue:`672`).
+
 * LArray.insert pos and axis arguments are deprecated because those were only useful for very specific cases and those
   can easily be rewritten by using an indices group (axis.i[pos]) for the "before" argument instead
   (closes :issue:`652`).
--- a/doc/source/changes/version_0_29.rst.inc
+++ b/doc/source/changes/version_0_29.rst.inc
@@ -1,4 +1,4 @@
-﻿Syntax changes
+Syntax changes
 --------------
 
 * new syntax
@@ -136,7 +136,6 @@ New features
     >>> import larray as la
     >>> la.pi
     3.141592653589793
-
 
 * allowed sessions to have metadata.
   Like for arrays, session metadata can be accessed using the syntax ``session.meta.name``.
@@ -329,3 +328,6 @@ Fixes
        a1   3   4   5
 
 * fixed LArray.combine_axes(wildcard=True).
+
+* fixed taking a subset of an array by giving an index along a specific axis using a string (strings
+  like 'axisname.i[pos]').
--- a/doc/source/changes/version_0_29.rst.inc
+++ b/doc/source/changes/version_0_29.rst.inc
@@ -1,4 +1,4 @@
-Syntax changes
+﻿Syntax changes
 --------------
 
 * new syntax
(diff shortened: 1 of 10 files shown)


gdementen:
do not review yet. I forgot to update the release notes :)

alixdamman:
I would prefer to stay closer to pandas than numpy. So I would choose drop instead of delete.

larray/core/array.py
```diff
@@ -5984,6 +5984,98 @@ def expand(v, length):
...
 
+    def drop(self, labels=None, indices=None, axis=None):
+        r"""Return array without some labels or indices along an axis.
+
+        Parameters
+        ----------
+        labels : scalar or Group
+            Label(s) or group to remove.
+        indices : int or list/array of ints, optional
+            Index to remove.
```
alixdamman: s/Index/Index(ices)/?

larray/core/array.py
```diff
@@ -5984,6 +5984,98 @@ def expand(v, length):
...
+        axis : axis reference (int, str or Axis), optional
+            Axis from which to remove labels. This is only required when using `pos` or when labels are valid on
```
alixdamman: by `pos` you mean `indices`?

larray/core/array.py
```diff
@@ -5984,6 +5984,98 @@ def expand(v, length):
...
         return concat(chunks, axis)
 
+    def drop(self, labels=None, indices=None, axis=None):
+        r"""Return array without some labels or indices along an axis.
+
+        Parameters
+        ----------
+        labels : scalar or Group
+            Label(s) or group to remove.
+        indices : int or list/array of ints, optional
```
alixdamman: 1. It is not clear if it is possible to use `lables` and `indices` in the same time. I know from the implementation of the method that the answer is no. I think it should be mentioned clearly that both arguments are exclusive in the doc of the method.
2. Given the examples below, I understand that `indices` can be an IGroup right?

larray/core/array.py
```diff
@@ -5984,6 +5984,98 @@ def expand(v, length):
...
+            Index to remove.
+        axis : axis reference (int, str or Axis), optional
+            Axis from which to remove labels. This is only required when using `pos` or when labels are valid on
+            several axes. Defaults to None, in which case the axis is derived (if present) or guessed from labels.
+
+        Returns
+        -------
+        LArray
+            Array with `labels` removed along `axis`.
```
alixdamman: s/with `labels`/with `labels` or `indices`/

larray/core/array.py
```diff
@@ -5984,6 +5984,98 @@ def expand(v, length):
...
+         a1   4
+
+        when deleting indices instead of labels, one must specify the axis explicitly:
+
+        >>> # arr1.drop('b.i[1]')
+        >>> arr1.drop(b.i[1])
+        a\b  b0  b2  b3
+         a0   0   2   3
+         a1   4   6   7
+
```
alixdamman: Give also an example where you pass an int or a list of int + the axis?

gdementen:
@alixdamman these are all good criticism... But they are also all dependent on the answer to the question I asked in #652

doc/source/changes/version_0_29.rst.inc
```diff
@@ -1,7 +1,8 @@
 ﻿Syntax changes
 --------------
 
-* new syntax
+* LArray.insert pos and axis arguments are deprecated because those were only useful for very specific cases and those
+  can easily be rewritten by using an indices group for the "before" argument instead (closes :issue:`652`).
```
alixdamman: I'm not sure that users know what an "indices group" is. Maybe add something like (axis.i[pos]) after "an indices group".

alixdamman:
Maybe also rename drop_labels as ignore_labels in this PR?

doc/source/changes/version_0_29.rst.inc
gdementen: ok

PR #674, opened 2018-07-27
getsetitem

gdementen:
rewrite of __getitem__ and __setitem__ so that we have much better support for all kinds of advanced indexing

needs a lot more tests & release notes

### DIFF
--- a/doc/source/api.rst
+++ b/doc/source/api.rst
@@ -50,8 +50,8 @@ Searching
    Axis.endingwith
    Axis.matching
 
-Modifying/Selecting/Searching
------------------------------
+Modifying/Selecting
+-------------------
 
 .. autosummary::
    :toctree: _generated/
@@ -69,6 +69,7 @@ Modifying/Selecting/Searching
    Axis.difference
    Axis.align
    Axis.split
+   Axis.ignore_labels
 
 Testing
 -------
--- a/doc/source/api.rst
+++ b/doc/source/api.rst
@@ -26,6 +26,8 @@ Axis.name                   Name of the axis. None in the case of an anonymous a
 :attr:`Axis.labels`         Labels of the axis.
 --------------------------- --------------------------------------------------------------
 :attr:`Axis.labels_summary` Short representation of the labels.
+--------------------------- --------------------------------------------------------------
+:attr:`Axis.dtype`          Data type for the axis labels.
 =========================== ==============================================================
 
 Copying
--- a/doc/source/changes/version_0_30.rst.inc
+++ b/doc/source/changes/version_0_30.rst.inc
@@ -1,0 +1,32 @@
+﻿Syntax changes
+--------------
+
+* new syntax
+
+
+Backward incompatible changes
+-----------------------------
+
+* backward incompatible changes
+
+
+New features
+------------
+
+* added a feature (see the :ref:`miscellaneous section <misc>` for details).
+
+* added another feature.
+
+
+.. _misc:
+
+Miscellaneous improvements
+--------------------------
+
+* improved something.
+
+
+Fixes
+-----
+
+* fixed something (closes :issue:`1`).
--- a/doc/source/changes/version_0_30.rst.inc
+++ b/doc/source/changes/version_0_30.rst.inc
@@ -23,7 +23,82 @@ New features
 Miscellaneous improvements
 --------------------------
 
-* improved something.
+* vastly improved indexing using arrays (of labels, indices or booleans). Many advanced cases did not work, including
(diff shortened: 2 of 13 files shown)


gdementen:
Oops, wait a minute (or two), I forgot to add release notes...

gdementen:
@alixdamman let's hope it is really ready for review this time 😄 

It's awfully hard to describe in the release notes. I don't expect anybody to understand them, but our users occasionally do write code which uses those features. If you have any idea to make this more understandable, feel free...

larray/tests/test_array.py
```diff
@@ -706,108 +706,117 @@ def test_getitem_structured_key_with_groups():
...
-    # replace the values_axis by the extra axis
-    key = LArray(['a1', 'a2', 'a2', 'a1'], [c])
-    assert arr[key].axes == [c, b]
-
-    # 2) key with the values axis (the one being replaced)
-    arr = ndtest([a, b])
-    key = LArray(['b2', 'b1', 'b3'], [b])
+    key = LArray(['a0', 'a1', 'a1', 'a0'], c)
+    # replace the target axis by the extra axis
+    expected = from_string(r"""
```
alixdamman: I'm wondering if this behavior is explained in the tutorial or not? If not, it must be explained.

larray/tests/test_array.py
```diff
@@ -706,108 +706,117 @@ def test_getitem_structured_key_with_groups():
...
+    # TODO: this does not work yet but should be much easier to implement with "align" in make_np_broadcastable
+    # 3bis) key with *part* of another existing axis (not the target axis)
+    # key = LArray(['a1', 'a0'], 'b=b0,b1')
+    # expected = from_string("""
+    # b  b0  b1
+    # \t  3   1""")
+    # assert_array_equal(arr[key], expected)
+
+    # 4) key has both the target axis and another existing axis
+    # TODO: maybe we should make this work without requiring astype!
```
alixdamman: why not passing `dtype=str` to `from_string`? It should work no?

larray/tests/test_array.py
alixdamman: > I think updating the tutorial is out of scope for this PR though.
Sorry, my intention was not to ask you to update the tutorial in this PR but rather to suggest to remember that we need to update the tutorial in the future

larray/tests/test_array.py
gdementen: It's a good idea which I didn't think about... but sadly it does not work:
```python
>>> from_string(r"""
... a\b b0 b1 b2
...  a0 a0 a1 a0
...  a1 a1 a0 a1""", dtype=str).dtype
dtype('O')

>>> from_string(r"""
... a\b b0 b1 b2
...  a0 a0 a1 a0
...  a1 a1 a0 a1""").astype(str).dtype
dtype('<U2')
```
Note that using dtype=np.str_ or dtype='<U2' does not work either.
TypeError: the dtype <U2 is not supported for parsing.

larray/tests/test_array.py
gdementen: No problem, I didn't think that either :)

doc/source/api.rst
```diff
@@ -26,6 +26,8 @@ Axis.name                   Name of the axis. None in the case of an anonymous a
 :attr:`Axis.labels`         Labels of the axis.
 --------------------------- --------------------------------------------------------------
 :attr:`Axis.labels_summary` Short representation of the labels.
+--------------------------- --------------------------------------------------------------
+:attr:`Axis.dtype`          Data type for the axis labels.
```
alixdamman: Add it also for LArray?

doc/source/changes/version_0_30.rst.inc
```diff
@@ -23,10 +23,85 @@ New features
...
+    a\c  c0  c1
+     a0  b0  b1
+     a1  b2  b0
+    >>> arr[key]
+    a\c  c0  c1
+     a0   0   1
+     a1   5   3
+
+    It also works if the key has the target axis (a), another existing axis (b) and an extra axis (c), but this is not
+    shown for brevity.
```
alixdamman: these examples are short, which is a good thing. But we have no clue of how this could be useful in real life (models). 
A great work on the tutorial is waiting us (with meaningful examples).

alixdamman:
See if you want to update the api.rst file before to merge but LGTM. Changelog can be left as it is IMO. But we must update the tutorial (Indexing section) before to release the 0.30 version.

doc/source/api.rst
gdementen: I tried... but it's already there :)

PR #679, opened 2018-08-27
fix #669 : made by() method of Group and Axis classes generate groups with names

### LINKED ISSUE #669: Make Axis.by() generate a list of named groups instead of anonymous groups
@gdementen good idea or bad idea?

alixdamman:
Needs a changelog


larray/core/group.py
```diff
@@ -911,20 +915,31 @@ def by(self, length, step=None):
...
+        >>> age[1:5].by(3, 2, '{start}->{end}')
+        (age.i[1:4] >> '1->3', age.i[3:6] >> '3->5', age.i[5:6] >> '5')
         >>> X.age[[0, 1, 2, 3, 4]].by(2)
-        (X.age[0, 1], X.age[2, 3], X.age[4])
+        (X.age[0, 1] >> '0:1', X.age[2, 3] >> '2:3', X.age[4] >> '4')
         """
+        def make_group(start, length, name_template):
+            g = self[start:start + length]
+            labels = g.eval()
+            if labels[0] != labels[-1]:
```
gdementen: this test is strange. I **suppose** you wanted to test if len(labels) == 1 but that is problematic if g.eval() a single label, or something like that. In that case, please add a comment. Otherwise, revert to a more obvious test.

larray/core/axis.py
```diff
@@ -204,6 +204,10 @@ def by(self, length, step=None):
             length of groups
         step : int, optional
             step between groups. Defaults to length.
+        name_template : str, optional
```
gdementen: Honestly, I am unsure having a name_template argument is worth the maintenance cost. I think we should stop adding features nobody asked for, unless we are sure somebody will use it soon. I **know** I have been guilty of adding many such features myself but...

larray/core/group.py
```diff
@@ -911,20 +915,31 @@ def by(self, length, step=None):
...
+            else:
+                g.name = str(labels[0])
+            return g
         if step is None:
             step = length
-        return tuple(self[start:start + length]
-                     for start in range(0, len(self), step))
+        if name_template is None:
+            name_template = '{start}:{end}'
+        return tuple(make_group(start, length, name_template) for start in range(0, len(self), step))
```
gdementen: I thought #669 was just about having a make_group do something like:

```python
        def make_group(start, length):
            g = self[start:start + length]
            g.name = str(g)
            return g
```
but handling single label groups like you did is probably a good idea.

larray/core/axis.py
```diff
@@ -204,10 +204,6 @@ def by(self, length, step=None, name_template=None):
             length of groups
         step : int, optional
             step between groups. Defaults to length.
-        name_template : str, optional
```
gdementen: hmmmm, now that I've seen some of the code in the reports for demo, I guess this could be useful after all. The question is whether they actually need that particular notation, or if they could do with the default one?

alixdamman:
I think adding a changelog worth it. I will add it.

alixdamman:
Could you check the last commit?

PR #680, opened 2018-08-28
fix #338 : implemented option 'exact' in align()

### LINKED ISSUE #338: implement option 'exact' in align
Instead of aligning, raise ValueError when indexes to be aligned are not equal.
http://xarray.pydata.org/en/stable/generated/xarray.align.html#xarray.align

alixdamman:
Needs changelog

### DIFF
--- a/larray/core/axis.py
+++ b/larray/core/axis.py
@@ -1225,10 +1225,11 @@ class Axis(ABCAxis):
         Axis(['a0', 'a1', 'a2'], 'a')
         >>> axis1.align(axis2, join='right')
         Axis(['a1', 'a2', 'a3'], 'a')
-        >>> axis1.align(axis2, join='exact')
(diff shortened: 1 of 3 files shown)


larray/core/axis.py
```diff
@@ -1201,6 +1205,11 @@ def align(self, other, join='outer'):
             if not isinstance(other, Axis):
                 other = Axis(other)
             return other
+        elif join == 'exact':
+            if not self.equals(other):
+                raise ValueError("{!r} is not equal to {!r}".format(self, other))
```
gdementen: I think the error message should be more precise, ie it should contain enough information to determine that we were trying to do align with join="exact". This will be especially important to understand what happens when .align will be called by another function. I don't think it is the case now but it will in the future.

gdementen:
message in Axis.align still needs to be updated to contain the word "exact", otherwise you have to guess that's what happened

alixdamman:
Could you please just check the last commit (i.e. the new changelog)?

PR #681, opened 2018-08-29
fix 671: implement 'intergerstring[labels]' and 'intergerstring.i[indexes]'

alixdamman:
Needs changelog

### DIFF
--- a/doc/source/changes/version_0_30.rst.inc
+++ b/doc/source/changes/version_0_30.rst.inc
@@ -153,6 +153,26 @@ Miscellaneous improvements
 
   Closes :issue:`669`.
 
+* allowed to specify an axis by its postion when selecting a subset of an array using the string notation:
+
+    >>> pop_mouv = ndtest('geo_from=BE,FR,UK;geo_to=BE,FR,UK')
+    >>> pop_mouv
+    geo_from\geo_to  BE  FR  UK
+                 BE   0   1   2
+                 FR   3   4   5
+                 UK   6   7   8
+    >>> pop_mouv['0[BE, UK]']   # equivalent to pop_mouv[pop_mouv.geo_from['BE,UK']]
+    geo_from\geo_to  BE  FR  UK
+                 BE   0   1   2
+                 UK   6   7   8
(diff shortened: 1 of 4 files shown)


alixdamman:
Because of PR #682, I had to update the current PR. So, it could be no longer "LGTM".

larray/core/axis.py
```diff
@@ -2312,7 +2312,16 @@ def _translate_axis_key(self, axis_key, bool_passthrough=True):
...
+        # Need to convert string keys to groups otherwise command like
+        # >>> ndtest((5, 5)).drop('1[a0]')
+        # will work although it shouldn't
+        if isinstance(axis_key, basestring):
+            key = _to_key(axis_key)
+            if isinstance(key, Group):
+                axis_key = key
+
         if isinstance(axis_key, ExprNode):
+            # ALD: Exception('yada') --> What?
```
gdementen: Oops, this obviously should never have been committed. ;-). This was a test to see if this code path was actually used. It seems like either it is not used, or there is no test exercising it... The idea was to check which option it is before either removing the code or adding tests... and it seems like I forgot to do that.
FWIW, the same comment applies a few lines below for the other "yada"

alixdamman: Do I remove these lines then?

gdementen: no, it would pollute your PR ;-). And I'd like to actually check the situation before removing them. I've opened an issue to avoid forgetting it: #711.

doc/source/changes/version_0_30.rst.inc
```diff
@@ -139,6 +139,25 @@ Miscellaneous improvements
...
+                 BE   0   1   2
+                 FR   3   4   5
+                 UK   6   7   8
+    >>> pop_mouv['0[BE, UK]']   # equivalent to pop_mouv[pop_mouv.geo_from['BE,UK']]
+    geo_from\geo_to  BE  FR  UK
+                 BE   0   1   2
+                 UK   6   7   8
+    >>> pop_mouv['0.i[0, 2]']   # equivalebt to pop_mouv[pop_mouv.geo_from.i[0, 2]]
+    geo_from\geo_to  BE  FR  UK
+                     BE   0   1   2
```
gdementen: hmm, the indentation of these lines seems wrong...

PR #687, opened 2018-09-12
fix issue 686: added xlrd and openpyxl in binder/environment.yml

alixdamman:


gdementen:
Unsure it's worth reviewing this kind of PR... Just apply it.

PR #691, opened 2018-09-14
bump version to 0.30-dev

gdementen:
PR with just the version bump so that we can accept it quickly and I can have clean release notes in my other PRs


doc/source/changes/version_0_30.rst.inc
```diff
@@ -0,0 +1,32 @@
+﻿Syntax changes
+--------------
```
alixdamman: I remember that you changed the template for the release notes in a (very) recent PR. You may want to use the new template?

PR #695, opened 2018-09-18
added python 3.7 to travis CI

gdementen:


gdementen:
apparently pytest-pep8 is not yet available for python3.7

alixdamman:
So, we need to add something like

```bash
if [[ "$TRAVIS_PYTHON_VERSION" == "3.7" ]]; then
        conda install pytest-pep8
fi
```

gdementen:
duh? this will fail too?!?

gdementen:
@alixdamman thanks for the info :)

PR #696, opened 2018-09-18
close #237: implemented LArray.equals(check_axes=False)

### LINKED ISSUE #237: implement larray_equiv and larray_nan_equiv

gdementen:
I still wonder if LArray.equiv wouldn't be better... or if check_axes=True isn't the better default value.

### DIFF
--- a/larray/core/array.py
+++ b/larray/core/array.py
@@ -5081,23 +5081,27 @@ class LArray(ABCLArray):
         return self.data.__float__()
 
     @deprecate_kwarg('nan_equals', 'nans_equal')
-    def equals(self, other, rtol=0, atol=0, nans_equal=False):
-        """
+    def equals(self, other, rtol=0, atol=0, nans_equal=False, check_axes=False):
+        r"""
         Compares self with another array and returns True if they have the same axes and elements, False otherwise.
 
         Parameters
         ----------
-        other: LArray-like
+        other : LArray-like
             Input array. aslarray() is used on a non-LArray input.
         rtol : float or int, optional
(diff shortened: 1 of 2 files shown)


alixdamman:
IMO, adding a `LArray.equiv()` method is definitively not a good idea. You can set check_axes=True as default value if you want.

gdementen:
I don't really want anything. I am just unsure. I can see some users complaining either way.

gdementen:
@alixdamman I couldn't find any mention of .equals in users' code, so I guess it won't be a problem. Ok to merge this PR as-is?

larray/core/array.py
```diff
@@ -5081,23 +5081,27 @@ def __float__(self):
...
         atol : float or int, optional
             The absolute tolerance parameter (see Notes). Defaults to 0.
-        nans_equal: boolean, optional
+        nans_equal : boolean, optional
             Whether or not to consider nan values at the same positions in the two arrays as equal.
             By default, an array containing nan values is never equal to another array, even if that other array
             also contains nan values at the same positions. The reason is that a nan value is different from
             *anything*, including itself. Defaults to False.
+        check_axes : boolean, optional
+            Whether or not to check that the set of axes and their order is the same on both sides. Defaults to True.
```
alixdamman: s/Defaults to True/Defaults to False

gdementen: nice catch ! That's what you get for inverting the default at the last minute :)

alixdamman: or working without specifications

PR #699, opened 2018-09-18
misc small docstring fixes

gdementen:


larray/inout/csv.py
```diff
@@ -34,12 +34,12 @@ def read_csv(filepath_or_buffer, nb_axes=None, index_col=None, sep=',', headerse
     ----------
     filepath_or_buffer : str or any file-like object
         Path where the csv file has to be read or a file handle.
-    nb_axes : int, optional
+    nb_axes : int or None, optional
```
alixdamman: I guess this also need to be changed in read_excel?

larray/inout/csv.py
```diff
@@ -34,12 +34,12 @@ def read_csv(filepath_or_buffer, nb_axes=None, index_col=None, sep=',', headerse
...
         Number of axes of output array. The first ``nb_axes`` - 1 columns and the header of the CSV file will be used
         to set the axes of the output array. If not specified, the number of axes is given by the position of the
         first column header including a ``\`` character plus one. If no column header includes a ``\`` character,
         the array is assumed to have one axis. Defaults to None.
-    index_col : list, optional
+    index_col : list or None, optional
```
alixdamman: idem

gdementen:
Well, it needs to be changed in a lot of other places indeed but I wouldn't bother at this point. These changes are in fact needed for my "quickbar" branch which inspects docstrings to get expected types of arguments and is a bit picky. I was using read_csv as my guinea pig. Haven't worked on that for months now but the goal of this PR is in fact just to upstream one of my branches so that I have one less branch to manage...

alixdamman:
OK. One day, we may use the type hints for the quickbar.

PR #710, opened 2018-10-22
added check_api_doc.py module in doc/source

alixdamman:
This is to be discussed but it would be nice to have something (even if ad-hoc) to check if something is missing in the file api.rst

### DIFF
--- a/doc/source/check_api_doc.py
+++ b/doc/source/check_api_doc.py
@@ -1,0 +1,396 @@
+from __future__ import print_function
+
+import sys
+import os
+import inspect
+import importlib
+import warnings
+
+from sphinx.util.inspect import safe_getattr
+
+_current_dir = os.path.dirname(__file__)
+_lib_path = os.path.abspath(os.path.join(_current_dir, '..', '..', 'larray'))
+_outdir = os.path.abspath(os.path.join(_current_dir, '..', 'build', 'check_api'))
+
+sys.path.insert(0, _lib_path)
+from larray import __version__
+
+_modules = ['larray', 'larray.random', 'larray.core.constants']
+_exclude = ['absolute_import', 'division', 'print_function', 'renamed_to']
+
+
+def _is_deprecated(obj):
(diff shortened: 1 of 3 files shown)


doc/source/check_api_doc.py
```diff
@@ -0,0 +1,112 @@
...
+    def add_item(item, api_doc_items):
+        item = item.astext().strip().split()
+        if len(item):
+            if isinstance(item, list):
+                item = item[0]
+            api_doc_items.append(item)
+
+    api_doc_items = []
+    directives.register_directive('autosummary', Autosummary)
+    cleanup_line = lambda line: line.replace(':attr:', '      ').replace('`', ' ')
```
gdementen: I don't really like lambdas... Yes they save one line, but they are less readable. Not worth it IMO.

alixdamman:
I found weird that users could modify the data on an array without any check on the axes and reversely. 
But I'm aware that implementing properties could impact the general performance of LArray objects. 
I'm OK to move the last 2 commits in another branch (PR).

gdementen:
shouldn't we run that script automatically somewhere during the release process ?

gdementen:
I don't know. What are the pro and con?

alixdamman:
Sounds like Numpy 1.16 is incompatible with the latest release of pytables : https://stackoverflow.com/questions/54210073/pd-read-df-throws-cannot-set-writable-flag-to-true-of-this-array

This makes unittests testing read and write files in HDF format to fail

setup.py
```diff
@@ -15,7 +15,8 @@ def readlocal(fname):
 DESCRIPTION = "N-D labeled arrays in Python"
 LONG_DESCRIPTION = readlocal("README.rst")
 SETUP_REQUIRES = []
-INSTALL_REQUIRES = ['numpy >= 1.13', 'pandas >= 0.13.1']
+# excludes numpy 1.16.* because it is incompatible with pytables 3.4
```
gdementen: AFAICT, it dates back much further than 3.4. I would write something like "pytables < 3.5 (which is not released yet)"

Issue #716, opened 2018-11-21
Test if ufunc interp works with LArray objects and document it properly if answer is yes


alixdamman:
Should appear in the API doc if it actually works with LArrays, which I have never tested (and I doubt is the case). Might be worth having specific documentation if it works well. My guess is that we should rather make a new LArray method for that one.

PR #717, opened 2018-11-22
include missing ufuncs doc to API documention

alixdamman:
readthedocs output is here: https://larray-test.readthedocs.io/en/documentation/api.html#utility-functions

I still need to find a way to run new specific doctests for ``where``, ``maximum`` and ``minimum`` but let's start the discussion.

Issues #715 and #716 will be fixed in another PR.

### DIFF
--- a/doc/source/api.rst
+++ b/doc/source/api.rst
@@ -444,17 +444,6 @@ Miscellaneous
    LArray.shift
    LArray.diff
    LArray.to_clipboard
-   round
-   floor
-   ceil
-   trunc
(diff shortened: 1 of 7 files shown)


gdementen:
BTW, could you make a separate PR with only the first commit? (and merge it directly, or even push directly to master without making a PR). I'd like to have that commit merged rather sooner than later.

alixdamman:
Of course, the changelog associated with issue 698 will need to be changed after fixing issue #723

PR #719, opened 2018-11-23
updated documentation of with_total (issue 693)

alixdamman:


larray/core/array.py
```diff
@@ -2718,30 +2718,48 @@ def with_total(self, *args, **kwargs):
 
         Examples
         --------
-        >>> arr = ndtest((3, 3))
+        >>> arr = ndtest("age=0..2;time=2013..2016")
```
gdementen: if you want to use something different than the default ndtest (which I am not 100% sure is better in the long run -- **we need to discuss this**), I would rather use sex and time, because having an age between 0 and 2 seems odd to me.

larray/core/array.py
```diff
@@ -2718,30 +2718,48 @@ def with_total(self, *args, **kwargs):
...
-             a1   3   4   5       60
-             a2   6   7   8      336
-        product   0  28  80        0
+        age\time  2013  2014  2015  2016  total
+               0     0     1     2     3      6
+               1     4     5     6     7     22
+               2     8     9    10    11     38
+           total    12    15    18    21     66
+
+        Choosing aggregate function
```
gdementen: s/Choosing/Using another/

gdementen: s/function/function and label/

larray/core/array.py
```diff
@@ -2718,30 +2718,48 @@ def with_total(self, *args, **kwargs):
...
+
+        >>> arr.with_total(op=mean, label='mean')
+        age\time  2013  2014  2015  2016  mean
+               0   0.0   1.0   2.0   3.0   1.5
+               1   4.0   5.0   6.0   7.0   5.5
+               2   8.0   9.0  10.0  11.0   9.5
+            mean   4.0   5.0   6.0   7.0   5.5
+
+        Specify an axis
```
gdementen: s/Specify/Specifying (to be consistent with the rest of the examples)

PR #721, opened 2018-11-23
issue 703 : updated behavior of pattern argument of Session.filter

alixdamman:


gdementen:
the changelog is not explicit enough for what people who used the .filter method need to do (add * to the end of their patterns)

PR #732, opened 2018-12-11
implemented set_options() (issue 274)

alixdamman:


larray/core/array.py
```diff
@@ -2407,7 +2422,9 @@ def as_table(self, maxlines=None, edgeitems=5, light=False, wide=True, value_nam
                 yield list(tick) + dataline.tolist()
 
     def dump(self, header=True, wide=True, value_name='value'):
-        """Dump array as a 2D nested list
+        """dump(header=True, wide=True, value_name='value')
```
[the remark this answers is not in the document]
alixdamman: It is not. I first included ``wide`` in the global options but then changed my mind. Not sure it is a good idea.

larray/util/misc.py
```diff
@@ -131,8 +136,9 @@ def get_min_width(table, index):
     return max(longest_word(row[index]) for row in table)
 
 
+# XXX: what fullinfo is used for and when ?
```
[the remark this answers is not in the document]
alixdamman: The ``fullinfo`` was already there in the "initial commit" =D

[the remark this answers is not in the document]
gdementen: hehe 😄 This is a leftover from the code copied from LIAM2 then. Now that I think of LIAM2, I think this was because this code is used when outputting to .csv files in LIAM2 and we want all the precision we can get in that case.

larray/util/misc.py
```diff
@@ -81,18 +81,23 @@ def prod(values):
...
 
 
-def format_value(value, missing, fullinfo=False):
+def format_value(value, missing, fullinfo=False, precision=None):
     if isinstance(value, float) and not fullinfo:
         # nans print as "-1.#J", let's use something nicer
         if value != value:
             return missing
+        elif precision is not None:
+            return '{:2.{precision}f}'.format(value, precision=precision).rstrip('0').rstrip('.')
```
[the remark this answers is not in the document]
alixdamman: concerning ``table2str`` and ``format_value``, why did you choose to not rely on ``np.array2string`` for printing the internal data? 
Because we could benefit of all printing options provided by ``np.set_printoptions``.

larray/util/misc.py
```diff
@@ -131,20 +133,27 @@ def get_min_width(table, index):
...
+def table2str(table, missing, summarize=True, maxwidth=None, numedges='auto', sep='  ', cont='...',
+              keepcols=0, precision=None):
     """
     table is a list of lists
     :type table: list of list
     """
     if not table:
         return ''
+
+    from larray.util.options import OPTIONS, DISPLAY_WIDTH, DISPLAY_PRECISION
```
gdementen: I would much prefer if this (taking the default value into account) was handled in `LArray.__str__` instead

larray/core/array.py
```diff
@@ -2355,6 +2360,12 @@ def as_table(self, maxlines=None, edgeitems=5, light=False, wide=True, value_nam
         if not self.ndim:
             return
 
+        # get default options
+        if maxlines is None:
+            maxlines = OPTIONS[DISPLAY_MAXLINES]
```
gdementen: I wonder if it wouldn't be better to handle this in `LArray.__str__` instead

larray/util/misc.py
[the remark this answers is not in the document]
gdementen: yes, but IIRC (I wrote this a **long** time ago), it was simpler not to use it because combining with the headers wasn't that easy I think. Especially computing the maximum width of each column (including its header !!) of the 2D "table" was problematic when using np.array2string. IIRC, I also could not find an option to remove the `[` `]` from the string that numpy produces. Now if we support more of the options of set_printoptions that numpy has, it **might** be worth it but currently I don't think it is.

larray/core/array.py
```diff
@@ -2296,19 +2298,23 @@ def __iter__(self):
...
         maxlines : int, optional
-            Maximum number of lines to show.
+            Maximum number of lines to show. If negative all lines are shown.
+            Defaults to -1.
         edgeitems : int, optional
             If number of lines to display is greater than `maxlines`,
             only the first and last `edgeitems` lines are displayed.
-            Only active if `maxlines` is not None.
-            Equals to 5 by default.
+            Only active if `maxlines` is not 0.
```
gdementen: is not -1.

larray/core/array.py
```diff
@@ -2388,7 +2394,7 @@ def as_table(self, maxlines=None, edgeitems=5, light=False, wide=True, value_nam
         other_colnames = self.axes[-1].labels.tolist() if wide else [value_name]
         yield axes_names + other_colnames
         # summary if needed
-        if maxlines is not None and height > maxlines:
+        if maxlines > 0 and height > maxlines:
```
gdementen: either `>= 0` or `!= -1`

larray/core/array.py
```diff
@@ -2296,19 +2298,23 @@ def __iter__(self):
...
-        """
+    def as_table(self, maxlines=-1, edgeitems=5, light=False, wide=True, value_name='value'):
+        r"""
         Generator. Returns next line of the table representing an array.
 
         Parameters
         ----------
         maxlines : int, optional
-            Maximum number of lines to show.
+            Maximum number of lines to show. If negative all lines are shown.
```
gdementen: I wouldn't bother supporting other negative numbers, so I would write:
`Defaults to -1 (all lines are shown).`

larray/core/array.py
```diff
@@ -2296,19 +2298,23 @@ def __iter__(self):
...
+            Defaults to -1.
         edgeitems : int, optional
             If number of lines to display is greater than `maxlines`,
             only the first and last `edgeitems` lines are displayed.
-            Only active if `maxlines` is not None.
-            Equals to 5 by default.
+            Only active if `maxlines` is not 0.
+            Defaults to 5.
+        light: bool, optional
```
gdementen: missing space before :

doc/source/changes/version_0_30.rst.inc
```diff
@@ -217,6 +217,42 @@ Miscellaneous improvements
...
+     a\b                  b0                  b1  ...                 b98                 b99
+      a0                 0.0  0.3333333333333333  ...  32.666666666666664                33.0
+      a1  33.333333333333336  33.666666666666664  ...                66.0   66.33333333333333
+     ...                 ...                 ...  ...                 ...                 ...
+    a498             16600.0  16600.333333333332  ...  16632.666666666668             16633.0
+    a499  16633.333333333332  16633.666666666668  ...             16666.0  16666.333333333332
+
+    Or to set global options:
+
+    >>> set_options(display_maxlines=10, display_precision=2) # doctest: +SKIP
```
gdementen: shouldn't we explain what the line does?

doc/source/changes/version_0_30.rst.inc
```diff
@@ -217,6 +217,42 @@ Miscellaneous improvements
...
 
 
+* added :py:obj:`set_options` allowing to set options for larray within a ``with`` block or globally:
+
+    >>> from larray import *
+    >>> arr = ndtest((500, 100), dtype=float) / 3
+
+    You can use ``set_options`` either as a context manager:
+
+    >>> with set_options(display_width=100, display_edgeitems=2):
```
gdementen: shouldn't we explain what the line does?

alixdamman:
We may also want to have something similar to the "option" functions as in pandas: 

- ``get_option()`` / ``set_option()``: get/set the value of a single option.
- ``reset_option()``: reset one or more options to their default value.
- ``describe_option()``: print the descriptions of one or more options.
- ``option_context()``: execute a codeblock with a set of options that revert to prior settings after execution.

See https://pandas.pydata.org/pandas-docs/stable/options.html#overview

alixdamman:
Seems like Travis changed its procedure. Looking at last Travis logs, I can see:
```
  git clone --depth=50 https://github.com/larray-project/larray.git larray-project/larray
  cd larray-project/larray
  git fetch origin +refs/pull/732/merge:
  git checkout -qf FETCH_HEAD
```

So, Travis will continue to fail until I rebase my own branch on upstream:master and then force push.

PR #740, opened 2019-02-04
fix #702 : fixed bug when writing metadata using HDF format

### LINKED ISSUE #702: bug when writing metadata
fix is:
```python
def _dump_metadata(self, metadata):
    root = self.handle._handle.root
    root._v_attrs.metadata = metadata
```

alixdamman:


alixdamman:
Maybe linked to the version used of pytables ?

gdementen:
I think it's more likely the version of pandas. But the first thing to do in any case when fixing a bug is  reproduce it. Otherwise you don't know whether you actually fixed it or not. If that's indeed related to a lib version, we don't need to add any extra test but at least we could document it and **possibly** blacklist that particular version.

PR #756, opened 2019-04-09
(issue 563): bypass cartesian_product_df() when reading a saved LArray object

alixdamman:


larray/inout/hdf.py
```diff
@@ -74,7 +74,7 @@ def read_hdf(filepath_or_buffer, key, fill_value=nan, na=nan, sort_rows=False, s
         _meta = attrs.metadata if 'metadata' in attrs else None
         if _type == 'Array':
             res = df_aslarray(pd_obj, sort_rows=sort_rows, sort_columns=sort_columns, fill_value=fill_value,
-                              parse_header=False)
+                              parse_header=False, dense=True)
```
gdementen: only use dense=True if _type has an explicit value, not if it was missing, otherwise we can't load a dataframe saved by pandas without larray

doc/source/changes/version_0_30.rst.inc
```diff
@@ -221,6 +221,8 @@ Miscellaneous improvements
 * renamed `a_min` and `a_max` arguments of :py:obj:`LArray.clip()` as `minval` and `maxval` respectively
   and made them optional (closes :issue:`747`).
 
+* improved speed of :py:obj:`read_hdf()` function when reading a stored LArray object (closes :issue:`563`).
```
gdementen: You should add that one must load and re-save their files to benefit for this !

PR #757, opened 2019-04-11
fix #335 : Standardize nan to NaN in all docstrings. 

### LINKED ISSUE #335: standardize nan spelling in docstrings
In the output, we use `nan` (this is specified in `LArray.__str__`). For the rest (parameters descriptions), we currently use a mix of `nan` and `NaN`.

alixdamman:
I chose the same logic as Numpy: use `NaN` in docstrings and `nan` when printing values.

Pandas chose to use NaN everywhere.

### DIFF
--- a/larray/core/array.py
+++ b/larray/core/array.py
@@ -1211,7 +1211,7 @@ class LArray(ABCLArray):
             b2    5.0
         Name: my_name, dtype: float64
 
-        Drop nan values
+        Drop NaN values
 
         >>> arr['b1'] = nan
         >>> arr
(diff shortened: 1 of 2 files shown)


gdementen:
otherwise LGTM (and should not even require a review)

gdementen:
Is there anything left to do before you can merge this?

alixdamman:
I am still not sure about the strategy to adopt with the NaN in the documentation. I prefer to move this issue to the 0.31 milestone

gdementen:
What's the problem exactly?

alixdamman:
If we except the doctests, both Numpy and Pandas use `NaN` in their documentation.
When it comes to display the "NaN" values, Numpy prints ``nan`` and Pandas ``NaN``.

In our case, since the raw data are stored in Numpy ndarray, I think we should the same strategy as Numpy.

PR #760, opened 2019-04-15
issue 759 : added script to include editor changelogs

alixdamman:


doc/source/merge_changelogs.py
```diff
@@ -0,0 +1,84 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+"""
+Append release notes from other projects of the larray-project constellation to the current changelog
+"""
+
+import os
+import requests
```
gdementen: move this below, this is 3rd party lib, not a a python stdlib (even if it should ;-))

doc/source/merge_changelogs.py
```diff
@@ -0,0 +1,84 @@
...
+def short(rel_name):
+    return rel_name[:-2] if rel_name.endswith('.0') else rel_name
+
+
+def relname2fname(release_name):
+    short_version = short(release_name)
+    return r"version_{}.rst.inc".format(short_version.replace('.', '_'))
+
+
+def yes(msg, default='y'):
```
gdementen: Can't we reuse those 3 functions (and possibly others) from some common lib? (or officially moved them in larray "core" and make other projects use that here???). I never actually thought this through (!!!) so I don't know what is the best solution, but I always wanted and still would like to avoid duplicating code between the larray projects...

doc/source/merge_changelogs.py
```diff
@@ -0,0 +1,84 @@
...
+          + '/{}/{}/'.format(branch, rel_changes_dir) + changelog_file
+    req = requests.get(url)
+    if req.status_code != requests.codes.ok:
+        raise ValueError("Content at URL {} cound not been found.".format(url))
+    external_changelog_content = req.text
+
+    # dump external changelogs
+    changelog_file = os.path.abspath(os.path.join('.', 'changes', changelog_file))
+    with open(changelog_file) as f:
+        lines = f.readlines()
```
gdementen: Is there a reason why you use .readlines(), splitlines() and writelines() instead of just using .read() and .write()? If so, please add a comment.

doc/source/merge_changelogs.py
```diff
@@ -0,0 +1,84 @@
...
+    with open(changelog_file, 'w', encoding="utf-8") as f:
+        new_section = """
+
+{section_name}
+{underline}
+
+""".format(section_name=section_name.upper(), underline='-' * len(section_name))
+        lines += new_section.splitlines(True)
```
gdementen: If you need to keep .splitlines (see comment above), please use .splitlines(keepends=True), as it is more readable.

make_release.py
```diff
@@ -78,18 +86,21 @@ def announce_new_release(release_name):
         print("update conda-forge feedstock steps:", ', '.join(f.__name__ for f, _ in update_feedstock_steps))
         print("or")
         print("Usage: {} -a|--announce release_name".format(argv[0]))
-        print("Usage: {} -m release_name".format(argv[0]))
+        print("Usage: {} -m|--meta release_name".format(argv[0]))
+        print("Usage: {} -l|--logs release_name".format(argv[0]))
```
gdementen: I am not a fan of this proliferation of commands. Can't this merge changelog be one normal "step" of the release process? I mean, is it realistic to do this step within the usual sequence of release steps, or will does it **need** to be a totally separate command? I did **not** think it through. Just asking the question.

alixdamman:
Not sure where to insert the new function `merge_changelogs` ? Before check_local_repo ?

alixdamman:
I am not sure to completely understand your comment. 
Do mean we should have a second directory 'core_changes' containing log files with core logs only under 'doc/source' and then merging the log files from the different projects into a new log file generated in 'doc/source/changes' ?

gdementen:
LGTM. Need anything else?

alixdamman:
Merging the PR https://github.com/larray-project/larray-editor/pull/177 ?

PR #766, opened 2019-05-24
documentation: updated readthedocs.yml to the version 2 + updated python version to 3.6

alixdamman:
It is always a great to discover that Readthedocs made some internal changes a few time ago and that changes break all your build attempts until you implement the necessary updates =D

The present commit seems to solve the problem:
https://readthedocs.org/projects/larray-test/builds/9121630/

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -86,6 +86,9 @@ For IO (HDF, Excel)
 
 - `pytables <http://www.pytables.org/>`__:
   for working with files in HDF5 format.
+- `xlwings <https://www.xlwings.org/>`__:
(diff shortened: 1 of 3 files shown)


gdementen:
btw, I notice that the changelog for past versions is broken in this test documentation. Probably a result of change the underline character in the changelog for 0.30. Will this be ok in the final release?

alixdamman:
After we merge the logs from the core and the editor project, the logs for the previous versions seems OK: https://larray-test.readthedocs.io/en/documentation/index.html

gdementen:
Okidoki!

PR #768, opened 2019-05-28
fixed dependency incompatibilities in larrayenv

alixdamman:


.travis.yml
```diff
@@ -45,7 +45,7 @@ install:
     # might want to avoid the later by installing all dependencies manually
     # except scipy and install pandas with --no-deps
     - conda create -n travisci --yes python=${TRAVIS_PYTHON_VERSION:0:3}
-            "numpy>=1.13,!=1.16.*" pandas pytables matplotlib xlrd openpyxl
```
gdementen: * I think the "numpy != 1.16" restriction could be lifted now that pytables 3.5 has been released
* It is usually a good practice to add a comment why we bump some package requirements, so that we know when we can lift the restriction (I know also forget to do that myself 😉). So why do you add pytables >= 3.4, pyqt >=5.6 and xlwings >= 0.11?

alixdamman:
I started to use ``==`` instead of ``>=`` but then I realized that because of the Python 3.5 support, we are limited to quite old versions of the libraries. It is not a big deal but we really need to drop Python 3.5 in a short term.

PR #771, opened 2019-06-07
(issue 744): use pythonic notation in doc of aggregation functions

alixdamman:


larray/example.py
```diff
@@ -46,6 +47,7 @@ def load_example_data(name):
         Example data to load. Available example datasets are:
 
         - demography
+        - demography_small
```
gdementen: shouldn't that be demography_eurostat?

larray/example.py
```diff
@@ -55,20 +57,25 @@ def load_example_data(name):
...
+    >>> print(demo.summary())   # doctest: +NORMALIZE_WHITESPACE
+    hh: time, geo, hh_type (26 x 3 x 7) [int64]
+    pop: time, geo, age, sex, nat (26 x 3 x 121 x 2 x 2) [int64]
+    qx: time, geo, age, sex, nat (26 x 3 x 121 x 2 x 2) [float64]
+    >>> demo = load_example_data('demography_eurostat')
+    >>> print(demo.summary())   # doctest: +NORMALIZE_WHITESPACE
+    Metadata:
+       title: Demographic datasets for a small selection of countries in Europe
+       source: demo_jpan, demo_fasec, demo_magec and demo_marcz tables from Eurostat
+    citizenship: citizenship ['TOTAL' 'NAT' 'FOR' 'STLS' 'UNK'] (5)
```
gdementen: WTF do these codes mean? Apparently the marriages array is bringing that in. Unsure it's worth it to have it and if we keep it the labels should be renamed to something more understandable.

larray/example.py
```diff
@@ -55,20 +57,25 @@ def load_example_data(name):
...
+    country: country ['Belgium' 'France' 'Germany'] (3)
+    gender: gender ['Male' 'Female'] (2)
+    time: time [2013 2014 2015] (3)
+    even_years: time[2014] >> even_years (1)
+    odd_years: time[2015] >> odd_years (1)
```
gdementen: It seems odd (pun intended) => a bug? that 2013 is not in there.

larray/example.py
```diff
@@ -55,20 +57,25 @@ def load_example_data(name):
...
+    hh: time, geo, hh_type (26 x 3 x 7) [int64]
+    pop: time, geo, age, sex, nat (26 x 3 x 121 x 2 x 2) [int64]
+    qx: time, geo, age, sex, nat (26 x 3 x 121 x 2 x 2) [float64]
+    >>> demo = load_example_data('demography_eurostat')
+    >>> print(demo.summary())   # doctest: +NORMALIZE_WHITESPACE
+    Metadata:
+       title: Demographic datasets for a small selection of countries in Europe
+       source: demo_jpan, demo_fasec, demo_magec and demo_marcz tables from Eurostat
+    citizenship: citizenship ['TOTAL' 'NAT' 'FOR' 'STLS' 'UNK'] (5)
+    country: country ['Belgium' 'France' 'Germany'] (3)
```
gdementen: The representation for axes in .summary seems odd. Why do we repeat the axis name? Aslo, couldn't we make that clearer by grouping objects by type. eg:
```
Metadata:
    ...
Arrays:
    ...
Axes:
    ...
Groups:
    ...
```

PR #772, opened 2019-06-11
larrayenv metapackage: updated restriction on ipykernel version

alixdamman:


gdementen:
assuming it does what I think it does, this LGTM 😉

PR #775, opened 2019-06-12
issue 754 : added 'update' method to the Session class

alixdamman:
needs changelog

### DIFF
--- a/larray/core/session.py
+++ b/larray/core/session.py
@@ -146,10 +146,6 @@ class Session(object):
             If keyword arguments are specified, the session is then updated with those key/value pairs
             (e.g.: ses.update(pop=pop, births=births, deaths=deaths)).
 
-        Returns
-        -------
-        None
-
         Examples
         --------
         >>> x, y = Axis('x=x0..x2'), Axis('y=y0..y3')
--- a/larray/core/session.py
+++ b/larray/core/session.py
@@ -115,7 +115,7 @@ class Session(object):
         *args : list of object
             Objects to add. Objects must have an attribute 'name'.
         **kwargs : dict of {str: object}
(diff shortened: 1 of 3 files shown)


larray/core/session.py
```diff
@@ -132,6 +132,103 @@ def add(self, *args, **kwargs):
...
+        arr2: z (3) [int32]
+        z: z ['z0' 'z1' 'z2'] (3)
+        arr3: x, z (3 x 3) [int32]
+        """
+        if other is None:
+            pass
+        elif hasattr(other, 'items'):
+            for k, v in other.items():
+                self[k] = v
+        elif hasattr(other, '__getitem__'):
```
gdementen: seems like a strange way to test other is an iterable. Is there a reason not to use isinstance(other, Iterable) ?

alixdamman: No, there is no good reason. I'll change that.

larray/core/session.py
```diff
@@ -132,6 +132,103 @@ def add(self, *args, **kwargs):
...
+
+        Parameters
+        ----------
+        other: Session or dict-like object or iterable with key/value pairs
+            Object containing key/value pairs to add or modify.
+        **kwargs:
+            If keyword arguments are specified, the session is then updated with those key/value pairs
+            (e.g.: ses.update(pop=pop, births=births, deaths=deaths)).
+
+        Returns
```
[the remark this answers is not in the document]
alixdamman: Seems you're right. Numpy doesn't specify a `Returns` section if the operation is made inplace ( https://docs.scipy.org/doc/numpy-1.12.0/reference/generated/numpy.ufunc.at.html )

gdementen:
LGTM, modulo changelog

PR #777, opened 2019-06-19
(issue 712): allow reindex(new_axis)

alixdamman:


larray/core/array.py
```diff
@@ -1374,71 +1376,104 @@ def reindex(self, axes_to_reindex=None, new_axis=None, fill_value=nan, inplace=F
...
         a1   c0   2   1
         """
         # XXX: can't we move this to AxisCollection.replace?
+        if isinstance(axes_to_reindex, basestring):
+            try:
+                axes_to_reindex = Axis(axes_to_reindex)
+            except:
+                pass
+        if isinstance(axes_to_reindex, (Group, Axis)):
+            new_axis = axes_to_reindex.copy() if isinstance(axes_to_reindex, Axis) else Axis(axes_to_reindex)
```
gdementen: why do you need the copy() here? I guess there is a good reason, but since I fail to see it, adding a comment wouldn't hurt 😉

alixdamman: I tried something different at first and I forgot to remove the .copy()

PR #778, opened 2019-06-19
(issue 477) : use raw docstrings (to avoid having to escape \)

### LINKED ISSUE #477: use raw docstrings (r"""bla""") to avoid having to escape \
We need to test it then decide whether we use it everywhere for consistency or only where necessary.

Special mention for LArray and AxisCollection.split_axes, from_list and from_string which need quadruple \ !

alixdamman:


larray/core/array.py
```diff
@@ -498,8 +498,8 @@ def _doc_agg_method(func, by=False, long_name='', action_verb='perform', extra_a
 
     doc_args = "".join(_arg_agg[arg] for arg in extra_args)
     doc_kwargs = "".join(_kwarg_agg[kw]['doc'] for kw in kwargs)
-    doc_varargs = """
-        \*axes_and_groups : None or int or str or Axis or Group or any combination of those
+    doc_varargs = r"""
+        *axes_and_groups : None or int or str or Axis or Group or any combination of those
```
gdementen: ??

PR #789, opened 2019-08-05
(issue 788) convert unicode string to bytes and reversely when working wiht HDF files

alixdamman:


alixdamman:
needs a changelog?

gdementen:
For the changelog, if the only consequence is to make sessions with "unicode axes" (or groups) smaller, then probably not (because those or not used yet AFAIK). It is also has impacts on arrays with those axes, then yes.

PR #790, opened 2019-08-06
updated dependencies requirements

alixdamman:
- removed exclusion of numpy==1.16.* version since pytables 3.5 is now available and fixed the incompatibility
- set condition pytables>=3.5 in travis.yml and larrayenv metapackage
- dropped support Python 3.5 in travis.yml
- updated travis.yml for Python 3.7 (both "dist: xenial" and "sudo: true" are the default now)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,20 +1,13 @@
 # References
 # - https://gist.github.com/dan-blanchard/7045057
 # - https://github.com/numba/numba/blob/master/.travis.yml
+dist: xenial         # required for Python >= 3.7
 language: python
 
 python:
     - "2.7"
-    - "3.5"
     - "3.6"
-
-# Work-around to enable python 3.7 without globally enabling sudo and dist: xenial for other build jobs
-# see https://github.com/travis-ci/travis-ci/issues/9815
-matrix:
-  include:
-    - python: 3.7
-      dist: xenial
-      sudo: true
+    - "3.7"
 
 branches:
     only:
@@ -44,12 +37,10 @@ install:
     # - pandas will bring dateutil numpy pytz setuptools six and scipy, we
     #   might want to avoid the later by installing all dependencies manually
     #   except scipy and install pandas with --no-deps
-    # - fixed versions of numpy and pytables because of incompatibilities
-    #   between pytables <= 3.4 and numpy >= 1.16
     # - pandas >= 0.20.0 is required since commit 01669f2024a7bffe47cceec0a0fd845f71b6f7cc
     #   (issue 702 : fixed bug when writing metadata using HDF format)
     - conda create -n travisci --yes python=${TRAVIS_PYTHON_VERSION:0:3}
-            "numpy==1.15.2" "pandas>=0.20" "pytables==3.4.4" matplotlib xlrd openpyxl
+            "numpy>=1.13" "pandas>=0.20" pytables matplotlib xlrd openpyxl
             xlsxwriter pytest pytest-pep8
     - source activate travisci
 
--- a/make_release.py
+++ b/make_release.py
@@ -39,14 +39,11 @@ def update_metapackage(context):
 
     # TODO: this should be echocall(redirect_stdout=False)
     print(fill('Updating larrayenv metapackage to version {version}'))
-    # - fixed versions of numpy and pytables because of incompatibilities
-    #   between pytables <= 3.4 and numpy >= 1.16
-    #   Didn't set pytables ==3.5 because it is not available for Python 3.5
     # - excluded versions 5.0 and 5.1 of ipykernel because these versions make the console useless after any exception
     #   https://github.com/larray-project/larray-editor/issues/166
     check_call(['conda', 'metapackage', 'larrayenv', version, '--dependencies', fill('larray =={version}'),
                 fill('larray-editor =={version}'), fill('larray_eurostat =={version}'),
-                "qtconsole", "matplotlib", "pyqt", "qtpy", "numpy ==1.15.2", "pytables ==3.4.4",
+                "qtconsole", "matplotlib", "pyqt", "qtpy", "pytables",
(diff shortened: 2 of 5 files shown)


.travis.yml
```diff
@@ -44,12 +36,10 @@ install:
     # - pandas will bring dateutil numpy pytz setuptools six and scipy, we
     #   might want to avoid the later by installing all dependencies manually
     #   except scipy and install pandas with --no-deps
-    # - fixed versions of numpy and pytables because of incompatibilities
-    #   between pytables <= 3.4 and numpy >= 1.16
     # - pandas >= 0.20.0 is required since commit 01669f2024a7bffe47cceec0a0fd845f71b6f7cc
     #   (issue 702 : fixed bug when writing metadata using HDF format)
     - conda create -n travisci --yes python=${TRAVIS_PYTHON_VERSION:0:3}
-            "numpy==1.15.2" "pandas>=0.20" "pytables==3.4.4" matplotlib xlrd openpyxl
+            "numpy=>1.13" "pandas>=0.20" "pytables=>3.5" matplotlib xlrd openpyxl
```
gdementen: you should **not** specify pytables >=3.5 as we do not really need it (besides isn't => a typo instead of >=?)

make_release.py
```diff
@@ -39,14 +39,14 @@ def fill(s):
 
     # TODO: this should be echocall(redirect_stdout=False)
     print(fill('Updating larrayenv metapackage to version {version}'))
-    # - fixed versions of numpy and pytables because of incompatibilities
-    #   between pytables <= 3.4 and numpy >= 1.16
-    #   Didn't set pytables ==3.5 because it is not available for Python 3.5
+    # - needs pytables =>3.5 to:
+    #      * get root._v_attrs available (used in core/metadata.py)
```
gdementen: hu? is it True? There ought to be an alternate syntax which worked with earlier versions of pytables, as this makes larray much harder to install. There is nothing in pytables 3.5 changelog related to that!?

alixdamman:
@gdementen I don't why I can simply add support for Python 3.7 without any hack in larray-editor project but not here ? Did I miss something ?

make_release.py
```diff
@@ -39,14 +39,14 @@ def fill(s):
...
+    #      * avoid incompatibility between numpy 1.16 and pytables 3.4
     # - excluded versions 5.0 and 5.1 of ipykernel because these versions make the console useless after any exception
     #   https://github.com/larray-project/larray-editor/issues/166
     check_call(['conda', 'metapackage', 'larrayenv', version, '--dependencies', fill('larray =={version}'),
                 fill('larray-editor =={version}'), fill('larray_eurostat =={version}'),
-                "qtconsole", "matplotlib", "pyqt", "qtpy", "numpy ==1.15.2", "pytables ==3.4.4",
+                "qtconsole", "matplotlib", "pyqt", "qtpy", "numpy =>1.13", "pytables =>3.5",
```
gdementen: same comments as above + typos in version dependences?

setup.cfg
```diff
@@ -3,12 +3,14 @@ test=pytest
 
 [tool:pytest]
 testpaths = larray
-# exclude (doc)tests from ufuncs (because docstrings are copied from numpy
-# and many of those doctests are failing
+# - exclude (doc)tests from ufuncs (because docstrings are copied from numpy
+#   and many of those doctests are failing
+# - deselect LArray.astype since doctests fails for Python 3.6 and numpy => 1.17
```
gdementen: => vs >=?

gdementen:
FWIW, "library=>number" is parsed as: "library[version='>number*']", which works in this case but is **not** what you meant. Please change those "=>" to ">=".

make_release.py
alixdamman: I remember that I had to update the version of pytables to 3.4.X of one of our users. I am think he had a 3.3.X version but I am not totally sure.

alixdamman:
Any clue explaining why tests are failing for Python 3.7 ? Did you made any change in the larray-editor project?

gdementen:
It seems like Travis-CI is still in the process of rolling out xenial. So it is mostly "random" whether we get trusty or xenial. So we should specify it after all (but globally instead of just for python 3.7). Sorry for the bad information...

https://changelog.travis-ci.com/xenial-as-the-default-build-environment-99476

alixdamman:
OK. This is what I thought.

.travis.yml
```diff
@@ -3,14 +3,16 @@
...
-python:
-    - "2.7"
-    - "3.6"
-
 # Work-around to enable python 3.7 without globally enabling sudo and dist: xenial for other build jobs
 # see https://github.com/travis-ci/travis-ci/issues/9815
 matrix:
   include:
+    - python: 2.7
+      dist: xenial
```
gdementen: use a global/top-level "dist: xenial" instead

.travis.yml
```diff
@@ -3,14 +3,16 @@
...
+      sudo: true
+    - python: 3.6
+      dist: xenial
+      sudo: true
```
gdementen: sudo: true is probably useless (as is sudo: false at the end of the script)

gdementen:
sigh... I meant top-level as in: https://docs.travis-ci.com/user/languages/python/#python-37-and-higher

alixdamman:
OK to merge?

gdementen:
Is "pytables>=3.5" really required on travis CI? I would hope it's not. Because if it is in travis, it should be also added to the dependencies in the "normal" conda recipe, but then larray might be difficult to install again. FWIW, I discovered larrayenv 0.30 is currently not installable at all on python 3.7 (even on a fresh miniconda install without many other packages making things harder). I am unsure why...

alixdamman:
It sounds like versions earlier to 3.4.2 of pytables are no longer available with conda. So, no, it is not really required

gdementen:
LGTM

PR #800, opened 2019-08-09
added stub function for run_editor_on_exception

gdementen:
this is a follow-up to larray-project/larray-editor#189


gdementen:
as far as I am concerned, once this is in, the release can go

PR #802, opened 2019-08-28
fix importing and exporting Session objects from/to CSV and Excel

alixdamman:


larray/core/array.py
```diff
@@ -1185,22 +1185,21 @@ def to_frame(self, fold_last_axis_name=False, dropna=None):
            b1    6   7
         """
         columns = pd.Index(self.axes[-1].labels)
+        axes_names = self.axes.display_names[:]
```
gdementen: Are you sure using display_names here is a good idea? You'll get `*` in axes names for wildcard axes and `{axis_pos}` for anonymous axes...

I am very nervous about this change. I fear it must break "something" somewhere given that to_frame is used in a lot of places (at least indirectly). Pandas dataframes with "{0}*" as explicit name would be ugly, right?
As will be plots with such labels.

doc/source/changes/version_0_32.rst.inc
```diff
@@ -55,4 +55,11 @@ Miscellaneous improvements
...
 ^^^^^
 
-* fixed something (closes :issue:`1`).
+* fixed reading/exporting sessions containing two or more axes/groups
+  with the same name (or anonymous) from/to CSV, Excel and HDF files (closes :issue:`803`).
+
+* fixed NaNs and None labels appearing in axes and groups when reading/exporting sessions
+  from/to CSV and Excel files (closes :issue:`804`).
+
+* fixed importing/exporting anonymous and/or wildcard axes to CSV and Excel files
```
gdementen: Was the problem present only when importing/exporting the axes objects themselves or arrays with such axes, or both?

doc/source/changes/version_0_32.rst.inc
```diff
@@ -55,4 +55,11 @@ Miscellaneous improvements
 Fixes
 ^^^^^
 
-* fixed something (closes :issue:`1`).
+* fixed reading/exporting sessions containing two or more axes/groups
```
gdementen: The work you have put into this PR makes me wonder even more whether *saving* sessions (and especially when using non LArray objects) to CSV and to Excel,  are worth it. I fear they will give us an endless stream of problems for little benefit.

larray/tests/test_array.py
```diff
@@ -3454,6 +3454,17 @@ def test_from_series():
     assert_array_equal(res, expected)
 
 
+def test_to_frame():
+    # array containing anonymous axes
+    arr = ndtest((Axis(2), Axis(2), Axis(2)))
+    df = arr.to_frame()
+    assert df.index.name is None
+    assert df.index.names == ['{0}*', '{1}*']
```
gdementen: is it really what we want???

doc/source/changes/version_0_32.rst.inc
alixdamman: No one noticed the bug meaning no one will notice if we drop the ability to save and load axes and groups with the CSV or Excel format. 
So, do I edit the title of the corresponding issue and drop Axis and Group objects when calling Session.save()/load() ?

gdementen: I am unsure where to draw the line/what's best:
1) support saving axes and groups (and possibly arrays with metadata) **only** to HDF (or other formats where we can store arbitrary things). In that case, we will need to output a warning when saving a session containing them to csv/excel. Or we could go the GIMP route and say there is a single "native" format (HDF or .la or whatever) and there are other formats we can export to, each with their own limitations.
2) only support saving sessions to HDF-like formats.
3) keep trying to support everything for all formats. Let's be clear: it would love to have that, but I fear the time spent on that could be better spent elsewhere.
4) possibly another intermediate option I am not seeing?
I think we need to discuss this face to face, as it will be easier.

alixdamman:
closes this PR because of issue #815 and comment https://github.com/larray-project/larray/pull/802#discussion_r326615032

PR #809, opened 2019-09-23
update the tutorial

alixdamman:

PR #811, opened 2019-10-02
issue 611: rename Larray class as Array

alixdamman:
Waiting for PR #809 to be merged before to finish this PR.

### DIFF
--- a/larray/core/array.py
+++ b/larray/core/array.py
@@ -2,7 +2,7 @@
 from __future__ import absolute_import, division, print_function
 
 """
-Matrix class
+Array class
 """
 
 # ? implement multi group in one axis getitem: lipro['P01,P02;P05'] <=> (lipro['P01,P02'], lipro['P05'])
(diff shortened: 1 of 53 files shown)


alixdamman:
The first and the three last commits are the most important.

larray/core/array.py
```diff
@@ -8027,63 +8027,69 @@ def transform(v):
         return self.apply(transform, dtype=dtype)
 
 
+class LArray(Array):
+    def __init__(self, *args, **kwargs):
+        warnings.warn("LArray has been renamed as Array.", FutureWarning, stacklevel=2)
+        Array.__init__(self, *args, **kwargs)
```
gdementen: can't we use `LArray = renamed_to(Array, 'LArray')` instead?

alixdamman: ``rename_to`` return a function and not a class. In addition, the printed warning message would have been `LArray() is deprecated. Use Array() instead`, i.e. with ( ) after class names.

gdementen: Calling a function and a class is essentially the same thing in Python (they are both callables) and I don't see the problem with that warning message.

PR #818, opened 2019-10-18
Dropped possibility to read and export Axis and Group objects from/to CSV and Excel files

alixdamman:


alixdamman:
There is still a problem with xlwings: exported axes names are given by array.axes.display_names.
In other words, anonymous axes are exported with names set to ``{x}*``.

larray/tests/test_session.py
```diff
@@ -103,173 +113,142 @@ def test_setitem(session):
...
+    s.load(fpath, engine=engine)
+    if engine == 'pandas_excel':
+        # Session.save() via engine='pandas_excel' always overwrite the output Excel files
+        assert s.names == ['e']
+    elif is_excel_or_csv:
+        assert s.names == ['e', 'f', 'g', 'h']
+    else:
+        assert s.names == session.names
+        assert s['a'].equals(a3)
+        if engine == 'pickle':
```
gdementen: Why does pickle needs a different code? I don't see any reason it **should**. Isn't this masking a bug? If this is not a bug, there should at least be a comment explaining the situation/why this difference in treatment is necessary.

doc/source/changes/version_0_32.rst.inc
```diff
@@ -58,3 +58,6 @@ Fixes
 
 * fixed the :py:obj:`Array.reindex()` printing a spurious warning message when the
   `axes_to_reindex` argument is the name of the axis to reindex (closes :issue:`812`)
+
+* fixed reading/exporting sessions containing two or more axes/groups
```
gdementen: from the code change below (which is surprisingly small), it seems you only changed the reading part, so I suppose the exporting part was fine, right?

larray/tests/test_session.py
alixdamman: The reason is that the key of a group is converted as a Numpy ndarray when using the HDF format while the pickle format keeps the original type of the key (in this case: a simple list). 
But the unittest function is fixed by by using the Group.equals() method.

doc/source/changes/version_0_32.rst.inc
alixdamman: Yes. For some reason I don't remember, the name of the loaded axis was set equal to the corresponding hdf key.

larray/core/array.py
```diff
@@ -2462,12 +2464,15 @@ def dump(self, header=True, wide=True, value_name='value', light=False, axes_nam
...
             all axes names will be included except the last. Defaults to True.
         na_repr : any scalar, optional
-            Replace missing values (NaN floats) by this value. Default to 'as_is' (do not do any replacement).
+            Replace missing values (NaN floats) by this value. Defaults to 'as_is' (do not do any replacement).
         maxlines : int, optional
             Maximum number of lines to show. Defaults to -1 (all lines are shown).
         edgeitems : int, optional
             If number of lines to display is greater than `maxlines`, only the first and last `edgeitems` lines are
             displayed. Only active if `maxlines` is not -1. Defaults to 5.
+        _axes_display_names : bool, optional
```
gdementen: are _parameters included in the api doc? If yes, I don't think we should document them the normal way, to avoid cluttering the api doc, but should rather add a comment with that info.

larray/tests/test_array.py
```diff
@@ -4134,6 +4134,16 @@ def test_dump():
                    ['a1', 1],
                    ['a2', 2]]
 
+    arr = ndtest((Axis('a0,a1'), Axis(2, 'b')))
```
gdementen: add a (short) comment to tell what you are testing. ie, something like: "with an anonymous axis"

larray/tests/test_array.py
```diff
@@ -4134,6 +4134,16 @@ def test_dump():
...
+    res = arr.dump()
+    assert res == [['b', 0, 1],
```
gdementen: This is a misleading representation of the data (one would think that 'b' is the name of the column for 'a0' and 'a1'), so I don't really like this, but I am unsure we can do anything about it, except change our format (use two header lines instead of one), which is obviously out of scope here. **Maybe** that should come out as `\b` instead??? That would be ugly but non-misleading. Even **if** we go that route, if both axes are anonymous/empty, the `\` should of course not be shown.

larray/inout/pandas.py
```diff
@@ -325,13 +325,17 @@ def df_asarray(df, sort_rows=False, sort_columns=False, raw=False, parse_header=
...
     else:
         axes_names = [decode(name, 'utf8') if isinstance(name, basestring) else name
                       for name in df.index.names]
         unfold_last_axis_name = isinstance(axes_names[-1], basestring) and '\\' in axes_names[-1]
-        return from_frame(df, sort_rows=sort_rows, sort_columns=sort_columns, parse_header=parse_header,
-                          unfold_last_axis_name=unfold_last_axis_name, cartesian_prod=cartesian_prod, **kwargs)
+        res = from_frame(df, sort_rows=sort_rows, sort_columns=sort_columns, parse_header=parse_header,
+                         unfold_last_axis_name=unfold_last_axis_name, cartesian_prod=cartesian_prod, **kwargs)
+
+    # ugly hack to avoid anonymous axes converted as axes with name 'Unnamed: x' by pandas
```
gdementen: this is indeed ugly, but I don't think we can do better, and since it is in df_asarray an not in from_frame, this is fine with me.

doc/source/changes/version_0_32.rst.inc
```diff
@@ -10,7 +10,9 @@ Syntax changes
 Backward incompatible changes
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-* other backward incompatible changes
+* Dumping and loading Axis and Group objects contained in a session is no longer possible
+  with the CSV and Excel formats. Please use the HDF format instead.
```
gdementen: please explain briefly why this was done. Something vaguely like "because that feature was broken and fixing it was not worth it because it was unused anyway (as far as we know)"

PR #819, opened 2019-10-21
add debug() in api and larray namespace

gdementen:


larray/inout/pickle.py
```diff
@@ -55,5 +55,6 @@ def _dump_metadata(self, metadata):
         self.data['__metadata__'] = metadata
 
     def close(self):
+        # FIXME: this is wrong in so many ways. If we only read the file, it will be re-dumped when closed?
```
alixdamman: why not override the save() method and move the code below to it? Instead of adding a FIXME.

alixdamman:
Except for the FIXME (open a new issue for it?), LGTM.

larray/inout/pickle.py
gdementen: because I didn't write that code and don't know the implications of modifying it. If it's that simple, I will try that.

gdementen:
It wasn't as simple as that to fix it properly, but not too hard either :). Merging...

PR #835, opened 2019-12-02
(issue 829): some changes in the tutorial

alixdamman:
On going PR


doc/source/tutorial/tutorial_miscellaneous.ipyml
```diff
@@ -58,18 +58,29 @@ cells:
...
-    pop.clip(10**7, 4*10**7)
+    population.clip(10**7, 4*10**7)
+
+
+- code: |
+    # Using vectors to define the lower and upper bounds
+    lower_bound = sequence(population.time, initial=5_500_000, inc=50_000)
+    upper_bound = sequence(population.time, 41_000_000, inc=100_000)
+    
+    print(lower_bound, '\n')
```
gdementen: does this render correctly (as nicely as cell outputs) in the documentation? I fear it might be different...

doc/source/tutorial/tutorial_transforming.ipyml
```diff
@@ -47,21 +47,15 @@ cells:
 
 
 - code: |
-    # replace all labels of the 'gender' axis by passing the list of all new labels
```
gdementen: I am unsure removing this entirely from the tutorial is a good thing because even if users should probably not do it, they will still want to do it and I fear (maybe this is unfounded) that if they don't find how to do it in the tutorial they will ask us. I guess we can try it this way and see what comes out of it.

doc/source/tutorial/tutorial_miscellaneous.ipyml
alixdamman: Yes, it does

gdementen:
I no longer have a clear overall picture of everything changed, but each change individually LGTM.

Issue #832, opened 2019-11-29
Make it easy to create a class that behaves like a Session


alixdamman:
The main idea of the present issue is to benefit from both auto-completion of PyCharm (by defining user classes) and all (nice) features offered by the Session class.

This implies to develop a mechanism to: 
- inherit from Session class,
- automatically push all attributes defined in the sub-class to the `self._objects` attribute of Session.
  
Another (big) advantage of using user-defined classes is to reduce significantly the number of global variables and function parameters in the models that are developed in house.

alixdamman:
@gdementen OK to "try" to implement this for the next release (0.33)?
I said "try".

gdementen:
I had in mind to do a quicker-than-usual release to fix the issues we found during the training (matplotlib import in the viewer, small tutorial/doc changes and the load data >2d without axes names problem), probably as 0.32.1 and then move on to liam2. So I am basically ok with the 0.33 milestone but that will depend when you want to release it.

alixdamman:
If we plan to do a 0.32.1 release, I think there is no reason to make 0.33 release soon.

gdementen:
Or... If you want to work on it before I am able to (probably somewhere in January), I give you the experimental code I have so far and you do it. So far, I have always linked this "static session" thing to the lazy session feature (see #727), but maybe we could dissociate the two. In that case it's pretty easy. The lazy part is a lot more tricky (when I say tricky, it's not necessarily hard to implement, but rather hard to figure out which behaviour is best for users).

alixdamman:
Related to issue #727

alixdamman:
See comment https://github.com/larray-project/larray/pull/840#issuecomment-567452278 for an example of how the ``StaticSession`` class should be used.

gdementen:
This is at least how **I** saw things. There are probably other options. I just don't think restricting values based on **values** defined at the class level is a good idea. It is just too weird in the Python world, while there are alternatives which are more usual (restrict on **types** defined at the class level) or restrict on values defined at the instance level (i.e. values defined in `__init__`)

alixdamman:
In the Demo model, they have a ``Parameters`` class in which they define all axes, groups and scalars of the model. The other classes contains only arrays but inherits from the ``Session`` class to mainly take benefit of its I/O methods. I tried to somewhat "reproduce" what is done in that model (consider all axes and groups defined in a class as frozen) plus to include the idea of ``ArrayDef`` but I see now that implementing a class just based on one model is wrong. 

[...]

PR #840, opened 2019-12-18
(issue 832): implement CheckedSession, CheckedParameters and CheckedArray 

alixdamman:
I know the name ``FrozenSession`` is misleading because values of array items can be modified but I couldn't find another better name to emphasis that axes of arrays and all non array items are 'frozen'.

I know that in your version of the ``StaticSession`` class you choose to let users to add or remove items and to print warnings message when there is a missing or extra item. I know you made that choice partly to not afraid present and potential users to use this new feature of LArray.
But personally, I can't convince myself that users take any care of warnings messages. I really can't. 
Letting them to add and remove items "on the fly" is way too dangerous and this partially breaks the main purpose of this new kind of Session, i.e. relying (safely) on 'autocomplete' to write models.


gdementen:
As I already tried to tell you earlier (sorry if this was not clear), this is **not** the way to go. You told me "it is just for testing", but you still do exactly the same here. You use values to define classes, instead of using types or "definitions". A FrozenSession (I also dislike that name) should be very similar to Python 3.7 dataclasses (maybe even use them but I am unsure that is possible) . In dataclasses you define the type of fields of your class, and optionally a default value. In what you did, you define a strict value which cannot change, and you cannot have several instances of that class at the same time. This is really a missed opportunity.

gdementen:
> But personally, I can't convince myself that users take any care of warnings messages. I really can't.

Hu? This is quite the opposite. Most of our users are panicked about warnings and consider them to be errors even when they are not.

> Letting them to add and remove items "on the fly" is way too dangerous and this partially breaks the main purpose of this new kind of Session, i.e. relying (safely) on 'autocomplete' to write models.

You sound like auto-completion is the end goal of everything, but it is just a nice plus. Speaking of something "way too dangerous" and autocompletion in the same sentence seems so odd to me...

gdementen:
About the name, here are a few suggestions: StaticSession (this would be telling for C/Java programmers, but I admit our target users are not), CheckedSession, DefinedSession, SessionDefinition

larray/core/session.py
```diff
@@ -1392,6 +1393,155 @@ def display(k, v, is_metadata=False):
...
+        is not permitted."""
+        raise ValueError(
+            "Adding or removing item and modifying axes of declared arrays is not permitted.".format(
+                cls=self.__class__.__name__
+            )
+        )
+
+    # XXX: not sure we should or not disable 'transpose()'?
+    __delitem__ = __delattr__ = _disabled
+    update = filter = transpose = compact = _disabled
```
gdementen: `update()` should not be disabled because it can be used to change the values of several *existing* arrays in one shot. 

Only adding new arrays should trigger a warning (or be disallowed, depending on the point of view) but this is already taken care of by `__setitem__()`, which `update()` uses, so it is not even necessary to override `update()`.

IMO, the other 3 methods should not be disabled either since they (should) just return a new (non frozen) Session.

larray/core/session.py
```diff
@@ -1392,6 +1393,155 @@ def display(k, v, is_metadata=False):
...
+
+    def copy(self):
+        instance = self.__class__()
+        for key, value in self.items():
+            instance[key] = copy(value)
+        return instance
+
+    def apply(self, func, *args, **kwargs):
+        kind = kwargs.pop('kind', Array)
+        instance = self.__class__()
```
gdementen: This should either be the code in the normal Session, so that `overridding` apply() here is not necessary, or not overriden at all (because I am unsure we actually want that behaviour for FrozenSession). I would rather have it return a normal Session to not be limited by the restrictions of the FrozenSession. If we do not do that, for example, my_frozen_session.apply(Array.sum, 'gender') would not work (because the returned axes are different).

larray/core/session.py
```diff
@@ -1392,6 +1393,155 @@ def display(k, v, is_metadata=False):
         return res
 
 
+# XXX: I wonder if we shouldn't create an AbstractSession instead of defining the _disabled()
+#      private method below.
+#      Auto-completion on any instance of a class that inherits from FrozenSession
+#      should not propose the add(), update(), filter(), transpose() and compact() methods
```
gdementen: **IF** we keep that "feature" of having those functions be non functional, then I agree with you, they should not be available at all **BUT** I am unconvinced about limiting so much what users can do (see below for more details)

gdementen:
To be clear, this is what we should strive for:

```python
class MyDemoSession(StaticSession):
    a = int                               # maybe using : instead of = is possible/desirable, I don't know
    b = ArrayDef("time=2010..2020")

# and then
d = MyDemoSession('path.h5')
# or
d = MyDemoSession(a=10, b=zeros("time=2010..2020"))
```
The advantage of this is that you can have several instances of the same session class without having to modify your code, making variants easier. We can even imagine dumping all parameters of each simulation run to a file, then being able to run several variants **just** by changing the path of the parameters file.

larray/core/session.py
```diff
@@ -1393,87 +1394,85 @@ def display(k, v, is_metadata=False):
...
-    The 'autocomplete' is a feature in which a software predicts the rest of a variable or function
-    name after a user typed the first letters. This feature allows users to use longer but meaningful
-    variable or function names (like 'population_be' instead of 'pbe') and to avoid creating an unwanted
-    new variable by misspelling the name of a given variable (e.g. typing 'poplation = something'
-    ('population' without u) will create a new variable instead of modifying the 'population' variable).
 
+# XXX: the name of the class below is not really important since it will serve as base
+#      for the LazySession and users are expected to inherit from the LazySession when
+#      they define their own classes
+class TypedSession(Session):
```
gdementen: I already like TypedSession much better than FrozenSession. But wouldn't naming it ConstrainedSession be better (we go beyond types when we mandate precise axes).

larray/core/session.py
```diff
@@ -1393,87 +1394,85 @@ def display(k, v, is_metadata=False):
...
+    ...     FIRST_PROJ_YEAR = int
+    ...     LAST_PROJ_YEAR = int
+    ...     AGE = Axis
+    ...     GENDER = Axis
+    ...     TIME = Axis
+    ...     G_CHILDREN = Group
+    ...     G_ADULTS = Group
+    ...     G_OBS_YEARS = Group
+    ...     G_PROJ_YEARS = Group
+    ...     population = ArrayDef(('AGE', 'GENDER', 'TIME'))
```
gdementen: I have some vague aversion about mixing arrays and parameters in the same Session. I cannot word it yet. Maybe it is just because that's not how you work in "demo", but I think there is something else behind my gut feeling. Technically I don't think we need anything different but I wonder what is better to show as an example: something simple but not optimal, or the "better" (but more complex) organisation.

larray/core/session.py
alixdamman: Good suggestion.

larray/core/session.py
```diff
@@ -1393,87 +1394,85 @@ def display(k, v, is_metadata=False):
...
-    ...     deaths = zeros((AGE, GENDER, TIME))
-    >>> m = ModelVariables()
-    >>> m.names    # doctest: +NORMALIZE_WHITESPACE
-    ['AGE', 'CHILDREN', 'ELDERS', 'FIRST_OBS_YEAR', 'GENDER', 'LAST_AGE', 'LAST_PROJ_YEAR', 'TIME', 'births',
-    'deaths', 'population']
+    Content of file 'model_variables.py'
+
+    >>> # ==== MODEL VARIABLES ====
+    >>> class ModelVariables(TypedSession):
+    ...     FIRST_OBS_YEAR = int
```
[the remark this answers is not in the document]
gdementen: Forget this. I have looked into it, but it seems like the typing module is not flexible enough to type multi-dimensional arrays yet (see https://github.com/python/typing/issues/516). We could invent our own syntax using annotations, and would cause no runtime problem but would be incompatible with any type checker out there, including mypy and PyCharm, so there is no point. Let's go for types and "array defs" as class variables like above.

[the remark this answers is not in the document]
alixdamman: Thanks for looked into it. I didn't try to inherit/use a dataclass because I thought that dataclass was limited to Python 3.7 or higher only.

larray/core/session.py
```diff
@@ -1393,87 +1394,85 @@ def display(k, v, is_metadata=False):
...
+    ...     deaths = ArrayDef(('AGE', 'GENDER', 'TIME'))
+
+    Content of file 'model.py'
+
+    >>> def run_model(variant_name, first_proj_year, last_proj_year):
+    ...     # create an instance of the ModelVariables class
+    ...     m = ModelVariables()
+    ...     # ==== setup variables ====
+    ...     # set scalars
+    ...     m.FIRST_OBS_YEAR = 1991
```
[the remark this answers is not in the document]
alixdamman: Right. 

But there is something else for what I would like some advices. The main drawbacks of allowing to **refer to axes defined at runtime** in ``var = ArrayDef((...))`` are:

1.   they must be referred using strings -> we loose the benefit auto-completion + make it dangerous to rename these axes (via refactoring (Shift + F6) in PyCharm)
2.   the following code:
```python
m = ModelVariables()
....
# save to Excel file -> save Array objects only (not Axis objects)
m.save('variant_0.xlsx')
...
# load variables for variant 0 from Excel file -> FAIL
m = ModelVariables('variant_0.xlsx')
```
will fail because Axis objects are not saved in Excel or CSV files.

what's your opinion on this?

larray/core/session.py
```diff
@@ -1393,87 +1394,85 @@ def display(k, v, is_metadata=False):
...
+    ...     # some code here
+    ...     # ...
+    ...     # ==== output ====
+    ...     # save all variables in an HDF5 file
+    ...     m.save('{variant_name}.h5', display=True)
+
+    Content of file 'main.py'
+
+    >>> run_model('proj_2020_2070', first_proj_year=2020, last_proj_year=2070)
+    dumping FIRST_OBS_YEAR ... Cannot dump FIRST_OBS_YEAR. int is not a supported type
```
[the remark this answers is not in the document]
alixdamman: In progress...

larray/core/session.py
alixdamman: What do you consider as parameters? Everything that doesn't change at runtime or read-only scalars?

larray/core/session.py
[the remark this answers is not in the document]
gdementen: IMO, this is not a problem. If they want to do this, they should not use an Excel file! Now, it's up to us to make this clear in the documentation. In the code, nothing needs to change I think as there is already a warning that the axis object is not saved.

larray/core/session.py
gdementen: Hehe, this innocent-looking question is actually a very good question, because it would help to have a clear idea and precise definitions of this both for our examples/documentation and our users models. I think I would define a constant as something which never ever changes (across all variants of a model) and a parameter as anything which can change from one variant or simulation to the next but is constant over the simulation. This might be a very personal take on the issue, I don't know. Unsure where to draw the line between input data and parameter though.

larray/core/session.py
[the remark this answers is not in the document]
gdementen: dataclasses are indeed python 3.7+, but (class) variables annotations (PEP 526) are python3.6+.
dataclasses are based on those (and this is what I actually wanted to try using -- sorry for using the wrong word -- those were synonyms in my head when they should not). Either way, this is not ready yet and I am unsure it will ever be flexible enough to be able to provide custom validation functions in Python (to validate that an array actually has the correct axes).

larray/core/session.py
```diff
@@ -1392,6 +1393,155 @@ def display(k, v, is_metadata=False):
...
+    """
+    The purpose of the present class is to be inherited by user defined classes where parameters
+    and variables of a model are defined (see examples below). These classes will allow users to
+    benefit from the so-called 'autocomplete' feature from development software such as PyCharm
+    (see the Notes section below) plus the main features of the :py:obj:`Session()` objects.
+
+    After creating an instance of a user defined 'session', some restrictions will be applied on it:
+
+    - **it is not possible to add or remove any parameter or variable,**
+    - **all non array variables (axes, groups, ...) cannot be modified,**
```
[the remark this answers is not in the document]
gdementen: In other words, if we want to enforce the readonly nature of some members, this should be explicitly declared by the user for each variable, and not be automatically True for scalars and False for arrays. This readonly enforcement can come at a later point though.

larray/core/session.py
```diff
@@ -1392,6 +1393,155 @@ def display(k, v, is_metadata=False):
...
+    - **only values of array variables can be modified, not their axes.**
```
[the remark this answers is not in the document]
gdementen: Honestly, I am unsure about the "more axes" part. Personally, I would certainly like to work like that (declare all required axes by my model and let it work with any dataset which has at least those axes), but then you lose the quick reference we have now were we can quickly lookup the array name and know its exact axes, instead of knowing minimum required axes you can rely on. Knowing exact axes without running the model has some value too. If I was a modeller, generic code would be more important than "precise" axes, but I wonder if our users wouldn't favour precise axes over generic/powerful code.

alixdamman:
@gdementen I have a problem with Python 2.7 and the HDF5 format (again):
```
ValueError: unsupported pickle protocol: 4

../../../miniconda/envs/travisci/lib/python2.7/site-packages/tables/atom.py:1227: ValueError
```
I think you solved the same kind of problem recently. Do you remember how to do that?

gdementen:
@alixdamman I don't remember having this issue (but my memory is so bad I am unsure), I guess had a different issue. Here the issue is because you are trying to read on Python2 an hdf file created using python 3 (and thus pickle protocol 4). There is no clean fix for this because it is a limitation (AFAICT) of PyTables which always use pickle latest protocol (see https://github.com/PyTables/PyTables/issues/472 and https://stackoverflow.com/questions/60067953/is-it-possible-to-specify-the-pickle-protocol-when-writing-pandas-to-hdf5). I don't know for sure why it worked before. I **suppose** it is something like you only used basic data types before (which have native hdf support and are thus not pickled) but now you are trying to dump more "complex" data types like (unicode?) strings. 

I see several workarounds: only dump numeric stuff in the examples (maybe explicit bytes string would work too?), or use different file names for Python2 and Python3 (so that python2 reads the python2 file and Python3 the python3 file), or use pandas/pytables "table" format (which only supports non-pickle types -- see https://github.com/pandas-dev/pandas/issues/4260). 

But maybe it is just time to drop Python2. FWIW, I have already started not expecting any more change in larray for LIAM2 (I am monkey-patching larray as I need), so I guess it would not get any more painful than it already is... I still need to do 0.32.2 though (because no support for Pandas 1.0 **is** bad) but that's not as easy as I would have liked because the release scripts do not work on a branch anymore.

alixdamman:
Looking at the examples I have written for the ConstrainedSession class, I fear that users will find cumbersome to first have to define the type and as a second step to give an initial value. 

Couldn't we allow users to declare either the type or a "default initializing" value ?  

Something like:
```python
class MyDemoSession(ConstrainedSession):
    a = 10
    b = ArrayDef("time=2010..2020")

# and then
d = MyDemoSession('path.h5')
# or
d = MyDemoSession(b=zeros("time=2010..2020"))  # where an instance attribute 'a' will be added with value 10
```

gdementen:
Hmmm. Having constants values directly in the class definition is completely fine to me but I had the idea that those would be constant values (over all possible variants) instead of being default values. Otherwise, it seems to **me** the boundary of what you can modify or not is fuzzy: you can modify something defined in the class if it is a scalar, but not if it is an array? And the name ConstrainedSession suddenly seems a bit strange.

larray/core/session.py
```diff
@@ -1475,6 +1477,139 @@ def display(k, v, is_metadata=False):
...
+        self.axes = axes
+
+
+class ConstrainedSession(Session):
+    """
+    Examples
+    --------
+    Content of file 'constants.py'
+
+    >>> # ==== CONSTANTS ====
```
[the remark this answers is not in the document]
alixdamman: OK

alixdamman:
> Hmmm. Having values directly in the class definition is completely fine to me but I had the idea that those would be constant values (over all possible variants) instead of being default values.

OK, so for each non-array declared variable, users have the choice between specifying a constant value (over all variant) or its type. Right?

alixdamman:
> This makes me think that we could also implement something like:

```python
d = MyDemoSession.zeros()
# or maybe even (though that might be not explicit enough)
d = MyDemoSession()
```
> which would initialize all arrays to zeros().

Or we could add an optional argument to ArrayDef:
```python
class MyDemoSession(ConstrainedSession):
    a = 10
    b = ArrayDef("time=2010..2020", fill_value=0)

# and then
d = MyDemoSession('path.h5')
# or
d = MyDemoSession()
```
?

gdementen:
> Or we could add an optional argument to ArrayDef:
> 
> ```python
> class MyDemoSession(ConstrainedSession):
>     a = 10
>     b = ArrayDef("time=2010..2020", fill_value=0)
 
This is probably better than my "d = MyDemoSession.zeros()" suggestion because it is more granular, but I am still unsure about this because then distinction between a constraint and a normal initialized array becomes very thin.

In any case, I would only implement this in a second stage, if it actually helps a practical use case.

alixdamman:
I need some help for the implementation of the ``ConstrainedSession.__init__`` method (see last commit). 
I'm not sure what to do.

larray/core/session.py
```diff
@@ -1561,14 +1533,55 @@ class ConstrainedSession(Session):
...
+        super().load(fname, names, engine, display, **kwargs)
+        for key in set(self._cls_attrs.keys()) - set(self.keys()):
+            warnings.warn(f"The variable {key} is declared in the {self.__class__.__name__} "
+                          f"class definition but has not found in file {fname}.")
+        for key in set(self.keys()) - set(self._cls_attrs.keys()):
+            warnings.warn(f"The variable {key} has been found in file {fname} but is not declared "
+                          f"in the  {self.__class__.__name__} class definition.")
+
+    def save(self, fname, names=None, engine='auto', overwrite=True, display=False, **kwargs):
+        for key in set(self._cls_attrs.keys()) - set(self.keys()):
```
gdementen: There should **not** be such checks and warnings during save! Those should have happened earlier (in `__setitem__` and `__setattr__` -- which is already the case AFAICT)!

larray/core/session.py
```diff
@@ -1590,7 +1603,9 @@ def _check_key_value(self, key, value):
         attr_def = getattr(cls, key, None)
         if attr_def is None:
             warnings.warn("'{}' is not declared in '{}'".format(key, self.__class__.__name__), stacklevel=2)
-        else:
+        elif isinstance(attr_def, (type, ArrayDef)):
+            if value is None:
```
gdementen: why do you need this special case for None?

larray/core/session.py
```diff
@@ -1602,6 +1617,17 @@ def _check_key_value(self, key, value):
...
+        else:
+            def equal(v1, v2):
+                if isinstance(v1, (Group, Axis)):
+                    return v1.equals(v2)
+                elif isinstance(v1, Array):
+                    return v1.equals(v2, nans_equal=True)
+                else:
+                    return v1 == v2
+            if not equal(value, attr_def):
+                warnings.warn(f"Cannot modify the value of the variable '{key}' declared as a constant in the "
```
gdementen: You need to **either** change the message (which sounds like the action was blocked) or raise an exception instead of using a warning.

larray/tests/test_session.py
```diff
@@ -672,15 +665,15 @@ def test_create_typedsession_instance(meta):
 
     # load from file
     ts = TestConstrainedSession(inputpath('test_session.h5'))
-    assert ts.names == ['a', 'a01', 'a2', 'ano01', 'anonymous', 'b', 'b024', 'e', 'f', 'g', 'h']
+    assert ts.names == ['a', 'a01', 'a2', 'ano01', 'anonymous', 'b', 'b024', 'c', 'e', 'f', 'g', 'h']
```
gdementen: all those ts variables look odd now that the class is no longer named typesession

gdementen:
> I need some help for the implementation of the `ConstrainedSession.__init__` method (see last commit).
> I'm not sure what to do.

The question is about the ordering of keys, right? Which ordering is lost? The one from the class or the one from the instance (when initializing from kwargs, a dict or another session)? You are right that we cannot keep both. I think we should try to maintain the **class** ordering and I think it should be possible to do that (but I haven't thought that much about this).

larray/core/session.py
```diff
@@ -1475,6 +1478,164 @@ def display(k, v, is_metadata=False):
...
+            self.add(**constant_cls_attrs)
+        else:
+            # Note: same as above
+            kwargs.update(constant_cls_attrs)
+            Session.__init__(self, *args, **kwargs)
+
+    def load(self, fname, names=None, engine='auto', display=False, **kwargs):
+        super().load(fname, names, engine, display, **kwargs)
+        for key in set(self._cls_attrs.keys()) - set(self.keys()):
+            warnings.warn(f"The variable {key} is declared in the {self.__class__.__name__} "
```
gdementen: I am **unsure** this warning should exist. Creating a session out of several files seems like something we should allow without a warning. Maybe the warning should be there only if names is None???

If key was in names and it's not in the file, this should be an error (not a warning) but I suppose super().load would already catch that.

larray/core/session.py
```diff
@@ -1475,6 +1478,164 @@ def display(k, v, is_metadata=False):
...
+                          f"class definition but has not found in file {fname}.")
+        for key in set(self.keys()) - set(self._cls_attrs.keys()):
+            warnings.warn(f"The variable {key} has been found in file {fname} but is not declared "
+                          f"in the  {self.__class__.__name__} class definition.")
```
gdementen: I am **unsure** this warning should exist either. It seems reasonable to me to instantiate a session with a file containing more variables than necessary. LIAM2 accepts this kind of things (if I transpose the concepts) and users have found this feature quite useful. The trick is that, in that case LIAM2 only loads the declared stuff. **Maybe** we should do this here too?

larray/core/session.py
gdementen: This is not innocent though. If we allow this, we must handle the case where we request an attribute which has been declared but not loaded (yet). I am leaning towards that being a good idea (because we will need something like that for LazySession anyway).

gdementen: But, to keep things simple, we might want to start without that feature and implement it later.

larray/core/session.py
alixdamman: I first tried to raise an exception but it breaks the ``__init__`` and the ``load`` methods. 
I think raising an exception is the proper way to do it but I am not sure how to bypass this check when calling the ``__init__`` or ``load`` method?

larray/core/session.py
alixdamman: To not loose the order of the variables declared in the user defined class, I need to find a way to add the non-constant variables with a "temporary" value when calling the ``__init__`` method. For the moment, I add the non-constant variables with the value set to None in the ``__init__`` method until the user set a proper value later.

larray/core/session.py
alixdamman: > The trick is that, in that case LIAM2 only loads the declared stuff.
So LIAM2 only loads the declared stuff but does not print any warnings if there are undeclared stuff in the file?

gdementen: Yes. The spirit is that you declare the fields (LIAM2 vocabulary) you need in your model and can used whatever file which provides those fields. This makes it possible to use the same input data files for several variants, some of which need extra fields.

larray/core/session.py
```diff
@@ -1475,6 +1478,173 @@ def display(k, v, is_metadata=False):
...
+    ...     # declare variables with defined types.
+    ...     # Their values will be defined at runtime but must match the specified type.
+    ...     variant_name = str
+    ...     first_proj_year = int
+    ...     obs_years = Axis
+    ...     proj_years = Axis
+    ...     population_obs = Array
+    ...     population_proj = Array
+    ...     # declare arrays with fixed axes
+    ...     mortality_rate = ArrayDef((AGE, GENDER))
```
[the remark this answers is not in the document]
gdementen: Also, I think it that it would be useful to provide a way for users to declare that an array must have some specific axes even if the labels of those axes are not specified yet. In the example above, this would mean the ability to write something like either:
```python
population_obs = ArrayDef((AGE, GENDER, "year"))
```
or better yet (if possible -- I think it would be tricky but possible):
```python
population_obs = ArrayDef((AGE, GENDER, obs_years))
```
the later might require using AxisDef **instances** instead of just "Axis" to be able to differentiate instances: eg:
```python
obs_years = AxisDef("time")
population_obs = ArrayDef((AGE, GENDER, obs_years))
```

The goal is keep the class definition as close as possible to a complete reference of what each variable is (and also be more precise in the safety checks)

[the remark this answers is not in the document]
gdementen: I also wonder how PyCharm behaves when using "show definition" on such an attribute. Does it show the class definition or the point where it was given a value "m.obs_years = Axis(...)"? I am guessing the later, when it is defined but I don't really know.

gdementen:
One more comment. We must make it easy for the user to **somehow** create a ConstrainedSession from **arbitrary** files (in any strange format they like). Obviously in that case they need to provide a function to load those files. There are two ways to approach this: 

* either ask the user to create a function which creates/loads arrays and axes and returns an instance of their specific session class. I don't imagine this would cause any problem.
```python
class MyData(ConstrainedSession):
    a = AxisDef()
    arr1 = ArrayDef(a)
def load_stuff(filepath):
    with open_excel(filepath) as wb:
        arr1 = wb[...][...].load()
    return MyData(a=..., arr1=arr1)
```
* allow users to override their session class's __init__ method. I think it would be cleaner, but depending on how ConstrainedSession is implemented, this **could** be harder to get working.

```python
class MyData(ConstrainedSession):
    a = AxisDef()
    arr1 = ArrayDef(a)
    def __init__(self, filepath):
        with open_excel(filepath) as wb:
            arr1 = wb[...][...].load()
        # or is there a better way?
        super().__init__(a=..., arr1=arr1)
```

I am unsure anything needs to change in the code, but I think testing this is important (to at least know where we stand) and providing an example of whatever is the "best" way in the tutorial would be a good idea I think.

larray/core/session.py
alixdamman: > The usual trick is to create a global variable with an empty object as a unique flag:
```python
NOT_LOADED = object()
```
Thanks for the trick.

alixdamman: > Also, it might (I have not thought this through) be useful to pass extra/optional arguments to a few functions including _check_key_value so that you know whether or not you are currently during __init__.

Since ``_check_key_value()`` is used in ``__setitem__`` and ``__setattr__`` methods and those methods do not accept extra arguments, I wonder if it is not better to add a flag as instance attribute to activate or deactivate the raising of an exception when a user try to set a value to a variable declared as constant?

gdementen: I don't know whether that's a good idea or not in this case, but you **can** define those methods with extra arguments (and use those extra arguments if you call those methods explicitly). It's just that the normal syntax/implicit way to call them will not **use** those extra arguments. See for example `LArray.__getitem__`

larray/core/session.py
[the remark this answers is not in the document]
alixdamman: > Also, I think it that it would be useful to provide a way for users to declare that an array must have some specific axes even if the labels of those axes are not specified yet.

You're right but can I move this suggestion to a separate issue and implement it later? This PR is complicated enough.

alixdamman:
> One more comment. We must make it easy for the user to somehow create a ConstrainedSession from arbitrary files (in any strange format they like).

OK to create a separate issue for that too?

gdementen:
> You're right but can I move this suggestion to a separate issue and implement it later? This PR is complicated enough.

> OK to create a separate issue for that too?

You split your work in PR in whatever way is more convenient for you. What matters for me is how the work is split between releases and I'd prefer if users had the whole thing in one release.

alixdamman:
That could be done for the second proposed issue but the first one will be more complicated to implement I think (but I can try to do it for the next release).

gdementen:
yes, more complicated, but more needed too. I am unsure the whole thing will really be usable for a model like "demo" without that. Would need to try though.

alixdamman:
> 
> 
> yes, more complicated, but more needed too. I am unsure the whole thing will really be usable for a model like "demo" without that. Would need to try though.

I'll do my best.

larray/core/session.py
```diff
@@ -1475,6 +1492,199 @@ def display(k, v, is_metadata=False):
...
+        return value
+
+
+# the implementation of the function below is inspired by the 'conbytes' function
+# from the types.py module of the 'pydantic' library
+def ConstantAxesArray(axes) -> Type[ConstrainedArrayImpl]:
+    if axes is not None and not isinstance(axes, AxisCollection):
+        axes = AxisCollection(axes)
+    namespace = {'expected_axes': axes}
+    return type('ConstrainedArrayImpl', (ConstrainedArrayImpl,), namespace)
```
gdementen, line 1545: Can we avoid giving the subclass the same name than its parent?

And isn't this another way to spell:
```python
if axes is not None and not isinstance(axes, AxisCollection):
    axes = AxisCollection(axes)
class ConstrainedArrayImpl(ConstrainedArrayImpl):
    expected_axes = axes
return ConstrainedArrayImpl
```
**If** so, I find it more readable using the class keyword.

larray/core/session.py
```diff
@@ -1475,6 +1492,199 @@ def display(k, v, is_metadata=False):
...
+
+    As for normal Session objects, it is still possible to add undeclared variables to instances of
+    classes inheriting from `ConstrainedSession` but this must be done with caution.
+
+    Warnings
+    --------
+    The :py:method:`ConstrainedSession.filter`, :py:method:`ConstrainedSession.compact`
+    and :py:method:`ConstrainedSession.apply` methods return a simple Session object.
+    The type of the declared variables (and the value for the declared constants) will
+    no longer by checked.
```
gdementen, line 1571: s/by/be

larray/core/session.py
```diff
@@ -1475,6 +1492,199 @@ def display(k, v, is_metadata=False):
...
+def ConstantAxesArray(axes) -> Type[ConstrainedArrayImpl]:
+    if axes is not None and not isinstance(axes, AxisCollection):
+        axes = AxisCollection(axes)
+    namespace = {'expected_axes': axes}
+    return type('ConstrainedArrayImpl', (ConstrainedArrayImpl,), namespace)
+
+
+class ConstrainedSession(BaseModel, Session):
+    """
+    This class is intended to be inherit by user defined classes in which the variables of a model are declared.
```
gdementen, line 1550: s/inherit/inherited

larray/core/session.py
```diff
@@ -1475,6 +1492,199 @@ def display(k, v, is_metadata=False):
...
+
+    def __init__(self, *args, **kwargs):
+        meta = kwargs.pop('meta', Metadata())
+
+        # create an intermediate Session object to not call the __setattr__
+        # and __setitem__ overridden in the present class
+        s = Session(*args, **kwargs)
+
+        # Warning: order of fields is not preserved (seems like fields with default values comes after)
+        items = dict(s.items())
```
gdementen, line 1655: FWIW, as far as the **items below are concerned, you do not need to convert s to a dict, you can use **s directly.
But maybe that does not work for the `items.update(self.__field_defaults__)` line?

gdementen:
I wonder if instead of ConstantAxesArray, we shouldn't go for something more generic like ConstrainedArray, so that we can also constrain the dtype:
```python
ConstrainedArray(atleastaxes=(AGE, GENDER), dtype=int)
```
or
```python
ConstrainedArray(exactaxes=(AGE, GENDER), dtype=int)
```

FWIW, I also thought about more constraints such as min_value or max_value, but that would probably be too much of a performance penalty (because then we have to check the array data itself in addition to its metadata) and we probably cannot guarantee it anyway because of inplace modifications via `Array.__setitem__`.

alixdamman:
> I wonder if instead of ConstantAxesArray, we shouldn't go for something more generic like ConstrainedArray, so that we can also constrain the dtype:
```python
ConstrainedArray(atleastaxes=(AGE, GENDER), dtype=int)
```
> or
```python
ConstrainedArray(exactaxes=(AGE, GENDER), dtype=int)
```
I hesitated to suggest to constrain the dtype and to rename ``ConstantAxesArray`` as ``ConstrainedArray``. But I didn't thought of `atleastaxes`. 

What about:
```python
ConstrainedArray(axes=(AGE, GENDER), exact_axes=True, dtype=int)
```
so that users could choose between "exact axes" and "at least axes"? 

> FWIW, I also thought about more constraints such as min_value or max_value, but that would probably be too much of a performance penalty (because then we have to check the array data itself in addition to its metadata) and we probably cannot guarantee it anyway because of inplace modifications via `Array.__setitem__`.

Let's start with specifying the axes and the dtype and **see if our users adhere to "constrained" sessions/arrays** (I really like the concept of "constrained" sessions/arrays but I don't know if our users will share the same enthusiasm).

Last question: how do you add emoticons in github comments?

larray/core/session.py
alixdamman: I guess what you would like to test is:
```python
if axes is not None and not isinstance(axes, AxisCollection):
    axes = AxisCollection(axes)
class ConstrainedArray(ConstrainedArrayImpl):
    expected_axes = axes
return ConstrainedArray
```
?
Good question but since the class ``ConstrainedArray`` is defined inside the function in this case, I wonder how PyCharm will react with auto-completion? 
I'll try.

gdementen:
> What about:
> 
> ```python
> ConstrainedArray(axes=(AGE, GENDER), exact_axes=True, dtype=int)
> ```
> 
> so that users could choose between "exact axes" and "at least axes"?

Looks fine to me too. There is then the question about the default value for exact_axes. Unsure anybody except me would use "at least axes", even though I think that is more powerful & creates more maintainable code but I admit I am not a normal user of the lib and I am obviously biased...

> Last question: how do you add emoticons in github comments?

When you type a colon `:`, it starts giving you emoticon options...  e.g. 😄 is `:smile:`

larray/core/session.py
gdementen: There are two questions in there: 1) the name of the generated class (using the same name might indeed help PyCharm) and 2) can we use the class keyword which I find more readable than the code using type() (AFAICT, it is equivalent but I could be wrong).

alixdamman:
> Looks fine to me too. There is then the question about the default value for exact_axes. Unsure anybody except me would use "at least axes", even though I think that is more powerful & creates more maintainable code but I admit I am not a normal user of the lib and I am obviously biased...

Honestly, I was unsure of the usefulness of implementing the "at least axes" feature. So what I suggest is to create a specific undocumented class (and function) just for you and not tell our users. 
If there is an actual demand for it, it will be already implemented and just adding documentation would be enough.  
If you agree, could you please create a separate issue for this?

> When you type a colon ``:``, it starts giving you emoticon options... e.g. 😄 is ``:smile:``

Thanks

gdementen:
> Honestly, I was unsure of the usefulness of implementing the "at least axes" feature. So what I suggest is to create a specific undocumented class (and function) just for you and not tell our users.
> If there is an actual demand for it, it will be already implemented and just adding documentation would be enough.
> If you agree, could you please create a separate issue for this?

No need for a separate class. Just don't implement the feature **yet**. I know the problem (a variant having more axes than the baseline) **will** arise **some**day but even then I guess our users would probably prefer adding the extra axes to the baseline than use "at least axes", so the demand could never come (unless I am some day responsible for a model).

alixdamman:
I pushed the last commit by mistake. Do not review it (yet).

alixdamman:
I was able to reactivate the `NOT_LOADED` mechanism by implementing my own version of the `ModelMetaclass` metaclass + not making `ConstrainedSession` to inherit from `pydantic.BaseModel`.

All tests are green on my computer.

However, `pydantic` is still an hard dependency (at least for the moment). Implementing our simplified version of `ModelField` will take a lot of time (but not impossible) but to tell the truth, I'm not it worth it in a sense that I'm not sure we will be able to really simplify it. I don't really want to copy/past an entire quiet big class and being able to only remove 30% or less of its code. 
The "30%" actually strongly depends on what types we want to be able to manage: only scalars + LArray types or tuple/list/dict/... also. IMO, we should include tuple/list/dict/...

What I can do for the moment is to do something like for the `Workbook` class: testing if `pydantic` is installed otherwise throw an ImplementedError in the `ConstrainedSession.__init__` if not.

alixdamman:
About the last pushed commit, could you please checkout the current branch on your own computer and make some autocomplete tests on `ConstrainedArray` objects with and without this last commit?

gdementen:
> I was able to reactivate the `NOT_LOADED` mechanism by implementing my own version of the `ModelMetaclass` metaclass + not making `ConstrainedSession` to inherit from `pydantic.BaseModel`.
> 
> All tests are green on my computer.

🥳 🎉
 
> However, `pydantic` is still an hard dependency (at least for the moment). Implementing our simplified version of `ModelField` will take a lot of time (but not impossible) but to tell the truth, I'm not it worth it in a sense that I'm not sure we will be able to really simplify it. I don't really want to copy/past an entire quiet big class and being able to only remove 30% or less of its code.
> The "30%" actually strongly depends on what types we want to be able to manage: only scalars + LArray types or tuple/list/dict/... also. IMO, we should include tuple/list/dict/...

This decision is yours as I have not studied the code so I can't really judge. But simplification of the parts which already work for us is clearly not a goal in itself. It can be interesting to do it IF we want to really diverge from pydantic. The question is: what is the option taking the least amount of time for us on the long run. 

Is it?

- semi-fork (what you have now). In this case, I would probably fix a semi-precise version of pydantic (in the run_constraints) to ensure your custom code is compatible with the library code.
- complete fork (copy all the code you did not modify as is).
- complete fork but simplified.
 
> What I can do for the moment is to do something like for the `Workbook` class: testing if `pydantic` is installed otherwise throw an ImplementedError in the `ConstrainedSession.__init__` if not.

Yes, I would prefer if it was like this at this point, because I don't know if many people will actually use ConstrainedSession.

alixdamman:
> This decision is yours as I have not studied the code so I can't really judge. But simplification of the parts which already work for us is clearly not a goal in itself. It can be interesting to do it IF we want to really diverge from pydantic. The question is: what is the option taking the least amount of time for us on the long run.

> Is it?

>   - semi-fork (what you have now). In this case, I would probably fix a semi-precise version of pydantic (in the run_constraints) to ensure your custom code is compatible with the library code.
>    - complete fork (copy all the code you did not modify as is).
>    - complete fork but simplified.

I honestly don't really know what is the best option. I think we can start with the "semi-fork" option (what I have implemented until know) **with `pydantic` as optional dependency** and see how many people will use `ConstrainedSession`.

> Yes, I would prefer if it was like this at this point, because I don't know if many people will actually use ConstrainedSession.

To be clear, my idea here is to let people use `ConstrainedSession` only if `pydantic` is installed as it is currently the case for the `open_excel()` function.  

dependencies of `pydantic` are given here: https://github.com/samuelcolvin/pydantic/blob/master/requirements.txt

alixdamman:
FWIW, I am convinced that `pydantic` have a real added value for our SDMX project even if I don't use `pandasSDMX` internally (I do not plan to use it actually). But that concerns only our users who need to deal with SDMX data.

gdementen:
> To be clear, my idea here is to let people use `ConstrainedSession` only if `pydantic` is installed as it is currently the case for the `open_excel()` function.

Yes, this is what I had in mind too.

> dependencies of `pydantic` are given here: https://github.com/samuelcolvin/pydantic/blob/master/requirements.txt

If I understand correctly what -r does in a requirements.txt files, those dependencies are awfully broad! 😱 They even depend on sqlalchemy (via their docs requirements), which is a great project but completely unrelated to what we are doing. All the more reason to make pydantic an optional dependency, I certainly don't want to bring such a huge pile of stuff in our dependencies for everybody.

alixdamman:
>> dependencies of pydantic are given here: https://github.com/samuelcolvin/pydantic/blob/master/requirements.txt

> If I understand correctly what -r does in a requirements.txt files, those dependencies are awfully broad! 😱 They even depend on sqlalchemy (via their docs requirements), which is a great project but completely unrelated to what we are doing. All the more reason to make pydantic an optional dependency, I certainly don't want to bring such a huge pile of stuff in our dependencies for everybody.

I didn't remember that I had to install the `sqlalchemy` lib with `pydantic`. I thought that the `requirements.txt` describes the **required** dependencies but it seems I was wrong. 
Indeed, I checked the `Install` page of the `pydantic` documentation: https://pydantic-docs.helpmanual.io/install/. This page starts by saying:
> Installation is as simple as:
```bash
    pip install pydantic
```
> **pydantic has no required dependencies except python 3.6, 3.7, or 3.8 (and the dataclasses package in python 3.6). If you've got python 3.6+ and pip installed, you're good to go.**

So, to be sure, I created a fresh new `TEST` conda environment running:
```bash
conda create --name TEST python=3.6
conda activate TEST
```
and then I installed `pydantic` from this new conda env:
```bash
(...)
The following packages will be downloaded:

    package                    |            build
    ---------------------------|-----------------
    dataclasses-0.7            |           py36_0          34 KB
    pydantic-1.5.1             |   py36he774522_0         156 KB
    ------------------------------------------------------------
                                           Total:         190 KB

The following NEW packages will be INSTALLED:

  dataclasses        pkgs/main/win-64::dataclasses-0.7-py36_0
  pydantic           pkgs/main/win-64::pydantic-1.5.1-py36he774522_0

Proceed ([y]/n)? y


Downloading and Extracting Packages
dataclasses-0.7      | 34 KB     | ############################################################################ | 100%
pydantic-1.5.1       | 156 KB    | ############################################################################ | 100%
(...)
```
It sounds like that what we see in the `requirements.txt` file are only **optional** dependencies that I don't need to install for this PR.  Also, what we can see from the installation of `pydantic` using conda is that `dataclasses` (required for Python 3.6 only) and `pydantic` represents respectively 34KB and 156KB. 

But I still will make `pydantic` as an optional dependency for `larray`.

alixdamman:
Needs another round of review ? I would like to push this before the next release.

larray/core/session.py
```diff
@@ -94,8 +96,9 @@ def __init__(self, *args, **kwargs):
         self.meta = meta
 
         if len(args) == 1:
+            assert len(kwargs) == 0
             a0 = args[0]
-            if isinstance(a0, str):
+            if isinstance(a0, basestring):
```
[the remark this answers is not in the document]
gdementen: this comment still applies

larray/core/session.py
```diff
@@ -1036,7 +1037,12 @@ def opmethod(self, other):
                             except Exception:
                                 res_item = nan
                     res.append((name, res_item))
-            return Session(res)
+            try:
+                # XXX: print a warning?
```
gdementen, line 1041: I think we had discussed this in the issue or this PR but I can't remember what the outcome was.

larray/tests/test_constrained_session.py
```diff
@@ -0,0 +1,703 @@
...
+    cs = TestConstrainedSession(a, a01, b=b_alt, a2=a2, e=e, f=f, g=g, h=h)
+    assert cs.b is b_alt
+
+    # test for "NOT_LOADED" variables
+    not_loaded_variables = ['a', 'a2', 'a01', 'e', 'g', 'f', 'h']
+    with pytest.warns(UserWarning) as caught_warnings:
+        cs = TestConstrainedSession()
+    assert len(caught_warnings) == len(not_loaded_variables)
+    for i, var_name in enumerate(not_loaded_variables):
+        assert caught_warnings[i].message.args[0] == f"No value passed for the declared variable '{var_name}'"
```
gdementen, line 76: I **think** a warning will be annoying. I think we should either allow this silently or raise an Exception. I don't know which.

larray/tests/test_constrained_session.py
```diff
@@ -0,0 +1,703 @@
...
+    cs['h'] = 5
+
+    # trying to set a ConstrainedArray variable using an array with axes in different order -> OK
+    cs['h'] = h.transpose()
+
+    # trying to set a ConstrainedArray variable using an array with wrong axes -> should fail
+    # a) missing axis
+    expected_error_msg = "Array 'h' was declared with axes {a, b} but got array with axes {a} ({b} axis is missing)"
+    with pytest.raises(ValueError) as error:
+        cs['h'] = ndtest(a3)
```
gdementen, line 190: We should either allow this (and upscale/broadcast the array) or not allow scalars to fill an array.
I think we discussed this somewhere.

gdementen (commented):
There were at least some review comments not handled (see this review comments) but I only remember this whole PR very vaguely. Too long ago, sorry.

larray/core/session.py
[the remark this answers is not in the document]
alixdamman: I think that was linked to some failures of the tests for Python 2.7 but since Python 2.7 is no longer supported, I can revert this changes.

larray/core/session.py
alixdamman: The discussion starts [here](https://github.com/larray-project/larray/issues/832#issuecomment-610850615) in the corresponding issue thread #832.

The outcome was to [let it as it is implemented now](https://github.com/larray-project/larray/issues/832#issuecomment-618399372).

larray/tests/test_constrained_session.py
alixdamman: If I remember well the problem with raising an Exception was that any code trying to build a `ConstrainedSession` by loading several files will fail (assuming the variables of such session are stored in separated input files).
   
If we choose to allow this silently, my biggest fear is that users may face obscure error messages when they will try to make any operation on a not loaded variable.

We might first allow this silently and then wait for users to complain.

alixdamman: What about replacing
```python
NOT_LOADED = object()
```
by 
```python
class NOT_LOADED(object):
    pass
```
?

This can make error messages when using unloaded variables more explicit (at least for both of us if we need to help them to debug).

larray/tests/test_constrained_session.py
alixdamman: I can't retrieve where we have discussed this but now you say it, it makes sense to me to allow broadcasting (= accept missing axes) to fill an array. 
But I would not agree to allow extra axes. In that case, the array is then no longer "constrained".

alixdamman:
Can I exceptionally rebase my local branch and push force ?

larray/tests/test_constrained_session.py
alixdamman: I removed the error if missing axes if commit 435159541216b1d052457540fc6ee48fda12b9bf

larray/tests/test_constrained_session.py
```diff
@@ -0,0 +1,704 @@
...
+Incompatible axis for array 'h':
+Axis(['a0', 'a1', 'a2', 'a3', 'a4'], 'a')
+vs
+Axis(['a0', 'a1', 'a2', 'a3'], 'a')"""
+    with pytest.raises(ValueError) as error:
+        cs['h'] = h.append('a', 0, 'a4')
+    assert str(error.value) == expected_error_msg
+
+    # trying to set a ConstrainedArray variable using an array with wrong dtype -> print a warning
+    with pytest.warns(UserWarning) as caught_warnings:
```
alixdamman: I removed the warning message in case of wrong dtype in commit cbff52c6c844398485114f2969bf21d706db31dc (too much annoying)

alixdamman:
FWIW, I have tested the current version of ConstrainedSession and ConstrainedArray in the DEMO model

.travis.yml
```diff
@@ -35,8 +35,10 @@ install:
     #   except scipy and install pandas with --no-deps
     # - pandas >= 0.20.0 is required since commit 01669f2024a7bffe47cceec0a0fd845f71b6f7cc
     #   (issue 702 : fixed bug when writing metadata using HDF format)
+    # - pydantic: cannot define numpy ndarray / pandas obj / LArray field with default value
```
gdementen, line 38: Are we blocked on an older version forever or is it a temporary problem?

README.rst
```diff
@@ -129,6 +129,12 @@ For plotting
 - `matplotlib <http://matplotlib.org/>`__:
   required for plotting.
 
+Miscellaneous
+~~~~~~~~~~~~~
+
+- `pydantic <https://github.com/samuelcolvin/pydantic>`__:
+  required to use `ConstrainedSession`.
```
gdementen, line 136: We probably want to add it to larrayenv, no?

larray/core/constrained.py
```diff
@@ -0,0 +1,533 @@
...
+        @classmethod
+        def __get_validators__(cls):
+            # one or more validators may be yielded which will be called in the
+            # order to validate the input, each validator will receive as an input
+            # the value returned from the previous validator
+            yield cls.validate
+
+        @classmethod
+        def validate(cls, value, field: ModelField):
+            # XXX: not sure we should accept a scalar as value but this allows to not repeat axes
```
gdementen, line 45: Is this comment still valid? scalar as values seem fine to me.

larray/core/constrained.py
```diff
@@ -0,0 +1,533 @@
...
+                # check compatible axes
+                try:
+                    cls.expected_axes.check_compatible(value.axes)
+                except ValueError as error:
+                    error_msg = str(error).replace("incompatible axes", f"Incompatible axis for array '{field.name}'")
+                    raise ValueError(error_msg)
+
+            # broadcast if needed
+            array = empty(axes=cls.expected_axes, dtype=cls.dtype)
+            array[:] = value
```
gdementen, line 69: This extra copy is both a good thing (some people find this more intuitive) and a bad thing (a tad less performant but more importantly could lead to unexpected errors given that the behavior is different than for "normal" sessions which only hold references to arrays). Unsure how to solve this (or if it needs solving at all) but this particular point will at least need to be documented explicitly (we possibily need to clarify the situation for normal Session too). 

Also, when is validate called? In the code like below, would it make an extra copy for each line?
```python
s.arr = zeros(...)
s.arr = s.arr + 1
```
If this is the case, can't we imagine a system where if the attribute was never initialized, it uses the current code path using empty, but if it was, only the check is done but not the copy?

larray/core/constrained.py
```diff
@@ -0,0 +1,533 @@
...
+    # from the types.py module of the 'pydantic' library
+
+    def ConstrainedArray(axes: AxisCollection, dtype: np.dtype = float) -> Type[Array]:
+        # XXX: for a very weird reason I don't know, I have to put the fake import below
+        #      to get autocompletion from PyCharm
+        from larray.core.constrained import ConstrainedArrayImpl
+        """
+        Represents a constrained array.
+        Its axes are assumed to be "frozen", meaning they are constant all along the execution of the program.
+        A constrain on the dtype of the data can be also specified.
```
gdementen, line 82: constraint

larray/core/constrained.py
```diff
@@ -0,0 +1,533 @@
...
+                if name in d:
+                    del d[name]
+            return d
+
+    class Parameters(ConstrainedSession):
+        """
+        Same as py:class:`ConstrainedSession` but:
+
+            * declared variables cannot be modified after initialization
+            * adding undeclared variables after initialization is forbidden.
```
gdementen, line 462: Unsure this second point should be allowed in ConstrainedSession either. What does your experience with playing with it gives in this regard?

larray/core/session.py
```diff
@@ -1013,7 +1014,7 @@ def __len__(self):
 
     # binary operations are dispatched element-wise to all arrays (we consider Session as an array-like)
     def _binop(opname, arrays_only=True):
-        opfullname = f'__{opname}__'
+        opfullname = '__%s__' % opname
```
gdementen, line 1017: Why this change? bad rebase?

larray/core/session.py
```diff
@@ -1061,7 +1067,7 @@ def opmethod(self, other):
     # element-wise method factory
     # unary operations are (also) dispatched element-wise to all arrays
     def _unaryop(opname):
-        opfullname = f'__{opname}__'
+        opfullname = '__%s__' % opname
```
gdementen, line 1070: ?

larray/tests/test_constrained_session.py
gdementen: I am fine with any solution as long as we give ourselve the opportinity to revisit this choice later. Makes me think that we might want to label the new feature "experimental" for one or two releases.

gdementen (commented):
Looks nice to me overall. Two things worry me though: the pydantic version and the systematic copy on each setattr (AFAIU).

.travis.yml
alixdamman: Hard to say. I opened an [issue ](https://github.com/samuelcolvin/pydantic/issues/2923) on the pydantic github repository. 
`pydantic` is still currently in active development but I don't how they manage their issues.

README.rst
alixdamman: Hum... Yes, indeed.
I already added pydantic to the list of packages in the function update_metapackage()

larray/core/constrained.py
alixdamman: I guess not. Since I allowed broadcasting, I don't see any reason to not accept scalar.

README.rst
gdementen: Sorry, I missed that 😞

larray/core/constrained.py
alixdamman: My idea was to "force" our users to declare all parameters in one place (discipline! discipline! discipline!).
But on the other hand, our users don't like to be forced (even encouraged) to follow good practices.
Not sure if it is actually a true good practice (I learned it was but is it true?).

Now after a second thought, I guess the best thing to do considering our users is to be more permissive in the implementation of LArray but showing the best way to use LArray objects in the tutorial.

larray/core/session.py
alixdamman: Yeah, sorry

larray/core/session.py
alixdamman: bad rebase

larray/core/constrained.py
alixdamman: Good point.
I'll make several tests to see if I can come with something better than this extra copy.

alixdamman: I just pushed a new version of `ConstrainedArrayImpl.validate()`. I still need to create an extra copy in case of missing axes but maybe there is another way in case missing axes to avoid to create an extra copy that I don't see.

larray/tests/test_constrained_session.py
gdementen: I forgot to comment about this... I think the best situation would be to check if the value dtype can be safely casted to the "destination" dtype and if so, just do it (no warning at all) but if it's not the case (e.g. when storing float values in an integer array) either warn about it or even raise an exception.

larray/core/constrained.py
gdementen: No, there is no way to add axes without copying: you cannot broadcast "inplace". FWIW, there is one exception to this, it is if one accepts the resulting array to be readonly (which .expand() can do as an option) but we clearly cannot use that option here.

larray/tests/test_constrained_session.py
gdementen: np.can_cast should be useful for this, I think

alixdamman:
> 
> 
> maybe your difficulty comes from the name of the method? 
Yes
> It would be clearer (for me at least) if it was called broadcast or something like that but I thought it would be more intuitive/clearer for our users to use a less technical term.
I give you the point.

alixdamman:
Do you know why Travis tests are no longer launched automatically in PRs ?

gdementen:
> Do you know why Travis tests are no longer launched automatically in PRs ?

No... No idea, sorry. 😞

gdementen:
> Do you know why Travis tests are no longer launched automatically in PRs ?

Probably due to: https://blog.travis-ci.com/2021-05-07-orgshutdown

gdementen:
I migrated our CI to travis-ci.com (instead of .org). The problem is we got a one-time 10000 credits (not sure what those correspond to) and they will probably run out quickly.

alixdamman:
closed & reopen to trigger a Travis build

alixdamman:
Travis is installing a very old version of pydantic (0.18.2). I need to check why.

alixdamman:
As a final discussion, I wonder if using a pejorative word such as `constrained` is a good idea considering our users. Even it is not orthodox, I would consider to hide the "constrained" by simply prefixing `Array` and `Session` with a `C`, i.e. `CArray` instead of `ConstrainedArray` and `CSession` instead of `ConstrainedSession`.

I guess we need the opinion of someone else from our team this time. What do you think ?

larray/core/constrained.py
```diff
@@ -327,7 +327,7 @@ class ConstrainedSession(Session, AbstractConstrainedSession, metaclass=ModelMet
         >>> # here the result of the multiplication of the 'population' array by the 'mortality_rate' array
         >>> # is automatically converted to an integer array (printing a warning message)
         >>> m.deaths = m.population * m.mortality_rate
-        >>> print(m.deaths.info)
+        >>> print(m.deaths.info)                            # doctest: +SKIP  (linux uses dtype=int64 instead of int32)
```
gdementen: wouldn't using np.int64 (or int32) explicitly when creating the array, instead of "int" solve the problem without having to skip the test?

larray/core/constrained.py
```diff
@@ -327,7 +327,7 @@ class ConstrainedSession(Session, AbstractConstrainedSession, metaclass=ModelMet
         >>> # here the result of the multiplication of the 'population' array by the 'mortality_rate' array
         >>> # is automatically converted to an integer array (printing a warning message)
```
gdementen, line 328: is it still the case?

larray/core/constrained.py
```diff
@@ -327,7 +327,7 @@ class ConstrainedSession(Session, AbstractConstrainedSession, metaclass=ModelMet
...
         >>> m.deaths = m.population * m.mortality_rate
-        >>> print(m.deaths.info)                            # doctest: +SKIP  (linux uses dtype=int64 instead of int32)
+        >>> print(m.deaths.info)                            # doctest: +SKIP
```
gdementen, line 330: couldn't we use np.int64 instead?

gdementen (commented):
> As a final discussion, I wonder if using a pejorative word such as `constrained` is a good idea considering our users. Even it is not orthodox, I would consider to hide the "constrained" by simply prefixing `Array` and `Session` with a `C`, i.e. `CArray` instead of `ConstrainedArray` and `CSession` instead of `ConstrainedSession`.
> 
> I guess we need the opinion of someone else from our team this time. What do you think ?

I don't think it's pejorative. It just says what it does... but we can **try** to find another name if you do not like it. I am not a fan of C*. What about Fixed* or Checked*? Asking other users about this is fine with me too.

alixdamman:
> What about Fixed* or Checked*? 

Checked* sounds the best choice to me.

larray/core/constrained.py
alixdamman: To not afraid our users. Lot of them have no idea what ``np.x`` means.

larray/core/constrained.py
alixdamman: see my comment above

larray/core/constrained.py
alixdamman: About the warning message ? 
No, indeed. Nice catch.

alixdamman:
One last review and merge as it is ?
This feature is really urgent for at least 2 of our in-house projects (demo + horblk) and must be included in the next release.

gdementen (approved):
LGTM overall. Congratulation for the huge work! Please address as many of my comments as you feel are necessary and merge without asking further review.

doc/source/api.rst
```diff
@@ -821,6 +820,30 @@ Load/Save
...
+
+CheckedSession
+==============
+
+.. autosummary::
+   :toctree: _generated/
+
+   CheckedSession
+
+Parameters
```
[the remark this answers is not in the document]
alixdamman: Now you say it, I find it more consistent too.

gdementen:
given the test failures, I guess you need to introduce a "needs_pydantic" decorator

alixdamman:
> you could try (if not done already) to install pydantic via pip

It worked. Thanks.

alixdamman:
@gdementen OK with the changelog ? (last commit)

gdementen:
@alixdamman: sure. I personally wouldn't explain what autocompletion is but... 🤷‍♂️

PR #843, opened 2020-01-09
842 dump and load scalars of sessions

alixdamman:


larray/inout/hdf.py
```diff
@@ -70,46 +70,54 @@ def read_hdf(filepath_or_buffer, key, fill_value=nan, na=nan, sort_rows=False, s
...
+                if name == 'None':
+                    name = None
+                key = pd_obj.values
+                if 'dtype_kind' in attrs and attrs['dtype_kind'] == 'U':
+                    key = np.char.decode(key, 'utf-8')
+                axis = read_hdf(filepath_or_buffer, attrs['axis_key'])
+                res = LGroup(key=key, name=name, axis=axis)
+        except KeyError:
+            scalar_key = key
+            attrs = store.root._v_attrs
```
gdementen: I would **MUCH** rather have scalars stored in a dedicated group (e.g. `__scalars__`) like for Axis and Group objects, than in the root node. AFAICT, this would already avoid conflicts for scalars named "metadata", "writer" or "type". Also, I would prefer (but I do **NOT** know if it is practical to do so!) if scalars were each stored as a standalone object (probably either a 0d array, or a 1d array of len 1) instead of being stored all as **attributes** of the root node. AFAICT, this would solve all other remaining name conflicts (with "system" attributes -- _v_attrnamessys) and would be conceptually cleaner: a scalar is some kind of data, not *meta*data, like attributes are.

doc/source/changes/version_0_33.rst.inc
```diff
@@ -49,7 +49,8 @@ New features
 Miscellaneous improvements
 ^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-* improved something.
+* scalar objects (i.e of type int, float, bool, string, data, time or datetime) belonging to a session
```
gdementen: s/data/date/ 😄

doc/source/changes/version_0_33.rst.inc
```diff
@@ -49,7 +49,8 @@ New features
...
+  are now also loaded and saved when using the HDF5 or pickle format (closes :issue:`842`).
```
gdementen: saved and loaded? (you first need to save them to be able to load them 😉)

larray/core/session.py
```diff
@@ -455,31 +469,48 @@ def save(self, fname, names=None, engine='auto', overwrite=True, display=False,
 
         Examples
         --------
+        >>> # scalars
+        >>> i, s = 5, 'string'
         >>> # axes
-        >>> a, b = Axis("a=a0..a2"), Axis("b=b0..b2")    # doctest: +SKIP
+        >>> a, b = Axis("a=a0..a2"), Axis("b=b0..b2")
```
gdementen: there is a reason all these doctests were skipped! This is because they depend on pytables which is an optional dependency!

larray/core/session.py
```diff
@@ -369,46 +372,57 @@ def load(self, fname, names=None, engine='auto', display=False, **kwargs):
         --------
         In one module:
 
+        >>> # scalars
+        >>> i, s = 5, 'string'
         >>> # axes
-        >>> a, b = Axis("a=a0..a2"), Axis("b=b0..b2")    # doctest: +SKIP
+        >>> a, b = Axis("a=a0..a2"), Axis("b=b0..b2")
```
gdementen: there is a reason all these doctests were skipped! This is because they depend on pytables which is an optional dependency!

larray/inout/hdf.py
gdementen: From the description of the "obj" argument of the table.Array class (at http://www.pytables.org/usersguide/libref/homogenous_storage.html#tables.Array), it seems like Python scalars might be natively supported by pytables...

larray/core/session.py
alixdamman: I removed (temporarily) the `# doctest: +SKIP` just for testing the outputs of the examples. 
I simply forgot to put them back before to commit. Sorry.

gdementen: no problem then

alixdamman: I didn't forget again about this. I will put back the `# doctest: +SKIP` just before to merge the PR.

larray/inout/common.py
```diff
@@ -1,11 +1,20 @@
...
 from collections import OrderedDict
 
+from larray.util.compat import basestring
 from larray.core.array import Array
 
 
+# only for HDF5 and pickle formats
+# XXX: also include scalars in Excel and CSV formats (via __scalars__ special sheet/.csv file)?
+#      But then why saving scalars but not Axis and Group objects. That would be inconsistent.
+# XXX: support list and dict?
```
alixdamman: @gdementen the two questions above are still open. What do you think?

gdementen: Q1: Even though I can see some value in the ability to save/dump scalars to Excel/CSV files even if groups/axes are not (even if that would be odd indeed), because that could be used to make some kind of control panel in an excel file to run a model, which would make for easy "GUI" for some models. **BUT** doing this via a Session might not be the best API (or at least this needs to be thought through) and in either case, IMO it does not need to be linked to this PR as I consider that lower priority.
Q2: That would be nice too, but also something with much lower priority IMO.

alixdamman: Q2: list and tuple objects are already supported with the current implementation of `PandasHDFHandler` and `PickleHandler`. But in that case, ``__scalars__`` is no longer a valid name...

gdementen: ??? They might be nearly supported but not entirely AFAICT as they do not satisfy any of the isinstance checks _dump_item, or am I missing something? Adding support for that might not be hard but requires additional tests, etc... In short it would required more time, so I would rule that as out of scope for now, but that's your call.

alixdamman: For the lists and tuples, I just have to add `list, tuple` to the `_supported_scalar_types` protected variable. It works (I already added the tests in test_session.py). But in that case, using `__scalars__` is wrong. What do you think about `__non_larray__`?

gdementen: Now that I think of it, having groups in `__groups__` and axes in `__axes__` for HDF files does not seem to be useful (since we are guaranteed not to have name collisions -- because they come from the same session -- and we store the object type in node attributes anyway), or is it? So I would either store everything as leafs on the root node (like arrays) (*), or create a group node per "broad" type: __axes__, __groups__, __scalars__, __sequences__, __mappings__, ...

(*) the inevitable HDF file format change due to #761 might be the opportunity to make this change too.

alixdamman: I would go for the first option (i.e. store everything as leafs on the root node).
Now, I'm thinking of it, we could use HDF groups to store several Session objects in the same HDF file?

gdementen: That's certainly possible and could be nifty, but I would like to avoid yet another feature nobody uses, and I don't see a practical use case for our current users yet, so **unless you see one**, I would wait until somebody asks for it (or until we find some real case where it would actually make things more practical).

gdementen: As a side note, moving everything to the root node has the disadvantage that we have to support two distinct file formats when reading HDF files (but since we will need that for #761 anyway, this is the right time to do it).

alixdamman: About `list`, `tuple` and `dict`, I understand your remark so I will only implement save/load of scalars.

larray/inout/hdf.py
```diff
@@ -136,46 +159,34 @@ def _open_for_write(self):
...
-        elif type == 'scalar':
-            hdf_key = '__scalars__/' + key
         else:
             raise TypeError()
         return read_hdf(self.handle, hdf_key, *args, **kwargs)
 
     def _dump_item(self, key, value, *args, **kwargs):
-        if isinstance(value, Array):
+        if isinstance(value, _supported_scalars_types):
+            value = ScalarHDF(value)
```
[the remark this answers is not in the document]
alixdamman: old reflex coming from the time I learned JAVA. I will change that.

larray/inout/hdf.py
```diff
@@ -73,53 +98,51 @@ def read_hdf(filepath_or_buffer, key, fill_value=nan, na=nan, sort_rows=False, s
...
+                name = str(pd_obj.name)
+            if name == 'None':
+                name = None
+            key = pd_obj.values
+            if 'dtype_kind' in attrs and attrs['dtype_kind'] == 'U':
+                key = np.char.decode(key, 'utf-8')
+            axis = read_hdf(filepath_or_buffer, attrs['axis_key'])
+            res = LGroup(key=key, name=name, axis=axis)
+        elif _type in {cls.__name__ for cls in _supported_scalars_types}:
+            res = pd_obj.values
```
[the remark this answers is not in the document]
alixdamman: Now it is there, let's finish this. I don't know when I will go back to work on PR #761 (partly because I don't like to work with pytables and prefer h5py but my intuition tells me we need pytables for our case).

larray/inout/hdf.py
```diff
@@ -13,11 +13,36 @@
...
+        _type = type(value).__name__
+        if not isinstance(value, _supported_scalars_types):
+            raise TypeError("Type {} which is not currently supported by the HDF5 format".format(_type))
+        self.value = value
+        self._type = _type
+
+    def to_hdf(self, filepath, key):
+        key = _translate_group_key_hdf(key)
+        s = pd.Series(data=self.value)
+        with LHDFStore(filepath) as store:
```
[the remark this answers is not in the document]
alixdamman: correct

larray/inout/hdf.py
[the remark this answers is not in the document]
gdementen: I thought h5py had the same features than pytables with a different API, except for querying (and optionally indexing) tables. And since we wouldn't need either feature, I would think using h5py would not be a problem. That said I never used h5py so I don't know for sure.

In either case, I wouldn't use h5py anyway because pandas relies on pytables and since we will have to keep the current code (using pytables indirectly) for backward compat for a while, depending on both pytables and h5py would be weird.

alixdamman:
@gdementen It is me or Travis-CI has some weird problem with Python 2.7? Sounds like Travis-CI fails to even create a conda environment.

gdementen:
> 
> 
> @gdementen It is me or Travis-CI has some weird problem with Python 2.7? Sounds like Travis-CI fails to even create a conda environment.

I don't think it has anything to do with Travis. It seems like conda dependencies on Python2.7 are not declared correctly (oh the irony). Conda 4.8 seems to depend on tqdm but the dependency is not installed along with the conda package (so I guess the dependency wasn't declared), while tqdm is installed automatically on python3+. A workaround would probably be to install tqdm explicitly before "conda update conda"

larray/inout/common.py
```diff
@@ -18,6 +18,8 @@
 _supported_scalars_types = (int, float, bool, basestring, date, time, datetime)
 _supported_types = _supported_larray_types + _supported_scalars_types
 _supported_typenames = {cls.__name__ for cls in _supported_types}
+if PY2:
+    _supported_typenames.add('str')
```
gdementen: This seems fishy. I think the problem is that we should not use basestring in this context (or maybe any context?) and use bytes and unicode individually instead.

alixdamman: why not just replace basestring by str? It should be enough, right?

gdementen: because then you don't support the same thing on Python2 and Python3, str on python2 is equivalent to bytes on Python3. That's why I would have used "unicode" instead (until we drop support for Python2)

alixdamman:
OK with the last commit?

gdementen:
sure

alixdamman:
So, OK to merge?

alixdamman:
Thanks for the review.

PR #846, opened 2020-01-31
fixed using Pandas >= 1.0 (closes #845)

### LINKED ISSUE #845: fix problems with pandas 1.0
Pandas 1.0 was just released and it breaks a lot of our test suite.

gdementen:


alixdamman:
The changes made in this commit are still compatible with the previous versions of pandas?
Shouldn't we update the condition of pandas version in the ``travis.yml`` file and in the ``setup.py`` module?

gdementen:
Yes, it is still compatible: Index and MultiIndex have been part of pandas (top-level) public API since forever (and we already used them from there in other places of the code). So, no, we don't need to update the pandas version anywhere. What we **might** want to do is release a 0.32.2 version of larray with the fix included. It all depends on when we release version 0.33 but it seems like it will be a while until we do that, and in the mean time larray will be broken any time a user does a conda update or when installing larray in an empty environment (where pandas 1.0 will be picked by default).

alixdamman:
Making a 0.32.2 release seems a good idea to me. It doesn't cost a lot to do it I think. 
However, I'm not sure we should send a broadcast email to inform our users when the release is done.

gdementen:
indeed

PR #848, opened 2020-02-21
Drop Python 2 support

alixdamman:


larray/util/compat.py
```diff
@@ -1,58 +1,21 @@
-import sys
 
 try:
     # the abstract base classes were moved to the abc sub-module in Python 3.3 but there is a backward compatibility
     # layer for Python up to 3.7
     from collections.abc import Iterable, Sequence
 except ImportError:
-    # needed for Python < 3.3 (including 2.7)
```
gdementen: AFAIK, the whole block can go, as we don't support python 3.2 anymore

larray/util/compat.py
```diff
@@ -1,58 +1,21 @@
...
 
-try:
-    import builtins
-except ImportError:
-    import __builtin__ as builtins
-
 try:
     from itertools import izip
 except ImportError:
     izip = zip
```
gdementen: the izip thing can also go, as it was for Python2

larray/util/compat.py
```diff
@@ -1,58 +1,21 @@
...
-
-def csv_open(filename, mode='r'):
-    assert 'b' not in mode and 't' not in mode
-    if PY2:
-        return open(filename, mode + 'b')
-    else:
-        return open(filename, mode, newline='', encoding='utf8')
-
 
 def decode(s, encoding='utf-8', errors='strict'):
```
gdementen: this one is trickier, but I think it the code using this should either be refactored to make this function unnecessary (unsure it's possible!!), or this function should be moved elsewhere as there should not be any compat module anymore

larray/core/session.py
```diff
@@ -303,22 +321,15 @@ def __setitem__(self, key, value):
...
             raise AttributeError("'{}' object has no attribute '{}'".format(self.__class__.__name__, key))
 
-    # TODO: implement meta.setter when Python 2.7 will be dropped
     def __setattr__(self, key, value):
         if key == 'meta':
-            if not isinstance(value, (list, dict, OrderedDict, Metadata)):
-                raise TypeError("Expected list of pairs or dict or OrderedDict or Metadata object "
-                                "instead of {}".format(type(value).__name__))
-            object.__setattr__(self, '_meta', value if isinstance(value, Metadata) else Metadata(value))
+            super().__setattr__(key, value)
```
gdementen: I don't think this code is necessary/will ever be executed as I **think** `__setattr__` will only be called if there is no method nor property with that name on the class. It is worth double checking though.

alixdamman: I tried but it seems that line is necessary : https://stackoverflow.com/a/15751159

gdementen: Yuk! I didn't know (or forgot) about the asymmetry between `__getattr__` and `__setattr__`. Please add a comment about this.

larray/core/group.py
```diff
@@ -608,7 +601,7 @@ def _to_keys(value, stack_depth=1):
     >>> _to_keys('P01;P02,P03;:')
     ('P01', ['P02', 'P03'], slice(None, None, None))
 
-    # evaluated variables do not work on Python 2, probably because the stack depth is different
+    # evaluated variables do not work
```
gdementen: this must either be fixed or moved out of the docstring

larray/tests/test_array.py
```diff
@@ -4557,7 +4557,6 @@ def test_diag():
     assert d.i[1] == 1.0
 
 
-@needs_python36
```
gdementen: since python3.6 is now the lowest supported version, you can also drop the needs_python36 decorator everywhere (including its definition)

alixdamman:
OK to merge?

gdementen:
yes, LGTM modulo squashing

PR #860, opened 2020-04-06
0.32.2

gdementen:
foward port 0.32.2 fixes to master

### DIFF
--- a/doc/requirements.txt
+++ b/doc/requirements.txt
@@ -1,0 +1,12 @@
+numpy
+pandas
+matplotlib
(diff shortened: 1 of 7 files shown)


gdementen:
this PR shouldn't be merged until I do the corresponding PR on releaser, but I don't have time to do this today.

gdementen:
also this PR should be cleaned-up (pandas fix changelog should be only on 0.32.2)

gdementen:
this relies on larray-project/releaser#32

PR #890, opened 2020-08-28
(issue 872): implemented isscalar() function

alixdamman:


larray/core/ufuncs.py
```diff
@@ -302,3 +303,36 @@ def broadcastify(func):
...
+    Returns
+    -------
+    bool
+        `True` if `element` is a scalar type, `False` if it is not.
+
+    Examples
+    --------
+    >>> isscalar(3.1)
+    True
+    >>> isscalar(np.array(3.1))
```
gdementen, line 326: I wouldn't show this in the doctest (too obscure) for our users. On the other hand, a test with an **larray** scalar array could be interesting mostly to avoid having to add the information about that corner case to the description or the "Returns" paragraph **but** I would move it much lower in the list, because it is a corner case.

larray/core/ufuncs.py
```diff
@@ -302,3 +303,36 @@ def broadcastify(func):
      a0   6   2   5   0
      a1   5   2   3   0
 """
+
+
+def isscalar(element: Any) -> bool:
```
gdementen, line 308: is isscalar an ufunc (numpy universal function)? I don't think it is, so moving this to another module would be a good idea I think.

alixdamman: Correct. I move it to util/misc.py or do I create a new module in the `core` directory?

gdementen: AFAIR, util/misc.py are for internal functions not api-exposed functions, so I wouldn't put it there. Maybe core/misc.py?

alixdamman: OK for core/misc.py

gdementen:
this PR LGTM, what's blocking it?

PR #892, opened 2020-08-28
(issue 874): implemented Axis.min() and Axis.max() methods

alixdamman:


gdementen (commented):
I would add either an explanation **or** an example that it fails on non-numeric labels.

gdementen:
what's blocking this PR (except failed tests?)

larray/core/axis.py
```diff
@@ -614,6 +614,60 @@ def equals(self, other):
...
+        --------
+        Fails on non-numeric labels.
+
+        Examples
+        --------
+        >>> time = Axis('time=1991..2020')
+        >>> time.min()
+        1991
+
+        >>> country = Axis('country=Belgium,France,Germany')
```
gdementen, line 636: isn't having the Warning AND a failed example a bit overkill? Isn't the warning enough by itself?

larray/core/axis.py
```diff
@@ -614,6 +614,60 @@ def equals(self, other):
...
+        --------
+        Fails on non-numeric labels.
+
+        Examples
+        --------
+        >>> time = Axis('time=1991..2020')
+        >>> time.max()
+        2020
+
+        >>> country = Axis('country=Belgium,France,Germany')
```
gdementen, line 663: idem

larray/core/axis.py
alixdamman: Depends on what users actually read. Some of them may only read the examples. 
It doesn't hurt to have both IMO.

gdementen:
LGTM then

PR #893, opened 2020-08-28
(issue 880): implemented Axis.astype() method 

alixdamman:


larray/core/axis.py
```diff
@@ -234,6 +236,60 @@ def by(self, length, step=None, template=None):
...
+
+        Parameters
+        ----------
+        dtype: str or dtype
+            Typecode or data-type to which the labels are cast.
+
+        casting: str, optional
+            Controls what kind of data casting may occur. Defaults to ‘unsafe’.
+
+                * ‘no’ means the data types should not be cast at all.
```
gdementen, line 251: I don't really like those weird quotes. **IF** we start using them, we should use them everywhere. I am -0 on this but if you really want that, go ahead and make an issue or PR to do that change.

larray/core/axis.py
```diff
@@ -234,6 +236,60 @@ def by(self, length, step=None, template=None):
...
+                * ‘equiv’ means only byte-order changes are allowed.
+                * ‘safe’ means only casts which can preserve values are allowed.
+                * ‘same_kind’ means only safe casts or casts within a kind, like float64 to float32, are allowed.
+                * ‘unsafe’ means any data conversions may be done.
+
+        inplace: bool, optional
```
gdementen, line 257: I **WONDER** whether or not we should implement any inplace operation in larray. For now, we implement only very very few of them and I feel like we should either implement that in as many as make sense or not at all. I agree that this one is more likely to be useful but well...

gdementen:
what is blocking this PR? It seems ready to me (except for the probably unrelated test failure)

alixdamman:
> what is blocking this PR? 

Something was still wrong. 
Last pushed commit solved it.

gdementen:
LGTM. Ideally, the following line in test_array.py could probably use the new method:

`arr = arr.set_axes([Axis(a.labels.astype(object), a.name) for a in arr.axes])`
could be changed to:
`arr = arr.set_axes([a.astype(object) for a in arr.axes])`

PR #894, opened 2020-09-02
(issue 889): replaced pytest-pep8 by pytest-flake8

alixdamman:


larray/tests/test_session.py
```diff
@@ -7,7 +7,8 @@
 import pandas as pd
 import pytest
 
-from larray.tests.common import (assert_array_nan_equal, inputpath, tmp_path, meta,
+from larray.tests.common import meta            # noqa: F401
```
gdementen, line 10: why do we need to import meta if it is unused?

gdementen (commented):
LGTM

larray/tests/test_session.py
alixdamman: `meta` is a [pytest fixture](https://docs.pytest.org/en/stable/fixture.html#fixtures-as-function-arguments) (`meta` is defined in the /tests/common.py module) and is actually used as argument in many unittests. 
It's just that flake8 is confused with pytest fixtures.

larray/tests/test_excel.py
```diff
@@ -299,12 +295,12 @@ def test_excel_report_setting_template():
 def test_excel_report_sheets():
     report = ExcelReport()
     # test adding new sheets
-    sheet_population = report.new_sheet('Population')
-    sheet_births = report.new_sheet('Births')
-    sheet_deaths = report.new_sheet('Deaths')
+    sheet_population = report.new_sheet('Population')       # noqa: F841
```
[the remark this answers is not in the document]
alixdamman: Good point. We actually don't need to store the result in a variable.

alixdamman:
Thanks for the review

Issue #887, opened 2020-08-07
document that .append also works when the axis already exists + deprecate .extend


gdementen:
```python
>>> arr = ndtest(3)
>>> arr
a  a0  a1  a2
    0   1   2
>>> arr.append('a', 42, 'a3')
a  a0  a1  a2  a3
    0   1   2  42
>>> arr2 = Array([42], 'a=a3')
>>> arr.extend('a', arr2)
a  a0  a1  a2  a3
    0   1   2  42
>>> arr3 = Array([42, 43], 'a=a3,a4')
>>> arr.extend('a', arr3)
a  a0  a1  a2  a3  a4
    0   1   2  42  43
>>> arr.append('a', arr3)
a  a0  a1  a2  a3  a4
    0   1   2  42  43
```

gdementen:
FWIW, I did a little poll about this for our heaviest users. So far the results are:
jd +1, rd +1, bm -0

Issue #429, opened 2017-09-18
Mixing slices and individual labels to select along the same axis


alixdamman:
The idea is to be able to select a subset as follow:
```python
>>> arr = ndtest(10)
>>> arr
        a  a0  a1  a2  a3  a4  a5  a6  a7  a8  a9
             0    1    2    3    4   5    6    7    8   9
>>> arr["a0:a4,a7"]
        a  a0  a1  a2  a3  a4  a7  a8  a9
              0   1    2    3   4    7    8    9
```
Is there any good reason why it doesn't work up to now?

gdementen:
this is basically a duple of #360. There is a reason but I have no time to explain that now.

alixdamman:
That reason is technical or philosophical? (this issue comes from the demand of one our current user)

gdementen:
I would say neither. It is because it has many syntax implications on other stuff and I never had both the time AND a clear enough mind to think this through. It is related to #23 and the whole constellation of related issues and how to make our API future-proof.

```python
>>> a = Axis('a=a0..a4')
>>> # when using the string syntax, it seems natural that:
>>> a['a0:a2,a4']
a[['a0', 'a1', 'a2', 'a4']]
>>> # and
>>> a['a0:a2;a4']
(a['a0':'a2'], a['a4'])
>>> # BUT I want to have a way to do it easily using the non string syntax AND keep both syntaxes consistent
>>> a['a0':'a2', 'a4']
a[['a0', 'a1', 'a2', 'a4']]
OR? 
(a['a0':'a2'], a['a4'])
```
and I have been leaning on the second one because we can easily add something to go from multiple groups to a single group, but not the opposite. But given the inconsistency with the string syntax, this yielded a deadlock. 

The question is thus: is there an alternative to make creating several groups containing slices on the same axis easily?

FWIW, for non slices, this is a non-issue:

```python
>>> a[['a0', 'a1', 'a2'], ['a4']]
(a['a0', 'a1', 'a2'], a['a4'])
>>> a[['a0', 'a1', 'a2', 'a4']]
a[['a0', 'a1', 'a2', 'a4']]
```

gdementen:
One alternative would be to use a "slice builder" (e.g S) and move subgroups in lists, but this feels too hard/unatural to my tastes:
```python
>>> a[[S['a0':'a2']], ['a4']]
(a['a0':'a2'], a['a4'])
>>> a['a0':'a2', 'a4']
[...]

PR #914, opened 2021-02-17
Tutorial: miscellaneous

alixdamman:
slight changes or improvements here and there

### DIFF
--- a/doc/source/tutorial/getting_started.ipyml
+++ b/doc/source/tutorial/getting_started.ipyml
@@ -28,6 +28,35 @@ cells:
 
 
 - markdown: |
+    <div class="alert alert-warning">
+    
+    **Note:** 
+    The tutorial is generated from Jupyter notebooks which work in the "interactive" mode (like in the LArray Editor console). 
+    In the interactive mode, there is no need to use the print() function to display the content of a variable.
+    Simply writing its name is enough.
+    The same remark applies for the returned value of an expression.<br><br>
+    In a Python script (file with .py extension), you always need to use the print() function to display the content of 
+    a variable or the value returned by a function or an expression.
+        
+    </div>
+
+
+- code: |
+    s = 1 + 2
+    
+    # In the interactive mode, there is no need to use the print() function 
+    # to display the content of the variable 's'.
+    # Simply typing 's' is enough
+    s
+
+
+- code: |
+    # In the interactive mode, there is no need to use the print() function 
+    # to display the result of an expression
+    1 + 2
+
+
+- markdown: |
     ## Create an array
 
 
@@ -40,9 +69,9 @@ cells:
 
 - code: |
     # define some axes to be used later
-    age = Axis(['0-9', '10-17', '18-66', '67+'], 'age')
-    gender = Axis(['female', 'male'], 'gender')
-    time = Axis([2015, 2016, 2017], 'time')
+    age = Axis('age=0-9,10-17,18-66,67+')
+    gender = Axis('gender=female,male')
+    time = Axis('time=2015..2017')
 
 
 - markdown: |
@@ -430,16 +459,6 @@ cells:
 
 
 - markdown: |
-    <div class="alert alert-warning">
-        If you are using a Python version prior to 3.6, you will have to pass a list of pairs
-        to the Session constructor otherwise the arrays will be stored in an arbitrary order in
-        the new session. For example, the session above must be created using the syntax:
-        
-        `demography_session=Session([('population', population), ('births', births), ('deaths', deaths)])`.
-    </div>
-
-
-- markdown: |
     One of the main interests of using sessions is to save and load many arrays at once:
 
 
@@ -536,7 +555,7 @@ metadata:
     name: python
     nbconvert_exporter: python
     pygments_lexer: ipython3
-    version: 3.7.3
+    version: 3.7.7
 nbformat: 4
 nbformat_minor: 2
 
--- a/doc/source/tutorial/tutorial_aggregations.ipyml
+++ b/doc/source/tutorial/tutorial_aggregations.ipyml
@@ -86,6 +86,35 @@ cells:
     population.sum(gender, (benelux, fr_de))
 
 
+- markdown: |
+    <div class="alert alert-warning">
+    
+    **Warning:** Mixing slices and individual labels inside the `[ ]` will generate **several groups** (a tuple of groups) instead of a single group.<br>If you want to create a single group using both slices and individual labels, you need to use the `.union()` method (see below).
+      
+    </div>
+
+
+- code: |
+    # mixing slices and individual labels leads to the creation of several groups (a tuple of groups)
+    except_2016 = time[:2015, 2017]
+    except_2016
+
+
+- code: |
+    # leading to potentially unexpected results
+    population.sum(except_2016)
+
+
+- code: |
+    # the union() method allows to mix slices and individual labels to create a single group
+    except_2016 = time[:2015].union(time[2017])
+    except_2016
+
+
+- code: |
+    population.sum(except_2016)
+
+
 # The lines below here may be deleted if you do not need them.
 # ---------------------------------------------------------------------------
 metadata:
@@ -102,7 +131,7 @@ metadata:
     name: python
     nbconvert_exporter: python
     pygments_lexer: ipython3
-    version: 3.7.3
+    version: 3.7.7
 nbformat: 4
 nbformat_minor: 2
 
--- a/doc/source/tutorial/tutorial_presenting_larray_objects.ipyml
+++ b/doc/source/tutorial/tutorial_presenting_larray_objects.ipyml
@@ -13,6 +13,34 @@ cells:
 
 
 - markdown: |
+    <div class="alert alert-warning">
+        
+    **Note:** 
+    The tutorial is generated from Jupyter notebooks which work in the "interactive" mode (like in the LArray Editor console). 
+    In the interactive mode, there is no need to use the print() function to display the content of a variable.
+    Simply writing its name is enough. The same remark applies for the returned value of an expression.<br><br>
+    In a Python script (file with .py extension), you always need to use the print() function to display the content of 
+    a variable or the value returned by a function or an expression.
+        
+    </div>
+
+
+- code: |
+    s = 1 + 2
+    
+    # In the interactive mode, there is no need to use the print() function 
+    # to display the content of the variable 's'.
+    # Simply typing 's' is enough
+    s
+
+
+- code: |
+    # In the interactive mode, there is no need to use the print() function 
+    # to display the result of an expression
+    1 + 2
+
+
+- markdown: |
     ## Axis
     
     An ``Axis`` represents a dimension of an Array object.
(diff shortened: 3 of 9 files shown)


alixdamman:
From the logs from Travis, I can see that the newer version of pandas has broken LArray...

doc/source/tutorial/tutorial_presenting_larray_objects.ipyml
```diff
@@ -364,12 +369,6 @@ cells:
     gender = Axis("gender=Male,Female")
     time = Axis("time=2013..2017")
     
-    # create and populate a new session in one step
-    # Python <= 3.5
-    demography_session = Session([('gender', gender), ('time', time), ('population', zeros((gender, time))),
```
gdementen, line 356: Note that this syntax is also useful when you have objects with names which are not valid Python identifiers (such as labels starting with a number). This probably needs to be mentioned somewhere, but probably not here.

gdementen (commented):
- tutorial: use print() everywhere: -1 on this. We can add a note about this/explain this at the start if this isn't the case already but users need to understand that rather sooner than later because most Python ressources on the net use that syntax.

- tutorial: use string syntax to generate axes in the getting_started: +0
- tutorial: removed the warning box about how to define Session objects…: +1
- tutorial -> tutorial_presenting_larray_objects notebook:
  - use the string syntax to declare axes:  +0 in general but a **-1** for the "age axis with strings" change as is. I think it is a good idea to show the consequence of having a mixed axis (need to index with strings) **but** it should contain some kind of explanation that it is a special case when you have a mixed axis. Without an explanation , this is going to confuse more than anything IMO. Also, I wonder if this is the best place to introduce that.
  - removed other axis (it is unlikely that users will mix letters and numbers when generating labels with the syntax start..end): +0
  - do not use numpy randint() function to generate fake data (to avoid to confuse readers): +1
  - avoid to define anonymous axes when calling function to create arrays with predefined values (confusing): +1
  - removed example explaining how to create Session objects with Python 3.5- (versions 3.5 or less are too old): +1 but see inline comment.
- tutorial -> tutorial_combine_arrays notebook:
  - removed subsection about extend() (redundant with append()/append()/insert()): strong **-1** as this is really not the same operation. We should probably make it so (#887) but it is not the case yet, and even if it was documented, keeping an example where the axis exist is a must IMO.
  - added example using argument after instead of before with the insert() method: +1

alixdamman:
>     * tutorial: use print() everywhere: -1 on this. We can add a note about this/explain this at the start if this isn't the case already but users need to understand that rather sooner than later because most Python ressources on the net use that syntax.

OK. OK. I did that because 2 of our users send me an email to complain about the fact that lines like:
```python
population.info
```
don't print anything when they run their scripts. They think that everything in the tutorial is transferable to their scripts as it is.
But you're right when you say "users need to understand that rather sooner than later because most Python ressources on the net use that syntax". I will reverse that commit and add a note somewhere.

>     * tutorial -> tutorial_presenting_larray_objects notebook:
>       
>       * use the string syntax to declare axes:  +0 in general but a **-1** for the "age axis with strings" change as is. I think it is a good idea to show the consequence of having a mixed axis (need to index with strings) **but** it should contain some kind of explanation that it is a special case when you have a mixed axis. Without an explanation , this is going to confuse more than anything IMO. Also, I wonder if this is the best place to introduce that.

OK

>     * tutorial -> tutorial_combine_arrays notebook:
>       
>       * removed subsection about extend() (redundant with append()/append()/insert()): strong **-1** as this is really not the same operation. We should probably make it so (#887) but it is not the case yet, and even if it was documented, keeping an example where the axis exist is a must IMO.

Hum. Well. What does really extend() is obscure to me and I can't find an example where extend() cannot be replaced by either insert() or append() or prepend(). To tell the truth, I never use it when I have to help someone.

doc/source/tutorial/tutorial_presenting_larray_objects.ipyml
alixdamman: Using a dictionary instead of a list of pairs would be clearer. Anyway, it is difficult and maybe not a good idea to show any pitfall or very specific cases in the tutorial. IMO, the tutorial is big enough right now to not add (very) specific cases (at least this one).

alixdamman:
Do you have an example in mind where the use of `extend()` cannot be replaced by either `insert()`, `prepend()` or `append()` ?

doc/source/tutorial/tutorial_presenting_larray_objects.ipyml
gdementen: Indeed a dict is better for Python3.7+, my bad. My comment was only valid for 3.6, but since we are doing away with it too, you can indeed simply remove that. Concerning special/tricky cases I think a tutorial is meant precisely to show those which users are likely to stumble into and how to "work around" them. In this case, I don't know how likely it is. I have seen array labels starting with numbers quite a few time, but never in a session, so removing the block is fine I guess.

alixdamman: Your comment above makes me think that I may add something about how to create a group mixing separated labels and a slice (using Group.union()). What do you think ?

doc/source/tutorial/getting_started.ipyml
```diff
@@ -27,6 +27,17 @@ cells:
     __version__
 
 
+- markdown: |
+    <div class="alert alert-warning">
+        The tutorial is generated from Jupyter notebooks which works in "interactive" mode (like in the LArray Editor).
```
gdementen, line 32: * s/works/work
* LArray Editor console?

doc/source/tutorial/getting_started.ipyml
```diff
@@ -27,6 +27,17 @@ cells:
     __version__
 
 
+- markdown: |
+    <div class="alert alert-warning">
+        The tutorial is generated from Jupyter notebooks which works in "interactive" mode (like in the LArray Editor). 
+        In the "interactive" mode, there is no need to use the function print() to display the content of a variable.
```
gdementen, line 33: s/the "interactive"/interactive/
s/the function print()/the print() function
also add a mention of expressions?

doc/source/tutorial/getting_started.ipyml
```diff
@@ -27,6 +27,17 @@ cells:
...
+        Simply writing its name is enough.
+        The function print() is only need if the content of more than one variable has to be displayed.<br><br>
```
gdementen, line 35: I would remove this sentence (as it is not entirely true), you can display a tuple in the console:
```python
>>> a = 1; b = 2
>>> a, b
(1, 2)
```

doc/source/tutorial/getting_started.ipyml
```diff
@@ -27,6 +27,17 @@ cells:
...
+        In a Python script (file with .py extension), you always need to use the print() function to display the content of 
+        a variable or the value returned by a function.
```
gdementen, line 37: add a mention of expressions?

gdementen:
> OK. OK. I did that because 2 of our users send me an email to complain about the fact that lines like:
> 
> ```python
> population.info
> ```
> 
> don't print anything when they run their scripts. They think that everything in the tutorial is transferable to their scripts as it is.

I know, I already had that question too.

gdementen:
Concerning the print() function, I think it would help if you mentioned 
1) an explicit example. Something like: "For example, we did not need to use a print() function to display the content of the __version__ variable above".
2) explicit explanation that when copy pasting code from the tutorial in a script file, you need to add those print() calls.

gdementen:
> Do you have an example in mind where the use of `extend()` cannot be replaced by either `insert()`, `prepend()` or `append()` ?

Probably not something user-friendly, because .append does work in that case (by accident -- this is the whole point of #887) but that's not the point here. The point is that those two methods were **meant to** do (and are documented as doing) different things and even if we can do both "things" using the same method, we should show both in the tutorial. So either you leave it as is (using .extend) or you tackle #887 first/at the same time and change that .extend call to .append.

gdementen:
PS: the difference between larray.append and .extend was meant to be exactly the same as the difference between list.append and .extend

doc/source/tutorial/tutorial_presenting_larray_objects.ipyml
gdementen: A good idea (unless you plan to tackle #429 "soonish", **in which case** it would be better to tackle it first to show the nice syntax directly) but I don't know what's the best place to add that...

doc/source/tutorial/tutorial_presenting_larray_objects.ipyml
```diff
@@ -106,6 +119,24 @@ cells:
...
+- markdown: |
+    <div class="alert alert-warning">
+    
+    **Warning:** Mixing slices and individual labels inside the `[ ]` will generate a **tuple of groups** instead of unique group.<br>If you want to create a unique group using both slices and individual labels, you need to use the `.union()` method (see below).   
+      
+    </div>
+
+
+- code: |
+    # mixing a slices and individual labels leads to the creation of a tuple of groups
```
gdementen, line 131: s/a slices/slices/
s/a tuple of/several distinct/ ?

doc/source/tutorial/tutorial_presenting_larray_objects.ipyml
```diff
@@ -106,6 +119,24 @@ cells:
     teens
 
 
+- markdown: |
+    <div class="alert alert-warning">
+    
+    **Warning:** Mixing slices and individual labels inside the `[ ]` will generate a **tuple of groups** instead of unique group.<br>If you want to create a unique group using both slices and individual labels, you need to use the `.union()` method (see below).
```
gdementen, line 125: s/a tuple of/several distinct/ ?
s/of unique/of a single/
s/a unique/a single/

doc/source/tutorial/getting_started.ipyml
```diff
@@ -29,15 +29,27 @@ cells:
...
+    Simply writing its name is enough.
+    The same remark applies for the returned value of an expression.<br><br>
+    In a Python script (file with .py extension), you always need to use the print() function to display the content of 
+    a variable or the value returned by a function or an expression.
+        
     </div>
 
 
+- code: |
+    s = "There is no need to use the print() function to display the content of the variable 's'."
```
gdementen, line 45: couldn't having the message within the variable be confusing?

doc/source/tutorial/getting_started.ipyml
```diff
@@ -29,15 +29,27 @@ cells:
...
+    s
+
+
+- code: |
+    "There is no need to use the print() function " + "to display the returned value of an " + "expression"
```
gdementen, line 50: Use something simpler?

```
>>> 1 + 2
3
```

doc/source/tutorial/tutorial_aggregations.ipyml
```diff
@@ -86,6 +86,35 @@ cells:
     population.sum(gender, (benelux, fr_de))
 
 
+- markdown: |
+    <div class="alert alert-warning">
+    
+    **Warning:** Mixing slices and individual labels inside the `[ ]` will generate a **tuple of groups** instead of unique group.<br>If you want to create a unique group using both slices and individual labels, you need to use the `.union()` method (see below).
```
gdementen, line 92: s/a tuple of/several distinct/ ?
s/of unique/of a single/
s/a unique/a single/

doc/source/tutorial/tutorial_aggregations.ipyml
```diff
@@ -86,6 +86,35 @@ cells:
...
+
+
+- code: |
+    # mixing a slices and individual labels leads to the creation of a tuple of groups 
+    except_2016 = time[:2015, 2017]
+    except_2016
+
+
+- code: |
+    # leading to unexpected results
```
gdementen, line 104: to potentially unexpected?

doc/source/tutorial/tutorial_aggregations.ipyml
```diff
@@ -86,6 +86,35 @@ cells:
...
+    population.sum(except_2016)
+
+
+- code: |
+    # the union() method allows to mix slices and individual labels to create a unique group
```
gdementen, line 109: single

gdementen:
LGTM, it's merging season 😉

gdementen:
btw the "the" you added were not typos. I think it's correct too but it's superfluous. In general, "the" is used less in English than "le" in French.

PR #917, opened 2021-02-19
(issue 915): replaced xrld by openpyxl to read .xlsx files when xlwings is not installed

alixdamman:


alixdamman:
add a log in the release notes for this ?

alixdamman:
Now, it's `_to_ticks` and `test_deprecated_methods` which fail for Python 3.7+

alixdamman:
Ready for review

gdementen:
> add a log in the release notes for this ?

I don't know. That's the kind of things other opensource projects would document but our users don't want to know.

larray/core/axis.py
```diff
@@ -3186,7 +3186,7 @@ def combine_axes(self, axes=None, sep='_', wildcard=False, front_if_spread=False
                     combined_labels = _axes[0].labels
                 else:
                     sepjoin = sep.join
-                    axes_labels = [np.array(label, np.str, copy=False) for label in _axes.labels]
+                    axes_labels = [np.array(label, str, copy=False) for label in _axes.labels]
```
gdementen, line 3189: This change on your part is correct (it is equivalent to the old deprecated code) but this smells like an error in the old code. I always thought np.str was numpy's native string type 😢. I will have to investigate this.

larray/core/group.py
```diff
@@ -399,18 +399,18 @@ def _to_ticks(s, parse_single_int=False):
...
-    array(['A', 'C', 'D', 'E', 'F', 'G', 'Z'], dtype='<U1')
-    >>> _to_ticks('U')                          # doctest: +NORMALIZE_WHITESPACE
-    array(['U'], dtype='<U1')
-    >>> _to_ticks('..3')                        # doctest: +NORMALIZE_WHITESPACE
-    array([0, 1, 2, 3])
-    >>> _to_ticks('01..12')                     # doctest: +NORMALIZE_WHITESPACE
-    array(['01', '02', '03', '04', '05', '06', '07', '08', '09', '10', '11', '12'], dtype='<U2')
-    >>> _to_ticks('01,02,03,10,11,12')          # doctest: +NORMALIZE_WHITESPACE
-    array(['01', '02', '03', '10', '11', '12'], dtype='<U2')
+    >>> list(_to_ticks('M , F'))                # doctest: +NORMALIZE_WHITESPACE
```
gdementen, line 402: do you know what broke those?

larray/inout/excel.py
```diff
@@ -185,7 +185,7 @@ def read_excel(filepath, sheet=0, nb_axes=None, index_col=None, fill_value=nan,
 
     Extract array from a given range (xlwings only)
 
-    >>> read_excel(fname, 'population_births_deaths', range='A9:E15')
```
gdementen: It is probably doable. See "pandas solution" in:
https://stackoverflow.com/questions/43327975/how-can-i-read-a-rangea5b10-and-place-these-values-into-a-dataframe-using-o
but not worth it at the moment. **Maybe** add an issue for this?
**Maybe** add a note in read_excel range argument docstring that it is not available using openpyxl engine?

larray/inout/excel.py
```diff
@@ -22,7 +23,6 @@
 __all__ = ['read_excel']
 
 
-# We use "# doctest: +SKIP" for all tests because they work only if xlrd (an *optional* dependency) is installed
```
gdementen, line 25: the doctest +SKIP should stay, for the same reason (openpyxl is also an optional dependency)

larray/inout/excel.py
```diff
@@ -226,12 +228,15 @@ class PandasExcelHandler(FileHandler):
     """
     def __init__(self, fname, overwrite_file=False):
         super(PandasExcelHandler, self).__init__(fname, overwrite_file)
+        _, ext = os.path.splitext(fname)
+        self.read_engine = 'xlrd' if ext == '.xls' else 'openpyxl'
+        self.write_engine = 'xlwt' if ext == '.xls' else 'openpyxl'
```
gdementen, line 233: Doesn't pandas default to xlsxwriter for writing? At least it used to be the case (see https://stackoverflow.com/questions/51571026/engines-available-for-to-excel-function-in-pandas). We don't necessarily have to use the same default, but there might be a good reason for that (I don't really know). From the code below (_dump_item), it seems like we used it by default too (???), so do you have a good reason to change that?

larray/core/axis.py
alixdamman: Here is the deprecation warning from numpy 1.20:
```
DeprecationWarning: `np.str` is a deprecated alias for the builtin `str`. To silence this warning, use `str` by itself. Doing this will not modify any behavior and is safe. If you specifically wanted the numpy scalar type, use `np.str_` here.
Deprecated in NumPy 1.20; for more details and guidance: https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations
```

larray/core/group.py
alixdamman: With numpy 1.20, the dtype associated with the returned array by _to_ticks() differs according to the Python version and the machine! Sometimes it is '<U2', sometimes '<U21' or even 'U<11'.

larray/inout/excel.py
alixdamman: But openpyxl is in .travis.yml ?

larray/inout/excel.py
alixdamman: As I understand, 'xlsxwriter' does not support the old '.xls' format.
But OK, I can write:
```python
self.write_engine = 'xlwt' if ext == '.xls' else None
```

larray/inout/excel.py
alixdamman: I create a new issue for this

alixdamman: > Maybe add a note in read_excel range argument docstring that it is not available using openpyxl engine

It is already the case ;)
```
    range : str, optional
        Range to load the array from (only supported for the 'xlwings' engine). (...)
```

larray/inout/excel.py
gdementen: I am unsure of the consequence of None, I just wondered why you did not use:
```python
self.write_engine = 'xlwt' if ext == '.xls' else 'xlsxwriter'
```

larray/inout/excel.py
gdementen: No idea but I guess so.

alixdamman: FWIW, I reverted the remove of the comment above + the remove of the "# doctest: SKIP" directives.

My question is: Since an optional dependency is actually installed when Travis creates the test environment, why do we have to keep the "# doctest: SKIP" directive ?

larray/inout/excel.py
alixdamman: Because `xlsxwriter` is an optional dependency. One may have installed `openpyxl` but not `xlsxwriter`. 
Or `xlsxwriter` is somehow not an optional dependency?

gdementen: well I consider it optional too.

alixdamman:
I tried to use skipif() decorator from pytest to allow testing doctests of the `read_excel()` function but it doesn't work for doctests.

gdementen:
It should be possible with some ugly code (inside the test):
https://docs.pytest.org/en/stable/doctest.html#skipping-tests

alixdamman:
> 
> 
> It should be possible with some ugly code (inside the test):
> https://docs.pytest.org/en/stable/doctest.html#skipping-tests

From your link:
> However using those functions is discouraged because it reduces the readability of the docstring.

I prefer to keep the SKIP directive.

gdementen:
LGTM % some squashing 😉

alixdamman:
Thanks for the review

PR #924, opened 2021-03-01
clean_getitem

gdementen:
a first batch of cleanups and performance improvements related to getitem and friends

there is still a **lot** of work in that domain but it is progressing very slowly given I do most of the work during my free time

### DIFF
--- a/larray/core/array.py
+++ b/larray/core/array.py
@@ -447,20 +447,11 @@ class ArrayPointsIndexer(object):
     def __init__(self, array):
         self.array = array
 
-    def _prepare_key(self, key, wildcard=False):
-        axes = self.array.axes
-        # 1) complete key & translate (those two cannot be dissociated because to complete
-        #    the key we need to know which axis each key belongs to and to do that, we need to
-        #    translate the key to indices)
-        translated_key = axes._translated_key(key)
-        # 2) transform keys to IGroup and non-Array advanced keys to Array with a combined axis
-        return axes._adv_keys_to_combined_axis_la_keys(translated_key, wildcard)
-
     def __getitem__(self, key):
-        return self.array.__getitem__(self._prepare_key(key), translate_key=False)
+        return self.array.__getitem__(key, points=True)
 
     def __setitem__(self, key, value):
-        self.array.__setitem__(self._prepare_key(key, wildcard=True), value, translate_key=False)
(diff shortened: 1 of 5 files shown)


larray/tests/test_excel.py
```diff
@@ -330,6 +330,7 @@ def test_excel_report_titles():
 
 
 @needs_xlwings
+@needs_pytables
```
alixdamman, line 333: At which step `pytables` is required by the `ExcelReport` class ?

alixdamman, line 333: At which step `pytables` is required by the `ExcelReport` class ?

alixdamman (commented):
Two questions and one suggestion.

larray/tests/test_excel.py
gdementen: because it loads example data which is stored in hdf

larray/core/array.py
```diff
@@ -6423,7 +6423,7 @@ def drop(self, labels=None):
         axis = group.axis
         indices = axis.index(group)
         axis_idx = self.axes.index(axis)
-        new_axis = Axis(np.delete(axis.labels, indices), axis.name)
+        new_axis = axis._drop_indices(indices)
```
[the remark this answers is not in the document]
gdementen: It does not really, I admit. I thought it should be a (public) method on Axis when I saw this code and refactored the thing then I realized it was using indices and not labels and backpedalled partially and made it a private method. So yeah, I'll revert that.

larray/core/axis.py
```diff
@@ -2883,10 +2844,34 @@ def _key_to_raw_and_axes(self, key, collapse_slices=False, translate_key=True):
...
+
+            # extract axis from Group keys
+            key_items = [(k1.axis, k1) for k1 in igroup_key]
+
+            # even keys given as dict can contain duplicates (if the same axis was
+            # given under different forms, e.g. name and AxisReference).
+            dupe_axes = list(duplicates(axis1 for axis1, key1 in key_items))
+            if dupe_axes:
+                dupe_axes = ', '.join(str(axis1) for axis1 in dupe_axes)
+                raise ValueError("key has several values for axis: %s\n%s" % (dupe_axes, key_items))
```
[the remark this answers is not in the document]
gdementen: thanks for spotting this... (bad rebase)

Issue #900, opened 2020-11-10
revisit ExcelReport api


gdementen:
@alixdamman: I just realized there was something odd about the ExcelReport API: users have to "register" things to dump in a delayed fashion then call "to_excel" to dump them all in one shot, for no good reason anymore.

That odd API was inherited from the proof of concept code I did that you used, but there, it was done for a performance reason. The goal was to be as fast as possible by joining all the data parts in a single big "data" array and dump the data sheet in one shot, thus limiting as much as possible the win32 api calls (it is much faster to do one big dump than many small ones). I don't remember how much time this saved but it was significant. There are still the many api calls to create the graphs, so I assume it is not **that** big either.

But what I don't like is that we now have that odd API without the speed advantage. So **if** we don't want the speed boost, I think we could make something more immediate and ideally better integrated with the normal "open_excel" API, instead of having a different "workbook" object (ExcelReport) and a different "sheet" object (ReportSheet), just for reports. Besides, some of the report functionality (dump several arrays after each other without having to specify where to dump each one) would be welcome in "normal" Excel sheets  (#826).

Issue #826, opened 2019-11-26
dump several arrays after each other on a sheet without specifying position


gdementen:
current, when you want to dump several tables below each other you have to provide precise positions. It would be nice to have a way to dump arrays below each other.

see demo/table_generator

gdementen:
we have more and more users who either asked for this explicitly or would benefit from this

gdementen:
FWIW, I found two workarounds:

```python
sheet[sheet.shape[0], 0] = something_to_write
# OR
sheet[sheet.used_range.rows.count, 0] = something_to_write
```

PR #926, opened 2021-03-29
(issue 900): implemented Range.make_plot()

alixdamman:
TODO @alixdamman : add method documentation + update Tutorial

### DIFF
--- a/larray/inout/xw_excel.py
+++ b/larray/inout/xw_excel.py
@@ -1,6 +1,8 @@
 import os
 import atexit
 
+from typing import Union, Dict, Callable, Any
+
 import numpy as np
 try:
     import xlwings as xw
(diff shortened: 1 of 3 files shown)


gdementen:
I **wonder** if it wouldn't be more logical to invert position and range. In test_make_plot that would be:
```python
sheet["B8"].make_plot("B2")
```

Also, the dump and make_plot calls look alien to each other. One is a method on the source, the other on the target.

```python
sheet["B2"] = population_be.dump()
sheet["B2"].make_plot("B8")
```

It would be better to have similar syntaxes for both. Either:
```python
sheet["B2"] = population_be.dump()
sheet["B8"] = population_be.plot(data="B2")  # <-- bad syntax because population_be is not really used UNLESS we make it dump the data too
```
or 
```python
sheet["B2"].add_table(population_be)
sheet["B8"].add_plot(data_source="B2")
```

gdementen:
implementing add_table and add_plot could be a way to tackle #826, #900, and #967.

larray/inout/xw_excel.py
```diff
@@ -806,19 +809,15 @@ def open_excel(filepath=None, overwrite_file=False, visible=None, silent=None, a
...
- a2   6   7   8
+>>> arr, arr2 = ndtest((3, 3)), ndtest((2, 2))
 
 create a new Excel file and save an array
 
 >>> # to create a new Excel file, argument overwrite_file must be set to True
 >>> with open_excel('excel_file.xlsx', overwrite_file=True) as wb:   # doctest: +SKIP
 ...     wb['arr'] = arr.dump()
-...     wb['arr']['A6'].add_plot('A1', title='simple graph')
+...     wb['arr']['A6'].add_table()
```
gdementen, line 819: isn't an argument missing to add_table?

larray/inout/xw_excel.py
```diff
@@ -806,19 +809,15 @@ def open_excel(filepath=None, overwrite_file=False, visible=None, silent=None, a
...
+...     wb['arr']['G1'].add_plot('A1', title='simple graph')
```
gdementen, line 820: This probably works, but I personally prefer using explicit data range instead of just the top cell. Both should be shown in the doctest though I think, because it's nice to know that it works.

gdementen:
* Shouldn't this be documented in the tutorial?
* Shouldn't add_table create Excel tables (potentially as an option)? I remember adding some code which could be useful for that in the mcafee manager.

alixdamman:
> Shouldn't add_table create Excel tables (potentially as an option)?

What do you mean by "Excel tables" ?

gdementen:
to be closer to a replacement for the ExcelReport API, we could allow passing an array as the datasource and add a data_sheet (or dump_data_to or whatever it is called) argument to add_plot;

```python
sheet["B8"].add_plot(data_source=population_be, data_sheet="data")
```
would be equivalent to:

```python
datasheet = workbook[data_sheet]
pos = (datasheet.shape[0], 0)
datasheet[pos] = population_be.dump()
sheet["B8"].add_plot(data_source=datasheet[pos])
```

gdementen:
Tentatively added to 0.34 milestone (using the assumption that I take over your code), unsure I will have time to do that.

alixdamman:
> Tentatively added to 0.34 milestone (using the assumption that I take over your code), unsure I will have time to do that.

OK. Please do.

PR #940, opened 2021-07-29
flake8_warnings

gdementen:


alixdamman (approved):
Long but necessary work. LGTM.

gdementen:
Honestly, I am unsure about E201 (whitespace after '[') and E241 (multiple spaces after ','), because there are so many of them. For the other warnings I am pretty happy with what I did.

I see pro and con for having them inline.

Pro:
* noqa is closer to the code, than a per-file ignore in setup.cfg, so if it no longer needs to be applied, there is more chance we actually think to remove it.
* it is more local, so we do not accidentally ignore other "mistakes" in the same file.

Con:
* kinda annoying to apply everywhere. But let's test how it goes.

In summary, I just reserve the right to change my mind on those two if it's too annoying in practice 😄.

PR #946, opened 2021-07-30
Fixci

gdementen:


gdementen:
hmmm, now it always uses the latest python version 😢 (3.9) instead of using the version matrix.

gdementen:
I gave up trying to not use conda-incubator/setup-miniconda. I seems to be kinda painful to do without it **IF** we want to keep using conda to install packages during tests.

The alternative would be to install dependencies using pip instead, which shouldn't be too hard, but since we/our users mostly use conda, I guess we should use that for the tests too to get environments as close as what the real world looks like for us.

gdementen:
Now that I think about it again (maybe too late), the main problem was that I couldn't find a way to activate the test environment within the action. But I probably could simply do like the basic setup-python command and use the "base" environment. Kinda ugly but I don't see any reason it wouldn't work since the other steps worked.

@alixdamman you think it's worth it?

PR #957, opened 2021-08-10
Tutorial (issue 855): Added explanations and examples of CheckedSession, CheckedParameters and CheckedArray

alixdamman:
Future output of readthedocs can be seen [here](https://larray-test.readthedocs.io/en/tutorial/tutorial/getting_started.html) for the Getting Started section and [here](https://larray-test.readthedocs.io/en/tutorial/tutorial/tutorial_sessions.html) for the Working With Sessions section.


doc/source/tutorial/tutorial_sessions.ipyml
```diff
@@ -470,16 +645,18 @@ cells:
...
-    compare(demography_session, demography_session_alternative, names=['baseline', 'lower_birth_rate'])
+    compare(s, s_alternative, names=['baseline', 'lower_birth_rate'])
     ```
     
     ![compare two sessions](../_static/compare_tutorial.png)
 
 
 - markdown: |
-    ## Session API
+    ## (Checked)Session API
```
gdementen, line 655: I know there are the parentheses, but I didn't realize at first this section was valid for (non checked) sessions too. I also **wonder** if the section is actually needed?

doc/source/tutorial/tutorial_sessions.ipyml
```diff
@@ -8,14 +8,115 @@ cells:
...
+    
+    The syntax to define a checked-session is a bit specific:
+    
+    ```python
+    class MySession(CheckedSession):
+        # Variables can be declared in two ways:
+        # a) by specifying only the type of the variable (to be initialized later)
+        var1: Type
+        # b) by giving a default value
+        var2 = initialization value
```
gdementen, line 91: shouldn't you mention that the type is autodetected in that case?

doc/source/tutorial/tutorial_sessions.ipyml
```diff
@@ -8,14 +8,115 @@ cells:
...
+
+
+- markdown: |
+    ### CheckedSession
+    
+    The syntax to define a checked-session is a bit specific:
+    
+    ```python
+    class MySession(CheckedSession):
+        # Variables can be declared in two ways:
```
gdementen, line 87: what about `var: Type = value`? Does that work?

doc/source/tutorial/tutorial_sessions.ipyml
```diff
@@ -8,14 +8,115 @@ cells:
...
+        # (convention is to declare parameters (read-only objects) in capital letters)
+        # Declare 'VARIANT' parameter as of type string.
+        # 'VARIANT' will be initialized when a 'Demography' session will be created
+        VARIANT: str
+        # declare variables of type Axis with an initialization value.
+        COUNTRY = Axis('country=Belgium,France,Germany')
+        GENDER = Axis('gender=Male,Female')
+        TIME = Axis('time=2013..2017')
+        # declare 'population' and 'births' arrays with an initialization value.
+        population = zeros([COUNTRY, GENDER, TIME], dtype=int)
```
gdementen, line 110: In this case I assume the axes are not checked, only the fact that the variable is an Array but since the axes are mentioned, this might be confusing. Maybe mention this fact explicitly?

We possibly **MIGHT** also want to autodetect the type/axes of arrays and make them constraints automatically in that case, in the same vein that the Array type is autodetected???

doc/source/tutorial/tutorial_sessions.ipyml
```diff
@@ -8,14 +8,115 @@ cells:
...
+        births = zeros([COUNTRY, GENDER, TIME], dtype=int)
+        # declare 'deaths' with constrained axes and dtype.
+        # Its type (Array), axes and dtype are not modifiable.
+        # It will be initialized with 0
+        deaths: CheckedArray([COUNTRY, GENDER, TIME], int) = 0
```
gdementen, line 115: Doesn't CheckedArray also support axes **names** in addition to real Axis objects? If so, it would probably be a good idea to mention this. I suppose there are case where that can be useful (axes can vary or are defined later).

doc/source/tutorial/tutorial_sessions.ipyml
alixdamman: Well, it's not no longer needed

doc/source/tutorial/tutorial_sessions.ipyml
alixdamman: Yes, indeed

doc/source/tutorial/tutorial_sessions.ipyml
alixdamman: Yes, it works but do we have to mention it?

alixdamman: Yes, it works and without giving the Type you loose the auto-completion on the methods of `var`.

doc/source/tutorial/tutorial_sessions.ipyml
alixdamman: No it doesn't.

doc/source/tutorial/tutorial_sessions.ipyml
```diff
@@ -271,6 +399,53 @@ cells:
         print()
 
 
+- markdown: |
+    ## Manipulating Checked Sessions
+    
+    **Note**: this section only concerns objects declared in checked-sessions.
+    
+    The specificity of  checked-sessions is that the type of the contained objects is protected (it cannot change). Any attempt to assign a value of different type will raise an error:
```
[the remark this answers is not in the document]
alixdamman: > Also, one important goal of CheckSession (IMO) that I don't see mentioned is to make sure you have all the needed (declared) variables in an input file you are loading.

All variables declared with an initialization value will not print any warning message if no data are loaded from the file. 
If no corresponding data is found in the file, then the initialization value is used.

gdementen:
👍

PR #964, opened 2021-08-12
fix950_pandas13

gdementen:


gdementen:
This fix might not be done in the best place. Could be better done in df_aslarray, so that other I/O paths benefit from this too but I am not in the mood to do anything better now 😉.

gdementen:
Since the same array with anonymous axes and integer labels is used in the tests of other I/O paths (csv, etc.), I guess it somehow works and this fix is enough after all, so I am merging this.

PR #975, opened 2021-09-24
Fixed more CheckedSession warning messages stacklevel and make test suite free of warning messages

gdementen:
@alixdamman: I would be grateful if you could try to keep the test suite free of (expected) warnings

### DIFF
--- a/larray/core/checked.py
+++ b/larray/core/checked.py
@@ -383,13 +383,15 @@ else:
         dtype: int32
         memory used: 968 bytes
 
-        >>> # note that it still possible to add undeclared variables to a checked session
-        >>> # but this must be done with caution.
-        >>> m.undeclared_var = 'undeclared_var'
+        It is possible to add undeclared variables to a checked session
+        but this will print a warning:
+
+        >>> m.undeclared_var = 'my_value'                   # doctest: +SKIP
+        UserWarning: 'undeclared_var' is not declared in 'ModelVariables'
 
         >>> # ==== output ====
         >>> # save all variables in an HDF5 file
-        >>> m.save(f'{variant_name}.h5', display=True)
+        >>> m.save(f'{variant_name}.h5', display=True)      # doctest: +SKIP
         dumping birth_rate ... done
         dumping births ... done
         dumping mortality_rate ... done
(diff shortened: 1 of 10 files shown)


gdementen:
The "source"/stacklevel of warnings is now correct in all but one weird case (see the FIXME next to pandas_hdf).

Concerning the CheckedSession warnings, I simply marked all the warnings I received as expected. Maybe some of them were not. I didn't really analyze the code to determine that. Could you check if that seems alright to you?

Also, some tests (data) could be changed to avoid the warnings in the first place but I didn't make that exercice and I don't think it's worth it to do it now.

larray/core/checked.py
```diff
@@ -383,13 +383,15 @@ class CheckedSession(Session, AbstractCheckedSession, metaclass=ModelMetaclass):
         dtype: int32
         memory used: 968 bytes
 
-        >>> # note that it still possible to add undeclared variables to a checked session
-        >>> # but this must be done with caution.
-        >>> m.undeclared_var = 'undeclared_var'
+        It is possible to add undeclared variables to a checked session
+        but this will print a warning:
+
+        >>> m.undeclared_var = 'my_value'                   # doctest: +SKIP
```
alixdamman, line 389: Is the warning tested in test_checkedsession.py ?

larray/core/checked.py
```diff
@@ -501,11 +501,7 @@ def __setstate__(self, state: Dict[str, Any]) -> None:
             object.__setattr__(self, '__dict__', state['__dict__'])
 
         def dict(self, exclude: Set[str] = None):
-            d = dict(self.items())
-            for name in exclude:
-                if name in d:
-                    del d[name]
-            return d
+            return {k: v for k, v in self.items() if k not in exclude}
```
alixdamman, line 508: nice simplification

alixdamman (approved):
I am just asking 2 questions but LGTM.
Thanks for the cleanup of CheckedSession and of the tests associated with Session and CheckSession.

larray/core/checked.py
gdementen: yes (cs.ax = 'ax')

setup.cfg
```diff
@@ -31,6 +31,7 @@ addopts = -v --doctest-modules
           #--cov
 
 filterwarnings =
+    ignore::DeprecationWarning:flake8.*:
```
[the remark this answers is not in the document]
gdementen: To ignore warnings coming from flake8 code itself: it uses some API which has been deprecated in recent versions of another package I am not sure of the name anymore (importlib-metadata or something like that)

PR #978, opened 2021-09-28
Made all I/O functions/methods to accept either a string or a pathlib.Path object for any path argument (issue 896)

alixdamman:


larray/inout/common.py
```diff
@@ -47,7 +48,7 @@ class FileHandler:
         Filename.
     """
     def __init__(self, fname, overwrite_file=False):
-        self.fname = fname
+        self.fname = Path(fname)
```
gdementen, line 51: only convert to Path if not isinstance(fname, Path)?

larray/inout/common.py
```diff
@@ -92,13 +93,12 @@ def close(self):
         raise NotImplementedError()
 
     def _get_original_file_name(self):
-        if self.overwrite_file and os.path.isfile(self.fname):
+        if self.overwrite_file and self.fname.is_file():
             self.original_file_name = self.fname
-            fname, ext = os.path.splitext(self.fname)
-            self.fname = f'{fname}~{ext}'
+            self.fname = self.fname.parent / f'{self.fname.stem}~{self.fname.suffix}'
```
gdementen, line 98: self.fname = self.fname.with_stem(self.fname.stem + '~')
?

larray/inout/csv.py
```diff
@@ -261,43 +261,47 @@ def __init__(self, fname, overwrite_file=False, sep=','):
         if fname is None:
             self.pattern = None
             self.directory = None
-        elif '.csv' in fname or '*' in fname or '?' in fname:
-            self.pattern = fname
-            self.directory = os.path.dirname(fname)
+        if not isinstance(fname, (str, Path)):
```
gdementen, line 264: this will break for fname is None! => use elif instead

also, you could simplify this slightly by moving the if isinstance(fname, str) check above

larray/inout/xw_excel.py
```diff
@@ -83,21 +84,24 @@ def __init__(self, filepath=None, overwrite_file=False, visible=None, silent=Non
...
                         raise ValueError(f"File {filepath} does not exist. Please give the path to an existing file "
                                          f"or set overwrite_file argument to True")
-                    if os.path.isfile(filepath) and overwrite_file:
+                    if filepath.is_file() and overwrite_file:
                         self.filepath = filepath
                         # we create a temporary file to work on. In case of crash, the original is not destroyed.
                         # the temporary file is renamed as the original file at close.
-                        filepath = basename + '~' + ext
-                    if not os.path.isfile(filepath):
+                        filepath = filepath.parent / f'{filepath.stem}~{filepath.suffix}'
```
gdementen, line 103: use with_stem (as above)?

larray/inout/xw_reporting.py
```diff
@@ -16,13 +18,14 @@
...
 
-def _validate_template_filename(filename):
-    filename, ext = os.path.splitext(filename)
-    if not ext:
-        ext = '.crtx'
-    if ext != '.crtx':
-        raise ValueError(f"Extension for the excel template file must be '.crtx' instead of {ext}")
-    return filename + ext
+def _validate_template_filename(filename: Union[str, Path]) -> Path:
+    filename = Path(filename)
```
gdementen, line 22: if not isinstance(..., Path)?

larray/inout/xw_reporting.py
```diff
@@ -41,7 +44,7 @@ def template_dir(self):
 
         Parameters
         ----------
-        template_dir : str
+        template_dir : Path
```
gdementen, line 47: support str too?

larray/inout/xw_reporting.py
```diff
@@ -60,8 +63,8 @@ def template_dir(self):
     @template_dir.setter
     def template_dir(self, template_dir):
         if template_dir is not None:
+            template_dir = Path(template_dir)
```
gdementen, line 66: if not isinstance(..., Path)?

gdementen (approved):
except from the (AFAICT) broken "None" case above, this LGTM

larray/inout/common.py
alixdamman: yes

alixdamman: ``with_stem`` only exists in Python 3.9+

alixdamman:
Could you review the last commit before I rebase on upstream/master ?

larray/inout/common.py
gdementen: oups, sorry

larray/inout/common.py
```diff
@@ -39,16 +41,21 @@ class FileHandler:
 
     Parameters
     ----------
-    fname : str or Path
+    fname : str or Path, optional
```
gdementen, line 44: It is not really optional at this point, as it has no default. It accepts None, which is not the same thing as being optional. Maybe it should be made optional, I don't know.

gdementen (approved):
I don't think asking for review again for something already reviewed once is worth our time (except for very tricky/large PR). Yes, I nearly always find something to improve but it's really details and I don't think it's worth our time

PR #979, opened 2021-10-01
Finalize add of return type hints

alixdamman:
@gdementen I wonder if we shouldn't either define [L|I]GroupKey types in types.py or even make them classes (there are a lot of private functions in group.py ?

### DIFF
--- a/larray/core/array.py
+++ b/larray/core/array.py
@@ -5631,7 +5631,7 @@ class Array(ABCArray):
         fullname = f'__{opname}__'
         super_method = getattr(np.ndarray, fullname)
 
-        def opmethod(self, other):
+        def opmethod(self, other) -> 'Array':
             res_axes = self.axes
 
             if isinstance(other, ExprNode):
@@ -5794,7 +5794,7 @@ class Array(ABCArray):
         fullname = f'__{opname}__'
         super_method = getattr(np.ndarray, fullname)
 
-        def opmethod(self):
+        def opmethod(self) -> 'Array':
             return Array(super_method(self.data), self.axes)
         opmethod.__name__ = fullname
         return opmethod
--- a/larray/core/axis.py
+++ b/larray/core/axis.py
@@ -333,7 +333,7 @@ class Axis(ABCAxis):
             Names of resulting axes. Defaults to None.
         regex : str, optional
             Use regex instead of delimiter to split labels. Defaults to None.
-        labels : bool, optional
+        return_labels : bool, optional
             Whether or not split labels must be returned (as a tuple of tuples). These labels are suitable for indexing
             via array.points[labels]. Defaults to False.
 
--- a/larray/core/axis.py
+++ b/larray/core/axis.py
@@ -4,7 +4,7 @@ import warnings
 from itertools import product
 from collections import defaultdict
 
-from typing import Union, Any
+from typing import Union, Iterable, Tuple, List, Dict, Sequence, Optional
 
 import numpy as np
 import pandas as pd
@@ -17,6 +17,7 @@ from larray.util.oset import OrderedSet
 from larray.util.misc import (duplicates, array_lookup2, ReprString, index_by_id, renamed_to, common_type, LHDFStore,
                               lazy_attribute, _isnoneslice, unique_list, unique_multi, Product, argsort, has_duplicates,
                               exactly_one)
+from larray.util.types import Scalar
 
 
 np_frompyfunc = np.frompyfunc
@@ -103,7 +104,7 @@ class Axis(ABCAxis):
         if name is not None and not isinstance(name, (int, str)):
             nametype = type(name).__name__
             raise TypeError(f"Axis name should be None, int or str but is: {name} ({nametype})")
-        self.name = name
+        self.name: str = name
         self._labels = None
         self.__mapping = None
         self.__sorted_keys = None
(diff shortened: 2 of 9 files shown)


larray/core/group.py
```diff
@@ -76,7 +79,7 @@ def irange(start, stop, step=None):
 _range_bound_pattern = re.compile(r'([0-9]+|[a-zA-Z]+)')
 
 
-def generalized_range(start, stop, step=1):
+def generalized_range(start, stop, step=1) -> Union[range, List[str]]:
```
gdementen, line 82: can be List[int] too

larray/core/group.py
```diff
@@ -194,7 +197,7 @@ def generalized_range(start, stop, step=1):
 _range_str_pattern = re.compile(r'(?P<start>[^\s.]+)?\s*\.\.\s*(?P<stop>[^\s.]+)?(\s+step\s+(?P<step>\d+))?')
 
 
-def _range_str_to_range(s, stack_depth=1):
+def _range_str_to_range(s, stack_depth=1) -> Union[range, List[str]]:
```
gdementen, line 200: can be List[int] too

larray/core/group.py
```diff
@@ -446,7 +449,7 @@ def _to_ticks(s, parse_single_int=False):
 _axis_name_pattern = re.compile(r'\s*(([A-Za-z0-9]\w*)(\.i)?\s*\[)?(.*)')
 
 
-def _seq_str_to_seq(s, stack_depth=1, parse_single_int=False):
+def _seq_str_to_seq(s, stack_depth=1, parse_single_int=False) -> Union[slice, List[Scalar], range, Scalar, None]:
```
gdementen, line 452: I don't see any case where this would return None?

larray/core/group.py
```diff
@@ -590,7 +593,7 @@ def _to_key(v, stack_depth=1, parse_single_int=False):
         raise TypeError(f"{v} has an invalid type ({type(v).__name__}) for a key")
 
 
-def _to_keys(value, stack_depth=1):
+def _to_keys(value, stack_depth=1) -> Sequence[Any]:
```
gdementen, line 596: it can return non-sequences too.

gdementen:
> @gdementen I wonder if we shouldn't either define [L|I]GroupKey types in types.py or even make them classes (there are a lot of private functions in group.py ?

Sorry, I don't understand what you mean.

larray/core/array.py
```diff
@@ -5631,7 +5631,7 @@ def _binop(opname):
         fullname = f'__{opname}__'
         super_method = getattr(np.ndarray, fullname)
 
-        def opmethod(self, other):
+        def opmethod(self, other: Union[Scalar, 'Array']) -> 'Array':
```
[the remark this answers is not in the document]
alixdamman: OK. But I think it could be useful (for me) to add type hints to methods which I don't fully understand how they (are supposed to) behave (`getitem`, `_binop`, all that private functions/methods linked to `getitem` and `aggregate`).

larray/core/axis.py
```diff
@@ -1561,11 +1563,11 @@ def __dir__(self):
         names = set(axis.name for axis in self._list if axis.name is not None)
         return list(set(dir(self.__class__)) | names)
 
-    def __iter__(self):
+    def __iter__(self) -> Iterable[Axis]:
         return iter(self._list)
 
     # TODO: move a few doctests to unit tests
-    def iter_labels(self, axes=None, ascending=True):
+    def iter_labels(self, axes=None, ascending=True) -> Sequence[Tuple[Axis, ...]]:
```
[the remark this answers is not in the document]
alixdamman: Yes. I didn't see the ``.i`` in the line:
```python
p = Product([axis.i for axis in axes])
```

larray/core/axis.py
```diff
@@ -1631,22 +1633,22 @@ def iter_labels(self, axes=None, ascending=True):
...
-    def __getstate__(self):
+    def __getstate__(self) -> List[Axis]:
         return self._list
 
     def __setstate__(self, state):
         self._list = state
         self._map = {axis.name: axis for axis in state if axis.name is not None}
 
-    def __getitem__(self, key):
+    def __getitem__(self, key) -> Union[Axis, 'AxisCollection', Dict[str, Axis]]:
```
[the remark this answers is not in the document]
alixdamman: There is no bug and the method doesn't return a dict. I just have misread the method...

larray/core/group.py
alixdamman: last line is:
```python
return _parse_bound(s, stack_depth + 1, parse_int=parse_single_int)
```
and in `_parse_bound()`, you can read:
```python
    s = s.strip()
    if s == '':
        return None
```

gdementen: ok. As a side note, I wonder if '' should be accepted by the function, but that's not for this PR.

larray/core/array.py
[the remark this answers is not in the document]
gdementen: Makes sense. Ok for arguments of functions (whether internal or not) but not in this PR, please. Also, I still find that typing **internal** attributes is not worth the added maintenance burden and decrease in readability.

alixdamman:
@gdementen could you review the last pushed commit (added type Key: 0dac948d977f87d01c3844b9f24bc5ef7ae6a8f4 ) ?

gdementen:
> 
> 
> @gdementen could you review the last pushed commit (added type Key: [0dac948](https://github.com/larray-project/larray/commit/0dac948d977f87d01c3844b9f24bc5ef7ae6a8f4) ) ?

Except for the failed test & test warnings, this LGTM

alixdamman:
Thanks for the review

PR #993, opened 2021-12-09
more misc stuff

gdementen:


larray/tests/test_array.py
```diff
@@ -2968,9 +2968,25 @@ def test_insert():
     # simple tests are in the docstring
     arr1 = ndtest((2, 3))
 
+    # Insert value without label
+    res = arr1.insert(42, before='b1')
+    expected = from_string(r"""
+    a\b  b0  None  b1  b2
```
alixdamman, line 2974: Does selection on a None label work ?
I made a small test and I got an error:
```python
>>> arr = ndtest((3, 3)).set_labels('b', {'b1': None})
>>> arr[None]
AssertionError: None (<class 'NoneType'>) is not scalar
```

I wonder if it is a good idea to accept None as a label ?

alixdamman (approved):
I have a small question but LGTM

larray/tests/test_array.py
gdementen: You got a point: .insert without a label produces an Array which is only partially functional... 
TL/DR: ... but I would either leave it like that anyway, or provide the string default value for `label` that you will suggest me 😉.

In theory, I think None should be a valid label (I see it as meaning we do not have a name/label for that observation). I see two options to fix this properly: 1) to add support for None as a label everywhere 2) to add some generic mechanism for autogenerated labels (like pandas which disambiguates some labels automatically in some circumstances). Now, in practice, given that there is little chance users actually need this in a model, I am not very eager to actually implement either. 

FWIW, 
* I implemented this because I feel it gives already more feedback than an error in an interactive context, and there are enough cases where you do not care about the exact label that this is already useful.
* you can change None labels back to something useful using setlabels: `arr.set_labels('b', {None: 'empty'})`
* a simpler option would be to provide a (string) default value for label. If you have a proposal of a better default label I am all hears, but anything I could come up felt too arbitrary, confusing, or not that useful because you cannot use it in a selection either (potentially because of current **or future** special characters). `''`, `'None'`, `'N/A'` (the problem with this one is: it  could be understood as the **data** is N/A, not the **label**), `'_'`, `'__'`, `'/'`, `'...'`, ...

gdementen:
I will merge this anyway (helps my work on other branches) but don't hesitate to provide suggestions for a better default value for `label`!

larray/tests/test_array.py
gdementen: Hmmmm, out of the above options, `'_'` seems the best to me. It does not collide with any current or foreseen special character and it has a history within Python to mean something you do not care about. It looks a bit odd in the viewer, but I can live with that. What do you think?

alixdamman: `'_'` is fine to me. If some users don't like how it appears in the viewer, the can simply provide a value to `label` ;)

alixdamman: > Now, in practice, given that there is little chance users actually need this in a model, I am not very eager to actually implement either.

I agree.

PR #997, opened 2022-01-20
tutorial -> merged Presentation LArray Objects section into Getting Started

alixdamman:


doc/source/tutorial/tutorial_presenting_larray_objects.ipyml
```diff
@@ -1,492 +0,0 @@
...
-
-
-- markdown: |
-    ### More on Session objects
-    
-    To know how to save and load sessions in CSV, Excel or HDF format, please refer to the [Loading and Dumping Sessions](tutorial_sessions.ipynb#Loading-and-Dumping-Sessions) section of the tutorial.
-    
-    To see how to work with sessions, please read the [Working With Sessions](tutorial_sessions.ipynb#Working-With-Sessions) section of the tutorial.
-    
-    Finally, see the [Session](../api.rst#session) section of the API Reference to explore all methods of Session objects.
```
gdementen, line 466: There is no link to the API section in Getting started. I would add one.

doc/source/tutorial/tutorial_presenting_larray_objects.ipyml
```diff
@@ -1,492 +0,0 @@
...
-    ones_like(arr)
-
-
-- markdown: |
-    Create an array using the special ``sequence`` function (see link to documention of ``sequence`` in API reference for more examples):
-
-
-- code: |
-    # With initial=1.0 and inc=0.5, we generate the sequence 1.0, 1.5, 2.0, 2.5, 3.0, ...
-    sequence(age, initial=1.0, inc=0.5)
```
gdementen, line 283: I **wonder** if this example is not better than the one in Getting started.

doc/source/tutorial/tutorial_presenting_larray_objects.ipyml
```diff
@@ -1,492 +0,0 @@
...
-- code: |
-    arr.axes
-
-
-- markdown: |
-    Get number of dimensions
-
-
-- code: |
-    arr.ndim
```
gdementen, line 316: AFAICT, .ndim, .shape, .size, .memory_used and .dtype do not appear anywhere in the tutorial anymore (only appear in the API reference), which is a shame. I agree these are a bit more advanced attributes which probably shouldn't be in getting started but I think they should appear _somewhere_ in the tutorial. Where exactly, I don't know though.

gdementen:
Globally, it is a very nice step in the right direction (thanks for doing this!). There are a couple of things which were only present in the removed section (not duplicated) and are thus completely removed from the tutorial, and I think this shouldn't be. The problem is where to put that information. I don't want to overload users with too much information in the Getting started section, so I guess that info should be either be moved to other existing sections if possible (???), or to a new section otherwise.

gdementen (commented):
my previous comment was meant as the global review alongside these comments 😉

doc/source/tutorial/tutorial_presenting_larray_objects.ipyml
alixdamman: There is still a "Miscellaneous" section. Maybe there ?

doc/source/tutorial/tutorial_presenting_larray_objects.ipyml
alixdamman: OK

doc/source/tutorial/tutorial_presenting_larray_objects.ipyml
alixdamman: There is one. Just below the title "## Grouping arrays in a Session".

gdementen: I guess this comment was meant as an answer to the previous thread....

doc/source/tutorial/tutorial_presenting_larray_objects.ipyml
gdementen: Yes, that's the best option I can think of.

doc/source/tutorial/tutorial_presenting_larray_objects.ipyml
gdementen: > There is one. Just below the title "## Grouping arrays in a Session".

It's not very visible IMO. Case in point, I did not see it. I think that an explicit mention like what we had in the "presenting objects" before would help.

> Finally, see the [Session](../api.rst#session) section of the API Reference to explore all methods of Session objects.

gdementen:
LGTM

PR #998, opened 2022-01-27
(issue 988) : fixed comparison between Array and None returning False (single value)

alixdamman:


larray/tests/test_array.py
```diff
@@ -2652,6 +2652,16 @@ def test_binary_ops(small_array):
     assert raw2_ge_la.axes == small_array.axes
     assert_array_equal(raw2_ge_la, raw2 >= raw)
 
+    # arrays filled with None
+    arr = full(small_array.axes, fill_value=None)
+    res = arr == None                                                       # noqa: E711
+    assert_array_equal(res, full(small_array.axes, fill_value=True))
```
gdementen, line 2658: use ones(small_array.axes) instead of full(..., fill_value=True)?

larray/core/array.py
```diff
@@ -5644,7 +5644,9 @@ def opmethod(self, other) -> 'Array':
             # TODO: ndarray should probably be converted to larrays because that would harmonize broadcasting rules, but
             # it makes some tests fail for some reason.
             if not isinstance(other, (Array, np.ndarray)) and not np.isscalar(other):
-                if isinstance(other, list):
+                if other is None:
+                    other = full(self.axes, fill_value=None)
+                elif isinstance(other, (list, Axis)):
```
gdementen, line 5649: using asarray() for the None case should work too I think (it used to work). It should return a "scalar array" (ie an array with 0 axes), which is then broadcasted during the numpy binop, so it should be the same result but computed much more efficiently.

doc/source/changes/version_0_34.rst.inc
```diff
@@ -67,3 +67,6 @@ Fixes
 ^^^^^
 
 * fixed Array.insert when no label is provided (closes :issue:`879`).
+
+* fixed comparison between Array and None returning False instead of an array of boolean values
```
gdementen, line 71: mention Axis too

gdementen (changes requested):

larray/tests/test_array.py
alixdamman: OK but why ?

larray/core/array.py
alixdamman: Are you sure it is computed much more efficiently with asarray() ? 
It sounds counterintuitive (to me at least).

To me, using full() allow to bypass the broadcasting, so I'm a bit confused.

I'm not saying I disagree but I simply don't understand.

doc/source/changes/version_0_34.rst.inc
alixdamman: In the same bullet ?
Because the problem with Axis doesn't look like the same with the one with None.

gdementen: It is the same problem AFAIK and you fixed it in the same commit too, but you can explain it in two different bullets I don't really care about that.

larray/tests/test_array.py
gdementen: Because it is a bit shorter and faster and if even we don't use those specialized functions, we could as well remove them.
FWIW, to be exactly equivalent, it should be 
```python
ones(small_array.axes, dtype=bool)
```

larray/core/array.py
gdementen: That's an excellent question. In fact, you are right that in the case of the current larray version, broadcasting a scalar Array  against a small array is sufficiently expensive that it is more efficient to "expand" the value manually

```python
>>> %timeit arr == asarray(None)
82.5 µs ± 138 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
>>> %timeit arr == full(arr.axes, fill_value=None)
62.2 µs ± 784 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```

However, given this is the exception to the general rule, it is a very rare case (and it's not the case if the other operand is larger), I do not think it is worth it to optimize for this case, so I still think the shorter "other is None or isinstance(other, (list, Axis)):" is better.

Now for the general performance of broadcasting, the binary operations with broadcasting are **faster** than the operations on "full" arrays which have the same axes/shape, except on very small arrays. I suppose this is because memory access is generally more expensive than CPU cycles. The larger the "full arrays", the more benefit you get from broadcasting.

If you want the details, there is some fixed overhead of broadcasting in larray, which means that the broadcasting binop itself is only faster than its full array equivalent binop when the arrays are "largish" **BUT** if you also take into account the time to create the full array from the "smaller" array, the broadcasting op is faster in all but the most extreme cases (like the scalar Array in a binop with a small array above).

larray/core/array.py
```diff
@@ -5644,9 +5644,7 @@ def opmethod(self, other) -> 'Array':
             # TODO: ndarray should probably be converted to larrays because that would harmonize broadcasting rules, but
             # it makes some tests fail for some reason.
             if not isinstance(other, (Array, np.ndarray)) and not np.isscalar(other):
-                if other is None:
-                    other = full(self.axes, fill_value=None)
-                elif isinstance(other, (list, Axis)):
+                if isinstance(other, (list, Axis)) or other is None:
```
gdementen, line 5647: Well, in fact, I think the best fix is to move the "is None" check in the line above, so that other is passed as is to the numpy binop

```python
if other is not None and not isinstance(other, (Array, np.ndarray)) and not np.isscalar(other):
```

and when looking more at this code, I think it's weirdly written (I know, I was the one to write it!), the two different levels of "if" clause have no point. The following code should be equivalent (not tested though!) and clearer IMO:

```python
            if isinstance(other, (list, Axis)):
                other = asarray(other)
            elif other is not None and not np.isscalar(other) and not isinstance(other, (Array, np.ndarray)):
                # support for inspect.signature
                return False
```

gdementen (approved):
LGTM. You could apply my suggestion below (or not)...

Issue #1079, opened 2023-09-19
building release packages outputs warnings with recent setuptools (?) version


gdementen:
The warning below is repeated for:
* 'larray.tests.data'
* 'larray.tests.data.demography_eurostat'
* 'larray.tests.data.examples'
* 'larray.tests.data.excel_with_links'
* 'larray.tests.excel_template'

See https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=780525&view=logs&jobId=656edd35-690f-5c53-9ba3-09c10d0bea97&j=656edd35-690f-5c53-9ba3-09c10d0bea97&t=986b1512-c876-5f92-0d81-ba851554a0a3

```rst
          # Package would be ignored #
          ############################
          Python recognizes 'larray.tests.data' as an importable package[^1],
          but it is absent from setuptools' `packages` configuration.

          This leads to an ambiguous overall configuration. If you want to distribute this
          package, please make sure that 'larray.tests.data' is explicitly added
          to the `packages` configuration field.

          Alternatively, you can also rely on setuptools' discovery methods
          (for example by using `find_namespace_packages(...)`/`find_namespace:`
          instead of `find_packages(...)`/`find:`).

          You can read more about "package discovery" on setuptools documentation page:

          - https://setuptools.pypa.io/en/latest/userguide/package_discovery.html

          If you don't want 'larray.tests.data' to be distributed and are
          already explicitly excluding 'larray.tests.data' via
          `find_namespace_packages(...)/find_namespace` or `find_packages(...)/find`,
          you can try to use `exclude_package_data`, or `include-package-data=False` in
          combination with a more fine grained `package-data` configuration.

          You can read more about "package data files" on setuptools documentation page:

          - https://setuptools.pypa.io/en/latest/userguide/datafiles.html


          [^1]: For Python, any directory (with suitable naming) can be imported,
                even if it does not contain any `.py` files.
                On the other hand, currently there is no concept of package data
                directory, all directories are treated like packages.
          ********************************************************************************
```

gdementen:
In setup.py, replacing `find_packages` by `find_namespace_packages` silences the warnings but I am not sure at this point whether it is the right thing to do.

PR #1110, opened 2024-07-05
0.34.3

### LINKED ISSUE #1104: Axis.difference(scalar_group) and Axis.difference([scalar_index_group]) fail silently
```python
>>> a = la.Axis('a=a0..a2')
```

The following work fine:

```python
>>> a.difference('a0')
Axis(['a1', 'a2'], 'a')
>>> a.difference(a.i[[0]])
Axis(['a1', 'a2'], 'a')
>>> a.difference(a[['a0']])
Axis(['a1', 'a2'], 'a')
>>> a.difference([a['a0']])
Axis(['a1', 'a2'], 'a')
```

But all the following do not:

```python
>>> a.difference(a['a0'])
Axis(['a0', 'a1', 'a2'], 'a')
>>> a.difference(a.i[0])
Axis(['a0', 'a1', 'a2'], 'a')
>>> a.difference([a.i[0]])
Axis(['a0', 'a1', 'a2'], 'a')
```

priority high even though this is seldom used because it is a silent failure. There is most likely the same problem in Axis.intersection and Axis.union.

I am unsure all those cases should be supported but the asymmetry between string and groups, or even LGroup and IGroup is puzzling, and the unsupported cases (if any) must fail loudly instead of silently.

gdementen:
release 0.34.3 changes.

WIP (still need to fix #1104 and #1079)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,7 +33,7 @@ jobs:
       fail-fast: false
       matrix:
         # os: ["ubuntu-latest", "macos-latest", "windows-latest"]
-        python-version: ['3.8', '3.9', '3.10', '3.11']
+        python-version: ['3.8', '3.9', '3.10', '3.11', '3.12']
 
     defaults:
       run:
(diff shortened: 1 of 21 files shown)


gdementen:
The test failure on Python 3.12 is related to datetime.utcnow deprecation, which is present in xlsxwriter in addition to openpyxl. Unsure why it did not trigger on my machine.

gdementen:
Oh well, I simply did not install xlsxwriter in my dev312 environment 🤷

PR #1150, opened 2025-12-05
checked_sess

gdementen:


gdementen:
TODO (self-reminder): add unit test before merging

